PR #3, opened 2020-11-26
Cypress setup

mbarrenechea:
### Overview

This PR adds:
- Cypress setup
  - Cypress is going to look for every file that contains `*__test__.js`. You can now add `__test__.js`on every place in the app. I didn't check if this will work with imports inside files but at least it's a start.
  - I added a couple of tests looking for the title of a page
  - Cypress github actions working

### Testing instructions

`yarn && yarn cy:run`


weberjavi (approved):
Microphone test, test :microphone: 

image

:tada:

PR #22, opened 2021-01-12
Feature/features

mbarrenechea:
### Overview

This PR adds the features inside the modal:
- Feature Item
- Feature List

image

### DIFF
--- a/app/components/tag/component.tsx
+++ b/app/components/tag/component.tsx
@@ -3,17 +3,15 @@ import cx from 'classnames';
 
 export interface TagProps {
   children: React.ReactNode;
-  className?: string;
+  className?: string | unknown;
 }
 
-export const Tag: React.FC<TagProps> = ({
-  children,
-  className = 'text-black bg-gray-300',
-}: TagProps) => (
+export const Tag: React.FC<TagProps> = ({ children, className }: TagProps) => (
   <div
     className={cx({
       'relative inline-flex rounded': true,
-      [className]: !!className,
+      [`${className}`]: !!className,
+      'text-black bg-gray-200': !className,
     })}
   >
     <div
--- a/app/components/features/item/component.stories.tsx
+++ b/app/components/features/item/component.stories.tsx
@@ -16,6 +16,7 @@ export const Default = Template.bind({});
 Default.args = {
   id: 1,
   name: 'Ecoregions',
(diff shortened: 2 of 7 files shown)


app/components/tag/component.tsx
```diff
@@ -3,17 +3,15 @@ import cx from 'classnames';
...
-export const Tag: React.FC<TagProps> = ({
-  children,
-  className = 'text-black bg-gray-300',
-}: TagProps) => (
+export const Tag: React.FC<TagProps> = ({ children, className }: TagProps) => (
   <div
     className={cx({
       'relative inline-flex rounded': true,
-      [className]: !!className,
+      [`${className}`]: !!className,
```
weberjavi, line 13: Though this works fine I was wondering if it would make to use a `prop` to set tag styles. Setting a `type` prop with some associated styles, like we are doing on the buttons, i.e:
`<Tag type={t.type}/>`  and having on the `interface` some `type: 'bioregional' | 'species' | 'source';`
I'm more inclined to do that because it would not allow devs to mess with the styles.
What do you think?

weberjavi (approved):
Nice and clean 👍🏻 
Left one comment in the code regarding `tag`s styling.
I'm also curious on how do you envision this component evolving, is it going to be a different one from the features displayed outside the modal? or you will handle that through props?
 
 
image

app/components/tag/component.tsx
mbarrenechea: As I don't know yet the different number of colors that we could have I'll leave it like that. But yes, you are right that, if we have a fixed number of them, we could implement them by theme

mbarrenechea:
@weberjavi It will live inside features folder. But I think it will add a lot of complexity. What do you prefer? 2 easy components to use or one complex? I think it's better to have more components than one mega component that do everything. I've been thinking a lot about this... 

Also, there are more components related to features like this one:
image

I'm thinking that I will need to create this structure
```
Features/
  raw-item/
  raw-list/
  selected-item/
  selected-list/
  target-item/
  target-list/  
```

What do you think?

weberjavi:
I like the idea of having the different features cards inside the `features` folder each one as a different component. I think it would scale on a simpler way, and also maintenance, refactoring and extension would be more manageable.
I'm all in with this 👌🏻

PR #27, opened 2021-01-14
Add a modal component

clementprdhomme:
### Overview

This PR adds a modal component to the project and Storybook.

<p align="center">
Modal with a “Lorem Ipsum” content and a blurred overlay
</p>

### Designs

Multiple example of modals can be found in [Figma](https://www.figma.com/file/hq0BZNB9fzyFSbEUgQIHdK/Marxan-Visual_V02?node-id=1523%3A8942).

Each of the modals of the visuals has a different size so I categorised them into 3 categories: narrow (e.g. prompts), default and wide (e.g. modals with a table inside).

Due to the limits of Tailwind, I've tried my best to keep the design's average sizes (~400px, ~520px and ~800px) as stable as possible for all screen sizes. Though, because the modal's size is a percentage of the viewport, the modal's size around the breakpoints differs vastly. If your screen is 1280px wide, that's when you'll see the modal closer to it's target size.

In Figma, the overlay is also blurry. This feature is not supported by all the browsers (e.g. Firefox), so when it's not possible to blur the background, a darker overlay is displayed instead.

### Testing instructions

Open Storybook, test different argument configurations and resize the browser to make sure the modal looks good on all screen sizes. You can also check the actions of the callbacks.

### Feature relevant tickets

None.

### DIFF
--- a/app/components/modal/component.stories.tsx
+++ b/app/components/modal/component.stories.tsx
@@ -1,0 +1,50 @@
+import React, { useState } from 'react';
+import { Story } from '@storybook/react/types-6-0';
+
+import Button from 'components/button';
+import Modal, { ModalProps } from './component';
+
+export default {
+  title: 'Components/Modal',
+  component: Modal,
+  parameters: { actions: { argTypesRegex: '^on.*' } },
+  argTypes: {
+    open: {
+      control: {
+        disable: true,
+      },
+    },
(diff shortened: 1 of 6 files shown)


mbarrenechea (approved):
Good job. I miss some animations but it's something we can add later

PR #32, opened 2021-01-21
Slider component

clementprdhomme:
### Overview

This PR adds a new Slider component that is to be used inside of forms.

<p align="center">
Slider with the label “label” and set to the value 42%
</p>

In the provided Storybook story, you can see the minimum requirements (props) that are necessary. In summary, the 3 essential pieces are:
- An external `<label />` that must have a `for` attribute (or `htmlFor` prop) and is saved in a React ref
- An `id` prop passed to `<Slider />` that has the same value as the label's `for` attribute
- A `labelRef` prop passed to `<Slider />` that is the ref to the `<label />` element

### Designs

The slider can be found in the [Figma project](https://www.figma.com/file/hq0BZNB9fzyFSbEUgQIHdK/Marxan-Visual_V02?node-id=1523%3A2317), on the artboard Marxan 07a.

### Testing instructions

The component is available in the Storybook both in Forms / Slider / Default and Forms / Form where you can see it integrated in a complete form.

In its standalone story, you can test disabling it, changing its state and its configuration such as the minimum and maximum values, the step increment and the value formatter.

### DIFF
--- a/app/components/forms/component.stories.tsx
+++ b/app/components/forms/component.stories.tsx
@@ -7,6 +7,7 @@ import Textarea from 'components/forms/textarea';
 import Select from 'components/forms/select';
 import Checkbox from 'components/forms/checkbox';
 import Radio from 'components/forms/radio';
+import Slider from 'components/forms/slider';
 import Button from 'components/button';
 import {
   composeValidators,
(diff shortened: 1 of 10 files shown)


mbarrenechea (approved):
Good job

PR #33, opened 2021-01-22
Setup of two connections for both services and the initial tables for geoprocessing DB.

aagm:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


api/src/migrations/api/1611329857558-initialDataMetadataEntities.ts
```diff
@@ -0,0 +1,89 @@
+import {MigrationInterface, QueryRunner} from "typeorm";
+
+export class initialDataMetadataEntities1611329857558 implements MigrationInterface {
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+      await queryRunner.query(`
+      CREATE TYPE "status" AS ENUM (
```
hotzevzl, line 7: can we make the type name more descriptive, to make it clear whose kind of process/entity this is a status of.

also, is it this really an attribute of scenarios? in any case, we can refine things later, no need to be super exact here now.

api/src/migrations/api/1611329857558-initialDataMetadataEntities.ts
```diff
@@ -0,0 +1,89 @@
...
+        'created',
+        'running',
+        'done',
+        'failure'
+      );
+
+      CREATE TYPE "tags" AS ENUM (
```
hotzevzl, line 14: maybe same as above, though this could be generic enough. though we may over time have different kinds of tags for different kinds of entities. again, without overthinking this now :)

api/src/migrations/api/1611329857558-initialDataMetadataEntities.ts
```diff
@@ -0,0 +1,89 @@
...
+        "feature_class_name" varchar,
+        "alias" varchar,
+        "description" varchar,
+        "property_name" varchar,
+        "intersection" uuid[],
+        "tag" tags NOT NULL,
+        "creation_status" status NOT NULL,
+        "created_at" timestamp NOT NULL default now(),
+        "created_by" uuid NOT NULL,
+        "modify_at" timestamp NOT NULL default now()
```
hotzevzl, line 31: very tiny detail - maybe `last_modified_at` instead?

api/src/migrations/api/1611329857558-initialDataMetadataEntities.ts
```diff
@@ -0,0 +1,89 @@
...
+      );
+
+      CREATE TABLE "features" (
+        "id" uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+        "feature_data_id" uuid NOT NULL,
+        "feature_class_name" varchar,
+        "alias" varchar,
+        "description" varchar,
+        "property_name" varchar,
+        "intersection" uuid[],
```
hotzevzl, line 26: do we need a `references <table>(<column>)` here?

api/src/migrations/api/1611329857558-initialDataMetadataEntities.ts
```diff
@@ -0,0 +1,89 @@
...
+        "created_at" timestamp NOT NULL default now(),
+        "created_by" uuid NOT NULL,
+        "modify_at" timestamp NOT NULL default now()
+      );
+
+      CREATE TABLE "scenarios" (
+        "id" uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+        "name" varchar,
+        "project_id" varchar NOT NULL REFERENCES "projects" ("id"),
+        "iso3" varchar(3) NOT NULL,
```
hotzevzl, line 38: I have created a `countries` table with iso2, iso3, name and a JSONB for dictionary of local names (if ever needed), so we can later update this to reference the countries table (or keep this column out of this migration and just add it via my PR) - if keeping it, I would call it `country_id` though, for clarity.

api/src/migrations/api/1611329857558-initialDataMetadataEntities.ts
```diff
@@ -0,0 +1,89 @@
...
+        "extent" geometry NOT NULL,
+        "wdpa_filter" jsonb default NULL,
+        "wdpa_threshold" int CHECK (wdpa_threshold BETWEEN 0 AND 100),
+        "admin_region_id" uuid,
+        "runs" int NOT NULL,
+        "blm" float8 NOT NULL,
+        "aditional_fields" jsonb,
```
hotzevzl, line 45: `s/aditional/additional/` - also consider if we want to call these "everything else" columns `metadata` consistently across the data schema.

api/src/migrations/api/1611329857558-initialDataMetadataEntities.ts
```diff
@@ -0,0 +1,89 @@
...
+        "status" status NOT NULL,
+        "parent_id" uuid REFERENCES "scenarios" ("id"),
+        "created_at" timestamp NOT NULL default now(),
+        "created_by" uuid NOT NULL,
+        "modify_at" timestamp NOT NULL default now()
```
hotzevzl, line 50: as above, `last_modified_at` maybe?

api/src/migrations/api/1611329857558-initialDataMetadataEntities.ts
```diff
@@ -0,0 +1,89 @@
...
+        "wdpa_filter" jsonb default NULL,
+        "wdpa_threshold" int CHECK (wdpa_threshold BETWEEN 0 AND 100),
+        "admin_region_id" uuid,
+        "runs" int NOT NULL,
+        "blm" float8 NOT NULL,
+        "aditional_fields" jsonb,
+        "status" status NOT NULL,
+        "parent_id" uuid REFERENCES "scenarios" ("id"),
+        "created_at" timestamp NOT NULL default now(),
+        "created_by" uuid NOT NULL,
```
hotzevzl, line 49: this could already work with `references users(id)`

api/src/migrations/api/1611329857558-initialDataMetadataEntities.ts
```diff
@@ -0,0 +1,89 @@
...
+        "feature_data_id" uuid NOT NULL,
+        "feature_class_name" varchar,
+        "alias" varchar,
+        "description" varchar,
+        "property_name" varchar,
+        "intersection" uuid[],
+        "tag" tags NOT NULL,
+        "creation_status" status NOT NULL,
+        "created_at" timestamp NOT NULL default now(),
+        "created_by" uuid NOT NULL,
```
hotzevzl, line 30: this could already work with `references users(id)`

api/src/migrations/api/1611329857558-initialDataMetadataEntities.ts
```diff
@@ -0,0 +1,89 @@
...
+        "modify_at" timestamp NOT NULL default now()
+      );
+
+      CREATE TABLE "output_results" (
+        "id" uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+        "scenario_id" uuid REFERENCES "scenarios" ("id"),
+        "run_number" int,
+        "score" float8,
+        "cost" float8,
+        "planing_units" float8,
```
hotzevzl, line 59: `s/planing/planning/`

api/src/migrations/api/1611329857558-initialDataMetadataEntities.ts
```diff
@@ -0,0 +1,89 @@
...
+        "parent_id" uuid REFERENCES "scenarios" ("id"),
+        "created_at" timestamp NOT NULL default now(),
+        "created_by" uuid NOT NULL,
+        "modify_at" timestamp NOT NULL default now()
+      );
+
+      CREATE TABLE "output_results" (
+        "id" uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+        "scenario_id" uuid REFERENCES "scenarios" ("id"),
+        "run_number" int,
```
hotzevzl, line 56: does this differ from `scenarios.runs`? if so, could we have this only in one table? it could be ok to have it in both places, but if it's the same concept I would call it the same `number_of_runs`? `runs_count`?

api/src/migrations/api/1611329857558-initialDataMetadataEntities.ts
```diff
@@ -0,0 +1,89 @@
...
+        "score" float8,
+        "cost" float8,
+        "planing_units" float8,
+        "connectivity" float8,
+        "connectivity_total" float8,
+        "mpm" float8,
+        "penalty" float8,
+        "shortfall" float8,
+        "missing_values" float8,
+        "aditional_fields" jsonb
```
hotzevzl, line 66: same as above, both in terms of typo + if we should call these fields all the same

api/src/migrations/api/1611329857558-initialDataMetadataEntities.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+      await queryRunner.query(`
+      DROP TABLE IF EXISTS "features";
+      DROP TABLE IF EXISTS "scenarios";
+      DROP TABLE IF EXISTS "output_results";
+
+
+      DROP TYPE IF EXISTS tags;
+      DROP TYPE IF EXISTS status;
```
hotzevzl, line 84: would you be ok to avoid the `IF EXISTS` here? these are all introduced in this migration (I will update mine to happen later than this in terms of timestamp so we can avoid having to check this), so it should be ok to insist that a plain `DROP` should work.

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+export class initialGeoDBSetup1611221157285 implements MigrationInterface {
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+      await queryRunner.query(`
+      CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
+      CREATE EXTENSION IF NOT EXISTS tablefunc;
+      CREATE EXTENSION IF NOT EXISTS plpgsql;
+      CREATE EXTENSION IF NOT EXISTS postgis;
+      CREATE EXTENSION IF NOT EXISTS postgis_raster; -- OPTIONAL
+      CREATE EXTENSION IF NOT EXISTS postgis_topology; -- OPTIONAL
```
hotzevzl, line 12: This needs wrapping in a guard in case this is ever run in pg < 13 - see https://github.com/Vizzuality/marxan-cloud/pull/23/files#diff-6771bdcda9d1d20d78abccfdd6dad042b632e7a1693145fba93697c5143eca2dR9

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+
+      CREATE TYPE "source_type" AS ENUM (
+        'user_imported',
+        'gbif',
+        'uicn',
```
hotzevzl, line 17: `s/uicn/iucn/`

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+        'intersection'
+      );
+
+      CREATE TYPE "shape_type" AS ENUM (
+        'square',
+        'hexagon',
+        'irregular'
+      );
+
+      CREATE TYPE "status" AS ENUM (
```
hotzevzl, line 28: See comment in the other migration about this

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+        "ogc_fid" int NOT NULL,
+        "the_geom" geometry CONSTRAINT admin_regions_geometry_valid_check CHECK (ST_IsValid(the_geom)),
+        "name_0" varchar,
+        "name_1" varchar,
+        "name_2" varchar,
+        "iso3" varchar(3),
+        "gid_0" varchar,
+        "gid_1" varchar,
+        "gid_2" varchar,
+        "level" adm_level NOT NULL,
```
hotzevzl, line 53: as big fan as I am of enums, couldn't we just use `admin_level integer CHECK (admin_level >= 0 AND admin_level <=2)`?

hotzevzl: I didn't check the syntax, please take it as pseudocode just to show the intention 😬

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+
+-- Administrative regions table.
+      CREATE TABLE "admin_regions" (
+        "id" uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+        "ogc_fid" int NOT NULL,
+        "the_geom" geometry CONSTRAINT admin_regions_geometry_valid_check CHECK (ST_IsValid(the_geom)),
+        "name_0" varchar,
+        "name_1" varchar,
+        "name_2" varchar,
+        "iso3" varchar(3),
```
hotzevzl, line 49: see suggestion in the other migration about this column

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+        "gid_0" varchar,
+        "gid_1" varchar,
+        "gid_2" varchar,
+        "level" adm_level NOT NULL,
+        "created_at" timestamp NOT NULL default now(),
+        "created_by" uuid NOT NULL,
+        "modify_at" timestamp NOT NULL default now()
```
hotzevzl, line 56: as above 😃

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+      ) PARTITION BY LIST(level);
+
+      CREATE TABLE admin_regions_0 PARTITION OF admin_regions FOR VALUES IN (0);
+      CREATE TABLE admin_regions_1 PARTITION OF admin_regions FOR VALUES IN (1);
+      CREATE TABLE admin_regions_2 PARTITION OF admin_regions FOR VALUES IN (2);
```
hotzevzl, line 61: 💓

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+
+      -- Protected Areas table.
+      CREATE TABLE "wdpa" (
+        "id" uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+        "wdpaid" float8,
+        "the_geom" geometry CONSTRAINT wdpa_geometry_valid_check CHECK (ST_IsValid(the_geom)),
+        "full_name" varchar,
+        "iucn_cat" varchar,
+        "shape_leng" float8,
```
hotzevzl, line 70: super tiny - I suggest to abbrev as `len` or to use `length` - `leng` doesn't sound quite right 😓

hotzevzl: unless it reflects an upstream name, ofc 😄

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+        "created_by" uuid NOT NULL,
+        "modify_at" timestamp NOT NULL default now()
+      ) PARTITION BY LIST(level);
+
+      CREATE TABLE admin_regions_0 PARTITION OF admin_regions FOR VALUES IN (0);
+      CREATE TABLE admin_regions_1 PARTITION OF admin_regions FOR VALUES IN (1);
+      CREATE TABLE admin_regions_2 PARTITION OF admin_regions FOR VALUES IN (2);
+
+      -- Protected Areas table.
+      CREATE TABLE "wdpa" (
```
hotzevzl, line 64: very minor - should we call this `wdpa_areas` or something like this, to include the "name of the thing the table is for" (rather than only its origin/kind)?

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+        "id" uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+        "wdpaid" float8,
+        "the_geom" geometry CONSTRAINT wdpa_geometry_valid_check CHECK (ST_IsValid(the_geom)),
+        "full_name" varchar,
+        "iucn_cat" varchar,
+        "shape_leng" float8,
+        "shape_area" float8,
+        "iso3" varchar(3) REFERENCES "admin_regions" ("iso3"),
```
hotzevzl, line 72: as above

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+        "status" text,
+        "desig" text,
```
hotzevzl, line 74: designation? characters in column names is *almost* cost-free (as usual, I am biased - I prefer longer, descriptive names - within reason 😁 )

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+        "created_at" timestamp NOT NULL default now(),
+        "created_by" uuid NOT NULL,
+        "modify_at" timestamp NOT NULL default now()
```
hotzevzl, line 77: as above

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+        "source" source_type
+      );
+
+      -- Usage of features data within scenarios table (will hold species and bioma data).
+      CREATE TABLE "scenario_features_data" (
+        "id" uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+        "feature_class_id" uuid,
+        "scenario_id" uuid,
+        "total_area" varchar,
+        "admin_area" timestamp,
```
hotzevzl, line 94: this doesn't sound right

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+        "target2" float8,
+        "targetocc" float8,
+        "sepnum" float8,
+        "created_at" timestamp NOT NULL default now(),
+        "created_by" uuid NOT NULL,
+        "modify_at" timestamp NOT NULL default now()
+      );
+
+      -- Planning units geometry data table.
+      CREATE TABLE "planing_units_geom" (
```
hotzevzl, line 107: `s/planing/planning/`

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+        "id" uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+        "the_geom" geometry UNIQUE NOT NULL CONSTRAINT features_geometry_valid_check CHECK (ST_IsValid(the_geom)),
+        "type" shape_type,
+        "size" int
```
hotzevzl, line 111: just a very mild suggestion - unless it's very clear what the measurement unit is, I find it useful to add it in the name for very generic column names such as `size`

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+      ) PARTITION BY LIST (type);
+
+      CREATE TABLE planing_units_geom_square PARTITION OF planing_units_geom FOR VALUES IN ('square');
+      CREATE TABLE planing_units_geom_hexagon PARTITION OF planing_units_geom FOR VALUES IN ('hexagon');
+      CREATE TABLE planing_units_geom_irregular PARTITION OF planing_units_geom FOR VALUES IN ('irregular');
```
hotzevzl, line 116: this is ok, but probably we'll have to revisit - I wonder if partitioning by geometry type is a good strategy to keep partitions relatively balanced and usefully balanced 🤔 - just a thought, but nothing to do there 😃

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+
+      -- Planning units usage by scenario.
+      CREATE TABLE "scenarios_pu_data" (
```
hotzevzl, line 119: I would stick to `<entity_one_plural>_<entity_two_plural>` for tables for M-M relationships, so this could be `scenarios_planning_units`

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+        "id" uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+        "pu_geomid" uuid NOT NULL REFERENCES "planing_units_geom" ("id"),
```
hotzevzl, line 121: I would stick to `<referenced_entity>_id`, so `planning_units_geom_id`

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+        "scenario_id" uuid,
+        "run_id" uuid,
+        "value" float8,
+        "missing_values" jsonb
+      );
+
+
+      COMMENT ON TABLE "admin_regions" IS 'will cover adm0,adm1 and adm2';
+      COMMENT ON TABLE "features_data" IS 'For geometry storage';
+      COMMENT ON COLUMN "features_data"."properties" IS 'Either have the properties as a jsonb or as columns on this tables';
```
hotzevzl, line 149: do we need this comment in the db? it looks more like something that could belong to a design document - I like the idea of having properties as a JSONB to start with as in this migration, and we could revisit this (or split out of the JSONB properties that work best as columns) later on.

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
hotzevzl: I see there's a comment for this, but (very personal preference!) I would just rename this `size_in_km2` and avoid the comment

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+      DROP TABLE IF EXISTS "planing_units_geom";
+      DROP TABLE IF EXISTS "scenario_features_data";
+      DROP TABLE IF EXISTS "features_data";
+      DROP TABLE IF EXISTS "admin_regions";
+      DROP TABLE IF EXISTS "wdpa";
+
+
+      DROP TYPE IF EXISTS source_type;
+      DROP TYPE IF EXISTS shape_type;
+      DROP TYPE IF EXISTS status;
```
hotzevzl, line 174: see my comment in the other migration regarding `IF EXISTS`

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+
+      DROP EXTENSION postgis_topology; -- OPTIONAL
+      DROP EXTENSION postgis_raster; -- OPTIONAL
+      DROP EXTENSION postgis;
+      DROP EXTENSION plpgsql;
+      DROP EXTENSION tablefunc;
+      DROP EXTENSION uuid-ossp;
```
hotzevzl, line 181: should be wrapped in guard for pg < 13; personally I would not bother about dropping `uuid-ossp` and `plpgsql`, same as `pgcrypto` in the migration where I added this - in theory not even the postgis stuff as we'd need it for sure and one it's enabled it's enabled - but ok to drop everything, ofc 😄

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
hotzevzl: this partition won't work as is now because of limitations to PostgreSQL partitions: https://www.postgresql.org/docs/13/ddl-partitioning.html#DDL-PARTITIONING-DECLARATIVE-LIMITATIONS

Specifically:
> Unique constraints (and hence primary keys) on partitioned tables must include all the partition key columns. This limitation exists because PostgreSQL can only enforce uniqueness in each partition individually.

so one way would be to use `PRIMARY KEY(id, level)` - in theory this could mean that a clash of ids across different partitions would prevent partitions from being joined later on, but since `id` is UUIDv4 this may happen after the end of the universe...

or we could switch to partitioning by hash, but I don't think this makes sense here

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
```diff
@@ -0,0 +1,185 @@
...
+        "modify_at" timestamp NOT NULL default now()
+      );
+
+      -- Planning units geometry data table.
+      CREATE TABLE "planing_units_geom" (
+        "id" uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+        "the_geom" geometry UNIQUE NOT NULL CONSTRAINT features_geometry_valid_check CHECK (ST_IsValid(the_geom)),
+        "type" shape_type,
+        "size" int
+      ) PARTITION BY LIST (type);
```
hotzevzl, line 112: this won't work, but here the issue is slightly different from the other partition on `admin_regions`, because of the `UNIQUE` constraint on the `the_geom` column, combined with the fact that we may not be able to use a hash partition by hash of `the_geom` because

```
ERROR:  operator class "hash_geometry_ops" of access method hash is missing support function 2 for type geometry
```

(and I doubt hash partitioning would be useful here in any case)...

I think we should try to find a different way to partition this - eventually. For the moment, using `PRIMARY KEY (id, type)` should work.

api/src/migrations/geoprocessing/1611221157285-initialGeoDBSetup.ts
hotzevzl: I think numeric enums won't work in any case, so either the enum is

```CREATE TYPE "adm_level" AS ENUM (
        '0',
        '1',
        '2'
      );
```

or we use the check on integers I suggested above (I still think it's the simplest option!)

api/src/migrations/api/1611329857558-initialDataMetadataEntities.ts
aagm: work_status

PR #36, opened 2021-01-26
chore: update docker compose config and make targets for API development (API and geoprocessing)

hotzevzl:
### Overview

Update docker-compose configuration, restoring what we had commented out.

Services that are not needed now will not start anymore if we use `make start-api`.

`make start` will attempt to spin up all the services, though the `app` one is not fully configured yet, so this target should not be used for now.

Additionally, it's now trivial to enter a `psql` console both for the API db (`make psql-api`) and for the geoprocessing db (`make psql-geo`).

### Designs

N/A

### Testing instructions

Stop all the Marxan services (`make stop`).

`make start-api` should start api and geoprocessing services, plus their backing pg services and the redis-api service (not used yet, but we may soon).

`make psql-api` and `make psql-geo` should lead to the psql console for the api and geoprocessing db, respectively.

### Feature relevant tickets

N/A

---

- [ ] Update CHANGELOG file

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ endif
 # Useful when developing on API components only, to avoid spinning services
 # which may not be needed.
 start-api:
-	docker-compose up --build api geoprocessing postgresql-api postgresql-geo-api redis-api
+	docker-compose up --build api geoprocessing
 
 # Start all the services.
 start:
--- a/Makefile
+++ b/Makefile
@@ -21,8 +21,11 @@ notebooks:
 stop:
 	docker-compose stop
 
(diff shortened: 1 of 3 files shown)


hotzevzl:
@aagm I think this PR will help to keep configs tidy, while allowing us to spin up locally only services that we need/are ready (I already added the `redis-api` service as dependency as it should be relatively lightweight and I think we'll need it soon in any case).

aagm:
Works, Ready to Rebase

aagm (approved):
ready to be reabase

PR #38, opened 2021-01-28
scenarios module - part 1

hotzevzl:
### Overview

This PR:

* adds all the columns of the current `scenarios` table to the `Scenarios` entity (with some per-prop documentation)
* fixes some missing cascades via a migration
* adds a stub of the `AdminRegion` entity (for the API service), mostly to validate the ORM setup for multiple connections
* refactors the ORM config so that we can easily use multiple connections in NestJS via TypeOrmModule, while sharing configuration code with the ORM CLI utility (`ormconfig.ts`) - as a positive externality of this, updates to the ORM configuration are now effective immediately in development once the app restarts - no need to rebuild the container image anymore 🚀)
* adds a new base class with entity columns which are common to several of our metadata classes (`createdAt`, `createdBy`, `lastModifiedAt`)
* introduces hard failure in `AppConfig.get()` when loading config values that are not defined, and no fallback default is provided
* adds a PG version check guard around the existing migration in the geoprocessing service, to avoid trying to run `CREATE EXTENSION` or `DROP EXTENSION` in a migration if the PostgreSQL server version is <13

### Designs

N/A

### Testing instructions

Things should get up from scratch (i.e. after a `make clean-slate`). This also means all migrations should run, with the NestJS TypeOrmModule correctly finding the migration files, and that querying an endpoint that loads entities (e.g. `curl http://localhost:3030/api/v1/scenarios -H 'Authorization: Bearer <token>'`) should work ok (even if they return no data or fake data).

Swagger docs for controllers (e.g. http://localhost:3030/swagger/#/Scenario/ScenariosController_create) should show the full schema for Scenarios in the response payload schema (I haven't worked on the create/update DTOs yet).

TypeORM's CLI util should run correctly (hence the new ormconfig setup is loaded correctly by the CLI util) - try manually reverting and the running again a migration to check this.

### Feature relevant tickets

https://www.pivotaltracker.com/story/show/176680127

---

- [x] Update CHANGELOG file

### DIFF
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -22,6 +22,7 @@ Unreleased
   - Get users
   - Get current user info (`/users/me`)
 - Stub `ProjectsModule`.
+- Stub `ScenariosModule`.
 - Stub roles.
 - PoC JSON:API serialization.
 - Authentication
@@ -32,6 +33,8 @@ Unreleased
 - Ping/liveness endpoint.
 - NestJS services are based on `nestjs-base-service`.
 - Integration with faker.js for mock data.
+- Shared ORM config setup, allowing to use multiple connections in NestJS as
+  well as via the TypeORM CLI util.
 
 ### Changed
 
--- a/api/ormconfig.ts
+++ b/api/ormconfig.ts
@@ -1,61 +1,16 @@
-import { AppConfig } from "utils/config.utils";
-
-const config = require('config');
+import { apiConnections } from "ormconfig";
 
 /**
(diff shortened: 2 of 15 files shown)


aagm (approved):
Ready to go

hotzevzl:
@aagm thanks for reviewing.

I think 5689ce4dea9136d8aaac4b61d13e8e52b7734d14 and 22ab71ad135f79a50bba87f8ea8063e80dd25a2e should fix this in general - I'll merge

PR #49, opened 2021-02-10
[MARXAN-74] remove fake data wirings

hotzevzl:
### Overview

Replace use of fake data from earlier phase of API development with reading/writing of actual data from/to db.

Plus e2e tests for Scenarios and some minor updates to other tests.

### Designs

N/A

### Testing instructions

This is a bit tricky to test - basically all endpoints (GET singular or plural, POST, PATCH and DELETE) are now using real data from the API db connection... So I think that short of manually hitting all endpoints, we may assume that I didn't forget to remove wirings (the TypeScript language server shouted at me for some bits I had forgotten during this PR).

In general, I think it's time to start upping our e2e tests to not only check success/failure as expected, but also to check that the data we receive from the API is what we expect - but for the moment, the tests we have don't do this yet.

### Feature relevant tickets

https://vizzuality.atlassian.net/browse/MARXAN-74

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/src/modules/countries/countries.controller.ts
+++ b/api/src/modules/countries/countries.controller.ts
@@ -51,6 +51,6 @@ export class CountriesController {
   @ApiOkResponse({ type: CountryResult })
   @Get(':id')
   async findOne(@Param('id') id: string): Promise<Country> {
-    return await this.service.fakeFindOne(id);
+    return await this.service.getById(id);
   }
 }
--- a/api/src/modules/organizations/organizations.controller.ts
+++ b/api/src/modules/organizations/organizations.controller.ts
@@ -73,7 +73,7 @@ export class OrganizationsController {
   @ApiOkResponse({ type: OrganizationResult })
   @Get(':id')
   async findOne(@Param('id') id: string): Promise<Organization> {
-    return await this.service.serialize([await this.service.fakeFindOne(id)]);
+    return await this.service.serialize([await this.service.getById(id)]);
   }
 
   @ApiOperation({ description: 'Create organization' })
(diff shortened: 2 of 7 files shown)


aagm:
I have 2 test failing 
```
Test Suites: 1 failed, 2 passed, 3 total
 Tests:       1 failed, 18 passed, 19 total
```
 
 ```
ScenariosModule (e2e) › Scenarios › Gets scenarios
 TypeError: Cannot read property 'type' of undefined
```
 
 In the test data i'm not seeing a seed for scenarios anywhere, can it be linked with this error? apart from that using the swagger doc to follow how to call the endpoints works as expected.

aagm (changes requested):
we will need to add the data to the fixtures right?

hotzevzl:
[...]
> In the test data i'm not seeing a seed for scenarios anywhere, can it be linked with this error? apart from that using the swagger doc to follow how to call the endpoints works as expected.

you are right - we're not adding scenarios anywhere - nor in a seed nor before trying to get a list of scenarios in this or any other e2e test.

I am fixing this (d6dbb0d1c6f909811154b2fb6bc4147c6ee2a5d3) by reordering the tests in this suite (Scenarios) so that we make sure we create at least one scenario before GETting many and checking that the first one returned (could be just one, at this stage, on a clean-slate system) matches our assertions.

In general, I think we can avoid (in general) adding more `INSERT`s to the SQL test data seed - that was an early way to get some data in for basic manual tests, but overall I would prefer to have test data inserted via a single source, which could be these e2e tests - I think there is a place for a SQL seed, but what I would like to avoid is to having to add both new `INSERT`s in a seed *and* in e2e tests. Of course whatever we create in e2e tests is not meant to *stay* in the db, but my feeling is that we should exploit e2e tests to *also* have a way to populate a system with test data.

Let's reflect about this (also going back to earlier discussions about property-based testing) - for this PR I suggest to limit the change to the reordering of tests I have added in the commit referenced above.

PR #53, opened 2021-02-10
Authentication

mbarrenechea:
### Overview

This PR adds:
- `useAuth` hook to be able to access the user and functions everywhere thanks to a provider
- Passport + Next connect for implementing local strategy, API for saving a secure cookie in the client and deserialize it

### Testing instructions

Go to Marxan web and use the correct credentials

## Things missing
- [x] Improve signing, signup forms to handle errors.
- [x] Improve signing, signup visually.
- [x] Improve header behaviour


hotzevzl:
@mbarrenechea we have added `displayName` as optional field for signups in an API PR ready for review, to follow the designs.

You won't see this in the OpenAPI docs on the dev instance until we approve/merge/deploy, so for reference the DTO is this one:
https://github.com/Vizzuality/marxan-cloud/blob/f4e8a4328c7770378d31ad81a4b765c006808ac5/api/src/modules/authentication/dto/sign-up.dto.ts

When editing users (no POST/PUT/DELETE endpoints for these yet - would you find it useful to have these sooner rather than later or are you happy to work with the other endpoints we already have? just let us know and we'll schedule accordingly), the API will allow to set other props such as (if needed) distinct first and last name, and a JSON `metadata` field (which in theory could be used as a generic K/V store, though since I am a nasty person I always prefer to *eventually* converge on versioned schemas to avoid sudden cosmic bangs due to random stuff creeping in JSON documents). We should also discuss what you would like to use for avatars. Sorry for hijacking an authn PR with user-related drivel, I guess we should move this to ~PT~ Jira 😬

mbarrenechea:
@hotzevzl Ok, I didn't implement yet the user creation. Good to know in advance all the params involved.

mbarrenechea:
@hotzevzl regarding the user edition, designers are working on this, so I'm not in a rush

app/pages/projects.tsx
```diff
@@ -0,0 +1,36 @@
...
+import Protected from 'layout/protected';
+
+const ProjectsPage: React.FC = () => {
+  const { user, errorRedirect } = useAuth();
+  const router = useRouter();
+
+  if (!user) {
+    router.push(errorRedirect);
+    return null;
+  }
```
clementprdhomme, line 18: This is already handled by the `Protected` component.

clementprdhomme (approved):
I reviewed the code and logged in/out of the platform too. The code is better than what I would have done 👍.

I added a few minor comments.

app/pages/projects.tsx
mbarrenechea: Thanks for the catch

app/components/button/component.tsx
```diff
@@ -30,15 +30,17 @@ export const Button: React.FC<ButtonProps> = ({
   theme = 'primary',
   size = 'base',
   className,
+  disabled,
```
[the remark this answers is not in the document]
mbarrenechea: 👍

pablopareja:
I've had a look at the architecture and methods and it looks quite cool! 👏 
I've made a suggestion [here](https://github.com/Vizzuality/marxan-cloud/pull/56) to use preliminary interfaces for both the `User` object and the `AuthContextProps`
@mbarrenechea Let me know what you think ;)

PR #62, opened 2021-02-19
Feature/new project

pablopareja:
image

### Overview

This PR adds all the interface elements plus logic necessary for a first iteration of the "New project" user flow.

### Designs

- https://www.figma.com/file/hq0BZNB9fzyFSbEUgQIHdK/Marxan-Visual_V02?node-id=1530%3A87

### Testing instructions

Go to `http://localhost:3000/projects/new` and check all the different options that are part of the form.

### Feature relevant tickets

- https://vizzuality.atlassian.net/browse/MARXAN-102?atlOrigin=eyJpIjoiMTcyNmNlYjZlZjFjNDBiNWI4ZDIxNWMyZDc2Yzc1MzkiLCJwIjoiaiJ9


app/components/button/component.tsx
```diff
@@ -6,6 +6,8 @@ const THEME = {
     'text-black bg-primary-500 hover:bg-primary-400 active:bg-primary-300 border border-primary-500 hover:border-primary-400 active:border-primary-300',
   'primary-alt':
     'text-primary-500 bg-transparent hover:bg-transparent active:bg-transparent border border-primary-500 hover:border-primary-400 active:border-primary-300',
+  'primary-white':
```
mbarrenechea, line 9: I'd call it directly `white`

app/layout/projects/new/component.tsx
```diff
@@ -0,0 +1,241 @@
...
+        <div className="relative flex items-center justify-center w-1/2 h-full text-white">
+          <Map
+            className=""
+            width="100%"
+            height="100%"
+            bounds={bounds}
+            minZoom={minZoom}
+            maxZoom={maxZoom}
+            viewport={viewport}
+            mapboxApiAccessToken={process.env.STORYBOOK_MAPBOX_API_TOKEN}
```
mbarrenechea, line 216: The map is not going to work in the app because you didn't send the .env variable for that.
```suggestion
            mapboxApiAccessToken={process.env.NEXT_PUBLIC_MAPBOX_API_TOKEN}
```

mbarrenechea (changes requested):
You must use `FieldRFF` to wrap every custom component. Otherwise, you don't know the current state of the form. You can check the sample in the storybook with selects.

mbarrenechea:
I would recommend base your layout in this PR https://github.com/Vizzuality/marxan-cloud/pull/65/files#diff-a90b63aa88d45665d0b00c4c973efa412f9514ffa791c74b98d192cc17a556d2R30

I mean, split your new component into two components. One for the Map and another one for the form. I think it's something we talked about

app/components/projects/planning-unit-grid/component.tsx
```diff
@@ -0,0 +1,60 @@
...
+          size={PlanningUnitButtonSizeProps.MEDIUM}
+          onClick={handleClick}
+        />
+        <PlanningUnitButton
+          unit={PlanningUnit.HEXAGON}
+          selected={unitSelected === PlanningUnit.HEXAGON}
+          size={PlanningUnitButtonSizeProps.MEDIUM}
+          onClick={handleClick}
+        />
+        <PlanningUnitButton
```
mbarrenechea, line 49: Remove this one, I think we will need to think about another way of dealing with a custom grid. After talking with Alicia it seems that you don't upload a custom shape, you upload a custom grid, so it will override the planning unit size too. It's a design change. Remove it from this list please

app/hooks/projects/types.ts
```diff
@@ -1,5 +1,15 @@
 import { AxiosRequestConfig } from 'axios';
+import { Response } from 'types/api-model';
 
+// useProjects
+export interface UseProjectsProps {
+  search?: string;
+}
+export interface UseProjectsResponse extends Response {
```
mbarrenechea, line 8: instead of using your own type Response I think you should use `UseQueryResult` from react-query

https://github.com/tannerlinsley/react-query/blob/master/src/react/types.ts#L42

mbarrenechea (changes requested):

mbarrenechea (changes requested):
About design:
- You add something to the selects because they have a border blue that shouldn't be there
- The spinner is always present. Check this => https://github.com/Vizzuality/marxan-cloud/blob/develop/app/layout/projects/list/component.tsx#L27
- There isn't a space between the header and the form/map
- the min-height of the form/map should be the maximum height available. Check this => https://github.com/Vizzuality/marxan-cloud/blob/develop/app/pages/projects/%5Bpid%5D/scenarios/new.tsx#L30

app/hooks/projects/types.ts
pablopareja: That change wouldn't be trivial. We would need to do something like this: https://blog.johnnyreilly.com/2021/01/strongly-typing-react-query-s-usequeries.html

_It wouldn't as simple as creating an interface that extends `UseQueryResult` since that can't be done just like that_

pablopareja:
> About design:
> 
>     * You add something to the selects because they have a border blue that shouldn't be there

I haven't added anything special to the selects as far as I know. I've been inspecting the select and that blue color seems to be originated from a class that's part of the theme definitions: `ring-2`

image
image

>     * The spinner is always present. Check this => https://github.com/Vizzuality/marxan-cloud/blob/develop/app/layout/projects/list/component.tsx#L27

This should be fixed in one of the latest commits 👍 

>     * There isn't a space between the header and the form/map

It should be fixed in one of the latest commits 👍 

>     * the min-height of the form/map should be the maximum height available. Check this => https://github.com/Vizzuality/marxan-cloud/blob/develop/app/pages/projects/%5Bpid%5D/scenarios/new.tsx#L30

I've tried using the same approach that you mentioned but I get an horizontal scroll then:
image

mbarrenechea (approved):

PR #66, opened 2021-02-23
fixed error issue with wdpa pipe

aagm:
fixed error  issue with wdpa pipe

### Overview

Fixed bug that prevented wdpa not running.

### Testing instructions

`make seed-geodb-data`

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/data/data_download/wdpa/Makefile
+++ b/data/data_download/wdpa/Makefile
@@ -1,3 +1,4 @@
+SHELL=/bin/bash
 .PHONY: import
 MAKEFLAGS := --jobs=$(shell nproc)
 MAKEFLAGS += --output-sync=target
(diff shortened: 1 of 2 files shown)


data/data_download/wdpa/Makefile
```diff
@@ -34,31 +35,47 @@ merge-simplify: data/wdpa/simp/0 data/wdpa/simp/1 data/wdpa/simp/2
...
+	@unzip -u $< -d data/wdpa/out
+	@for f in data/wdpa/out/*.zip; \
+		do echo "$$(basename $$f)"; \
+		mkdir -p $@/temp; \
+		unzip "$$f" -d $@/temp; \
+		mv $@/temp/*-polygons.shp $@/$$(basename $$f .zip).shp; \
+   		mv $@/temp/*-polygons.shx $@/$$(basename $$f .zip).shx; \
+		mv $@/temp/*-polygons.dbf $@/$$(basename $$f .zip).dbf; \
+      	mv $@/temp/*-polygons.prj $@/$$(basename $$f .zip).prj; \
+        mv $@/temp/*-polygons.cpg $@/$$(basename $$f .zip).cpg; \
```
hotzevzl, line 73: I think this may potentially break if at some point upstream starts putting in a folder more than one file with suffix `-polygons.shp` and so on, but at that stage the pipeline will break and we'll fix it.
It may be slightly worse in case they split some of these files into two or more with different prefix only for some areas, because the error would be flaky, but I'd say that this is unlikely in general, so I can take my tin foil hat off here :)

hotzevzl (commented):
Testing this on a test VM, the WDPA import exits with errors, from what I understand related to missing UUIDs for some rows being inserted in the `wdpa` table - see attached container log for details

[848ba9cf604004b6cf71ee8e291b391c73dad5fa8e8b08c4a875c1d74278a61d-json.log](https://github.com/Vizzuality/marxan-cloud/files/6031851/848ba9cf604004b6cf71ee8e291b391c73dad5fa8e8b08c4a875c1d74278a61d-json.log)

data/data_download/wdpa/Makefile
aagm: hopefully, when this breaks we will no longer need it :crossed_fingers: . Let me review the issues in the log file.

aagm: after reviewing and reruning the seed on the instance... things are working as expected, you can check that the data is populated in the db

aagm:
make finish successfully in the VM :D so merging the pr; we have data!

PR #72, opened 2021-02-24
Header: title edition

mbarrenechea:
### Overview

This PR adds:
- Form instead of titles for allowing project and scenario edition

### Testing instructions
https://marxan-git-feature-header-title-edition-vizzuality1.vercel.app/projects/b0a14339-732a-430b-a01d-6bc06494b05f/scenarios/374b2483-a5ce-4a76-8b24-f4cafb49aa6b/edit

Go there and edit names. You can try to put a very long name if you want

### Feature relevant tickets
https://vizzuality.atlassian.net/browse/MARXAN-105

### DIFF
--- a/app/hooks/projects/types.ts
+++ b/app/hooks/projects/types.ts
@@ -1,3 +1,5 @@
+import { AxiosRequestConfig } from 'axios';
+
 export interface UseSaveProjectProps {
-  requestOptions: {}
+  requestConfig: AxiosRequestConfig
 }
--- a/app/hooks/projects/types.ts
+++ b/app/hooks/projects/types.ts
@@ -1,0 +1,3 @@
+export interface UseSaveProjectProps {
+  requestOptions: {}
+}
--- a/app/layout/header/title/component.tsx
+++ b/app/layout/header/title/component.tsx
@@ -1,9 +1,14 @@
-import React from 'react';
+import React, { useCallback } from 'react';
 
-import { useProject } from 'hooks/projects';
-import { useScenario } from 'hooks/scenarios';
+import { useProject, useSaveProject } from 'hooks/projects';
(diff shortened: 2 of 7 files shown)


app/hooks/projects/types.ts
```diff
@@ -0,0 +1,3 @@
+export interface UseSaveProjectProps {
+  requestOptions: {}
```
pablopareja, line 2: I think we should probably create an interface for the request options as well. The good thing about doing that is that we could reuse that definition in many other hooks.

app/layout/header/title/component.tsx
```diff
@@ -14,22 +19,116 @@ export const Title: React.FC<TitleProps> = () => {
...
+
+  // Scenario mutation and submit
+  const saveScenarioMutation = useSaveScenario({
+    requestOptions: {
+      method: 'PATCH',
+      url: `/${scenarioData?.id}`,
+    },
+  });
+
+  const handleScenarioSubmit = useCallback(async (data) => {
```
pablopareja, line 55: Regarding the user experience of this feature, it feels weird to me that the focus is kept on the input after clicking `Enter`. I only knew that the update was successful because I had the developer's console open. I think that it would be more intuitive for users if the input lost its focus once the request has been completed and the update was effective.

pablopareja (changes requested):

app/hooks/projects/types.ts
mbarrenechea: You are right, I'm going to use the AxiosRequestConfig
```
import { AxiosRequestConfig } from 'axios';

export interface UseSaveProjectProps {
  requestConfig: AxiosRequestConfig
}
```

app/layout/header/title/component.tsx
mbarrenechea: Ready

PR #76, opened 2021-03-01
get admin areas by country

hotzevzl:
### Overview

New or updated endpoints:

`GET /api/v1/countries/:id/administrative-areas` should list (with pagination) admin areas within a given country, with `:id` being the ISO3166 alpha-3 code for the country (uppercase).

`GET /api/v1/administrative-areas` should list (with pagination) admin areas.

`GET /api/v1/administrative-areas/:id` should return the level 1 or level 2 admin area matching the id provided.

MARXAN-135 includes discussions of more advanced filters for searching admin areas within a country, but I will do that after we define the requirements for these filters, in a separate PR. This can be reviewed as is.

### Designs

N/A

### Testing instructions

See description above.

We do have some e2e/contract tests in this PR, but they will fail in CI until we change the GitHub actions flow to include seeding the geodb with a minimal set of gadm data.

Running `docker-compose exec api yarn test:e2e` locally should work, however, if the local instance has GADM data.

### Feature relevant tickets

https://vizzuality.atlassian.net/browse/MARXAN-135
https://vizzuality.atlassian.net/browse/MARXAN-134

---

- [x] Update CHANGELOG file


hotzevzl:
@aagm during today's sprint meeting I said we would not need more filters for the expected UX, although I later realized we would need an endpoint to list level2 areas given a level1 id - or API consumers would need to go through some juggling to do this client-side, which would result in a brittle solution.

Nevertheless, I'll add this other feature to a separate PR - this can still be reviewed as-is so we can avoid keeping it open for too long. I have kept the Jira story as In progress for this reason, in any case.

aagm:
Ok the migration fails because it needs an index based on the partition table, so my suggestion and after the issue is already addressed in this [pr 74](https://github.com/Vizzuality/marxan-cloud/pull/74) to drop the migration.

geoprocessing/src/migrations/geoprocessing/1614256630000-AddCountriesView.ts
```diff
@@ -0,0 +1,28 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddCountriesView1614256630000
+  implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+CREATE UNIQUE INDEX only_countries_by_gid ON admin_regions(gid_0) WHERE (gid_0 IS NOT NULL AND gid_1 IS NULL AND gid_2 IS NULL);
```
aagm, line 7: i suggest to drop this indexes from the migration, as the table is a partitioned one and after the patch mentioned in #74 this is no longer needed

aagm (changes requested):

geoprocessing/src/migrations/geoprocessing/1614256630000-AddCountriesView.ts
```diff
@@ -0,0 +1,28 @@
...
+CREATE UNIQUE INDEX only_countries_by_name ON admin_regions(name_0) WHERE (name_0 IS NOT NULL AND name_1 IS NULL AND name_2 IS NULL);
+CREATE VIEW countries AS (
+  SELECT
+    ogc_fid,
```
aagm, line 11: ogc_fid is also a ghost parameter introduced by ogr and was also fixed in #74

aagm (changes requested):

geoprocessing/src/migrations/geoprocessing/1614256630000-AddCountriesView.ts
hotzevzl: ok, done

geoprocessing/src/migrations/geoprocessing/1614256630000-AddCountriesView.ts
hotzevzl: ok, done, thanks!

aagm:
This is good to merge Andrea!

hotzevzl:
> Im having issues with GET /api/v1/administrative-areas as endpoint not found

got it, we don't have an endpoint that returns a global list of admin areas - the above can only be used as `GET /api/v1/administrative-areas/:id`

PR #77, opened 2021-03-01
[MARXAN-128] Allow to delete one's own user

hotzevzl:
### Overview

`DELETE /api/v1/users/me` should (soft-) delete one's own user

Basically we set `isActive = false` and `isDeleted = true`, which flags the user for deletion **and** prevents them from logging in, while keeping the actual user account in the db.

We don't deal with the details of physically deleting the user at this stage - this would mean taking some decisions on "detaching" the user account from db entities it may be linked to. We may need to do something here because of data protection compliance, but we could also potentially deal with this by removing personal identification data from users flagged for deletion, without having to deal with more complexity which could be avoided in this phase.

### Designs

N/A

### Testing instructions

`DELETE /api/v1/users/me` should succeed for logged-in users

After this, any existing and still-valid (timestamp-wise) JWT tokens must be rejected, and it should not be possible to obtain new JWT tokens via `POST /api/v1/auth/sign-in`.

### Feature relevant tickets

https://vizzuality.atlassian.net/browse/MARXAN-128

---

- [x] Update CHANGELOG file

### DIFF
--- a/api/test/countries.e2e-spec.ts
+++ b/api/test/countries.e2e-spec.ts
@@ -23,8 +23,8 @@ describe('CountriesModule (e2e)', () => {
     const response = await request(app.getHttpServer())
       .post('/auth/sign-in')
       .send({
-        username: E2E_CONFIG.users.aa.username,
-        password: E2E_CONFIG.users.aa.password,
+        username: E2E_CONFIG.users.basic.aa.username,
+        password: E2E_CONFIG.users.basic.aa.password,
       })
       .expect(201);
 
--- a/api/src/main.ts
+++ b/api/src/main.ts
@@ -40,7 +40,11 @@ async function bootstrap() {
   const swaggerDocument = SwaggerModule.createDocument(app, swaggerOptions);
   SwaggerModule.setup('/swagger', app, swaggerDocument);
 
-  app.useGlobalPipes(new ValidationPipe({ transform: true }));
+  app.useGlobalPipes(
+    new ValidationPipe({
+      transform: true,
+    }),
+  );
   app.useGlobalFilters(new AllExceptionsFilter());
 
   await app.listen(3000);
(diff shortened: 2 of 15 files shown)


api/test/countries.e2e-spec.ts
```diff
@@ -53,13 +53,13 @@ describe('CountriesModule (e2e)', () => {
...
       const response = await request(app.getHttpServer())
         .get(
-          `/api/v1/countries/${aCountry.gid0}/administrative-areas?page[size]=25`,
+          `/api/v1/countries/${aCountry.attributes.gid0}/administrative-areas?page[size]=25`,
         )
         .set('Authorization', `Bearer ${jwtToken}`)
         .expect(200);
 
       const resources = response.body.data;
-      expect(resources[0].type).toBe('administrative-areas');
```
aagm, line 62: is this intentional?

aagm (changes requested):
maybe is just because as you said to me earlier this week this prs are a bit chaotic, but I think the country stuff has slipped through this. 
I also miss a test for deleting users. 

Apart from that testing the delete endpoint works as expected and does not allow me to sign in again.

api/test/countries.e2e-spec.ts
hotzevzl: yes and no - it was ok when I created this PR, but I have eventually amended it to read `admin_areas` in the latest commits before merging #76 as it was by then out of sync with what the API would return

hotzevzl:
> maybe is just because as you said to me earlier this week this prs are a bit chaotic, but I think the country stuff has slipped through this.

yep, that was because I based each PR on the previous one to be able to deploy on marxan49 all the successive changes for FEs to work with them.

> I also miss a test for deleting users.

that's right - I had stashed this test file away and never remembered to bring it back, doing that now.

Adding tests is the change requested here, right? As for the countries changes, I have rebased this PR on the latest `develop` now and so those changes are not included in the PR anymore.
> 
> Apart from that testing the delete endpoint works as expected and does not allow me to sign in again.

PR #78, opened 2021-03-02
Notifications

mbarrenechea:
### Overview

This PR adds:
- Toast Provider
- Toast Container
- Toast Component

You can always get the functions from the toasts by using the `useToasts` hook. 

### Testing instructions
- Try to signin wrong to see an error notification
- Try to edit a project name or scenario name


pablopareja (changes requested):
Changes proposed here: https://github.com/Vizzuality/marxan-cloud/pull/80

PR #85, opened 2021-03-03
[MARXAN-127] [MARXAN-125] edit users

hotzevzl:
### Overview

**Once ready for review, this PR will still need to be rebased on `develop` before merging**

This PR adds:

* updating user metadata
* updating a user's own password (if they can provide their current password for validation)

### Designs

N/A

### Testing instructions

See e2e tests.

### Feature relevant tickets

https://vizzuality.atlassian.net/browse/MARXAN-127
https://vizzuality.atlassian.net/browse/MARXAN-129

---

- [x] Update CHANGELOG file


aagm:
After doing pull from the branch thigs started to be unnestable, i did run the make tests and i had 
Test Suites: 3 failed, 3 passed, 6 total
Tests:       14 failed, 23 passed, 37 total

aagm (changes requested):
Maybe again this is because things have slipped in this pr that shouldn't but I got some inconsistencies in projects (`project/dto/create.project.dto.ts`) and user modules (`users/users.controller.ts`) and tests failed

hotzevzl:
@aagm all tests are passing now in my local env when running them *against a running API instance loaded with GADM and WDPA data * (`docker-compose exec api yarn test:e2e`).

Please note that when running tests against your local instance/db, the user bb@example.com is actually "soft-deleted", so you will need to run `update users set is_active = true, is_deleted = false, password_hash = crypt('bbuserpassword', gen_salt('bf')) where email = 'bb@example.com';` after running the docker compose command above or successive tests involving this user will fail.

When running them via `make test-e2e-api` the tests for countries fail, but I think this must have something to do with data being loaded during the setup phase of this make target.

I think this is as good as it can get for this PR - indeed the users tests were failing because I was firing requests to get a fresh JWT token before each individual test, but this would not work when changing the password for a second test user (`bb@example.com`). I'll check the countries test in a future PR.

hotzevzl:
merging as discussed off-band - the issue left is 4 tests failing (for `CountriesModule`) in the CI workflow, which I'll look at in a different PR.

PR #86, opened 2021-03-04
Update the Modal and add two prompt components

clementprdhomme:
### Overview

This PR updates the modal to use the [React Aria](https://react-spectrum.adobe.com/react-aria/index.html) library and adds two prompt components: `AlertPrompt` and `ConfirmationPrompt`.

<p align="center">
AlertPrompt saying someone else is already editing the scenario. The user can click ok or request access.
</p>

### Designs

- [`AlertPrompt`](https://www.figma.com/file/hq0BZNB9fzyFSbEUgQIHdK/Marxan-Visual_V02?node-id=2991%3A20046)
- [`ConfirmationPrompt`](https://www.figma.com/file/hq0BZNB9fzyFSbEUgQIHdK/Marxan-Visual_V02?node-id=2991%3A3921)

In the designs, the colour of the grey button is lighter, though the `Button` component only provides a darker shade of grey. I suppose the designers added this variant of the button after we implemented the component.

If you think it's necessary, I see no problem in adding a third variant (`tertiary`) and using it in the prompts.

### Testing instructions

You can test the components in Storybook.


mbarrenechea (commented):
I feel unnecessary the AlertPrompt component. As soon as you start adding these functions `onClickSecondaryAction`, what happens if we have a third one?. 
I think we should use directly a Modal with children. But it's my opinion.

Good job BTW.

app/components/alert-prompt/component.tsx
```diff
@@ -0,0 +1,105 @@
+import React from 'react';
+import classnames from 'classnames';
+
+import Modal, { ModalProps } from 'components/modal';
+import Button from 'components/button';
+import Icon, { IconProps } from 'components/icon';
+
+export interface AlertPromptProps {
```
pablopareja, line 8: Would you mind moving this interface definition to a `types.ts` file in the local folder?

app/components/alert-prompt/component.tsx
```diff
@@ -0,0 +1,105 @@
...
+  description?: string;
+  /**
+   * Whether the prompt is diplayed
+   */
+  open: ModalProps['open'],
+  /**
+   * Whether the user can close the prompt without having to accept nor refuse
+   */
+  dismissible?: ModalProps['dismissable'],
+  /**
```
pablopareja, line 25: Great to see the properties documented 👏 
We should do the same in all components from now on 💯

app/components/confirmation-prompt/component.tsx
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
+import classnames from 'classnames';
+
+import Modal, { ModalProps } from 'components/modal';
+import Button from 'components/button';
+import Icon, { IconProps } from 'components/icon';
+
+export interface ConfirmationPromptProps {
```
pablopareja, line 8: Same thing about moving this to a `types.ts` file 🙏

pablopareja (changes requested):
I've added some minor code change requests to the code. Also, shouldn't we add an option to show a cancel/no button?
According to the current designs that's needed in some cases, for instance the modal to confirm the deletion of a project 👇 
image

Other than that, great work! 👏

clementprdhomme:
Well, if you have multiple alert prompts in the UI, you will need a component otherwise you'll duplicate code. I also don't believe there will more than two actions, but you're right, we never know.

I've removed the `AlertPrompt` component now.

clementprdhomme:
@pablopareja I moved the types as you suggested.

Regarding your general comment, I think you confused the `AlertPrompt` and `ConfirmationPrompt` components. The one you're looking for it `ConfirmationPrompt`. I deleted the other one anyway following Miguel's comment.

pablopareja (approved):

pablopareja:
Cool, thanks! 🙌 
Let's merge it 🔨

PR #87, opened 2021-03-05
Feature/data/output models

tamaramegan:
### Overview

The relevant notebooks to check are marxan_ouputs.ipynb and marxan_utils.ipynb. The marxan_ouputs notebook contains the validation models constructed for each of the output files (except for the sen.dat and log.dat). 

https://vizzuality.atlassian.net/browse/MARXAN-147
---

- [ ] Update CHANGELOG file


aagm (changes requested):
Before merging it let's do the changes we have discussed this morning; instead of a dynamic type for the solution matrix, it should be seen as a List of solutions. After that we can merge :D

tamaramegan:
The new OutputSolutionsMatrix model has been added. Right now, the model has 2 Fields: SolutionsMatrix (the Solution ID) and I have created a Dict with planning units and value (eg: {P1:0,P2:1,P3:1...}. Its a Dict instead of a List to keep the planning unit number associated to the result, but let me know if you would prefer to have another structure.

PR #88, opened 2021-03-08
PU grids generation.

aagm:
### Overview
Job system and first queue + Job to generate Planning units related to a user project area.
Also the testing system for QA this. 

### Testing instructions

Running Tests should pass.
Don't forget to do a clean slate; the bd updates PostGIS to 3.1 and requires refresh the data.
You need to create a Project and automatically a new job is created if the project provides a valid region or extent. That job is going to be processed by the geoprocessing and generate the PU grid. PU grid will only be attached at the scenario level. But the required geometries will be ready on the PU table to be accessible through the project extent.

### Feature relevant tickets

https://vizzuality.atlassian.net/browse/MARXAN-148

---

- [ ] If this PR adds a feature that should be tested for regressions when
- [x] Update CHANGELOG file


hotzevzl:
@aagm this is great! lots of nice things working together - I'm really liking BullMQ already 😅 

See my branch https://github.com/Vizzuality/marxan-cloud/tree/feature/api/MARXAN-148-generate-planing-unit-grids%2Bandrea-suggestions for some ideas on how to deal with the test issues you were having.

Do keep in mind that we should try (in my opinion) to limit our testing to making sure that we trigger and process things as we mean to and as API clients may expect (though these "contracts" are still very much in flux at this stage, of course, but I mean in principle), that behaviours that are not desired should not happen, and that we are "piloting" these job queues in a meaningful way, but we should be careful at the same time to not end up "testing that BullMQ and Redis work", which is - again in my opinion - a common issue when dealing with what are in practice integration tests.

But in any case, it's often a fine line, and I'm sure we'll develop a good sense of what makes sense to test and what we can assume that is tested elsewhere, as we keep working on these kinds of async contexts.

hotzevzl:
also, I didn't get to check why Jest hangs indefinitely if the queue cleanup is running on teardown (`afterAll()`) - so I have left that disabled via a bool constant for now, but we should eventually clean up our test data from Redis automatically on teardown, of course. but nothing to worry about for the moment.

api/test/planning-units.e2e-spec.ts
```diff
@@ -0,0 +1,220 @@
...
+      const response = await request(app.getHttpServer())
+        .post('/api/v1/organizations')
+        .set('Authorization', `Bearer ${jwtToken}`)
+        .send(E2E_CONFIG.organizations.valid.minimal())
+        .expect(201);
+
+      anOrganization = response.body.data;
+
+      expect(anOrganization.type).toBe('organizations');
+    });
```
hotzevzl, line 117: this is something I would discourage to take from my (bad) early examples in other tests 😅 

This is all opinion, but basically at the beginning I was creating "parent" entities in self-contained tests, but in practice we shouldn't be testing that creating an organization works as expected in a test suite focused on planning units - in newer tests I have used some test util classes to wrap organization/project/scenario creation and I then create these parent entities as needed in `beforeAll()`, and delete them if needed in `afterAll()`. I will adapt older tests bit by bit, so I am happy for this to stay here if you are ok for me to include this in my cleanup 😅  (same for the deleting projects/org test further down).

geoprocessing/test/planning-units-processor.e2e-spec.ts
```diff
@@ -0,0 +1,27 @@
...
+function delay(ms: number) {
+  return new Promise(function (resolve) {
+    return setTimeout(resolve, ms);
+  });
+}
+
+describe('planning units jobs (e2e)', () => {
+  jest.setTimeout(2 * 1000);
+  it('executes the child job processor with mock data', async () => {
+    const createPlanningUnitsDTO: Partial<Job> = {
```
hotzevzl, line 15: this test prevents the geoprocessing service to build for me.

without full context, I got this to build with a few type narrowings - see my suggestions about this in planning-units.job.ts and in the e2e config for geoprocessing

hotzevzl, line 15:
```suggestion
    const createPlanningUnitsDTO: Pick<Job<PlanningUnitsJob>, 'data' | 'id' | 'name'> = {
```

geoprocessing/test/e2e.config.ts
```diff
@@ -0,0 +1,48 @@
...
+  PlanningUnitsJob,
+  PlanningUnitGridShape,
+} from 'src/modules/planning-units/planning-units.job';
+
+export const E2E_CONFIG: {
+  planningUnits: {
+    creationJob: {
+      valid: {
+        customArea: (options: unknown) => Partial<PlanningUnitsJob>;
+        adminRegion: (options: unknown) => Partial<PlanningUnitsJob>;
```
hotzevzl, line 12:
```suggestion
        customArea: (options: unknown) => PlanningUnitsJob;
        adminRegion: (options: unknown) => PlanningUnitsJob;
```

geoprocessing/test/planning-units-processor.e2e-spec.ts
```diff
@@ -0,0 +1,27 @@
+import { Job } from 'bullmq';
+import defaultExport from '../src/modules/planning-units/planning-units.job';
```
hotzevzl, line 2:
```suggestion
import { createPlanningUnitGridFromJobSpec, PlanningUnitsJob } from '../src/modules/planning-units/planning-units.job';
```

geoprocessing/test/planning-units-processor.e2e-spec.ts
```diff
@@ -0,0 +1,27 @@
...
+  jest.setTimeout(2 * 1000);
+  it('executes the child job processor with mock data', async () => {
+    const createPlanningUnitsDTO: Partial<Job> = {
+      id: '1',
+      name: 'create-pu',
+      data: E2E_CONFIG.planningUnits.creationJob.valid.customArea({
+        countryCode: 'NAM',
+      }),
+    };
+    const value = await defaultExport(createPlanningUnitsDTO);
```
hotzevzl, line 22:
```suggestion
    const value = await createPlanningUnitGridFromJobSpec(createPlanningUnitsDTO);
```

geoprocessing/src/modules/planning-units/planning-units.job.ts
```diff
@@ -0,0 +1,72 @@
...
+  countryId?: string;
+  adminRegionId?: string;
+  adminAreaLevel1Id?: string;
+  adminAreaLevel2Id?: string;
+  planningUnitGridShape: PlanningUnitGridShape;
+  planningUnitAreakm2: number;
+  extent?: Record<string, unknown>;
+}
+
+export default async (job: Job<PlanningUnitsJob> | any) => {
```
hotzevzl, line 23:
```suggestion
export const createPlanningUnitGridFromJobSpec = async (job: Pick<Job<PlanningUnitsJob>, 'data' | 'id' | 'name'>) => {
```
just a tiny detail, but I recommend to name exports explicitly to help understand at a glance what they are - this suggestion is mainly about the type signature (see comments below for `planning-units-processor.e2e-spec.ts`) but I also suggest the naming thingy (and use that in the e2e-spec file suggestions)

hotzevzl (changes requested):
@aagm this is massive, I love this PR and the amazing amount of work behind it, well done!

as usual a lot of style notes, and a couple of change requests:

* I could not build the geoprocessing service without some kind of changes to the test that goes with it (I have added suggestions, in case you are ok with them and they work for you)
* logic and templated SQL in planning-units.job.ts can be made more robust and safer

for the latter, no need to handle all possible cases now, but I would suggest to handle at least the happy path more robustly and by using query params where possible (see notes inline)

geoprocessing/src/modules/planning-units/planning-units.job.ts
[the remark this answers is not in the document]
hotzevzl:
```suggestion
export default async (
  job: Pick<Job<PlanningUnitsJob>, 'data' | 'id' | 'name'>,
) => {
```

this "works for me" together with the changes I suggested in the planning units processor e2e test file

api/src/modules/planning-units/planning-units.service.ts
```diff
@@ -0,0 +1,53 @@
...
+  constructor() {
+    this.queueEvents.on('completed', (job) => {
+      this.logger.log(
+        `this job ${job.jobId} for ${this.queueName} is completed`,
+      );
+    });
+  }
+
+  public async create(creationOptions: CreatePlanningUnitsDTO): Promise<void> {
+    await this.planningUnitsQueue.add('create-pu', creationOptions);
```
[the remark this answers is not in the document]
aagm: Yes, that idea was floating on my mind, I decided not yet to do it as I will probably need your wisdom there, but definitely something to do in the near future.

api/src/modules/projects/projects.service.ts
```diff
@@ -31,6 +32,7 @@ type ProjectFilterKeys = keyof Pick<
 >;
 type ProjectFilters = Record<ProjectFilterKeys, string[]>;
 
+const logger = new Logger('projects');
```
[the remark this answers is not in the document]
aagm: ok that was a leftover from the rebase, thank you for catching it!

api/src/modules/projects/projects.service.ts
```diff
@@ -185,6 +188,17 @@ export class ProjectsService extends AppBaseService<
...
-    return planningArea;
+    return planningArea
+  }
+
+  async actionAfterCreate(
+    model: Project,
+    createModel: CreateProjectDTO,
+    info?: AppInfoDTO,
+  ): Promise<void> {
+    if (createModel.planningUnitAreakm2 && createModel.planningUnitGridShape) {
```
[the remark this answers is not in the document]
aagm: Ideally, it should be transparent, the reality is that right now it will only take care of regular generated grids (as we currently do not support user grid uploads). in my mind these are the possibilities:
An area(extent)/region/custom grid is not supplied -> nothing is triggered 
A custom extent (area) is supplied, but not planning unit area type + grid size or user grid -> nothing is triggered
A planning unit area type + grid size is supplied but  not an area(extent)/region/custom grid is provided -> nothing is triggered

An area(extent)/region is provided + a planning unit area type(regular grid hexagon or square or irregular) + grid size -> PU grid generation is triggered.

A custom grid uploaded by the user is provided -> Regular grid creation should not be triggered, but we should upload that grid as a custom one set the PlanningUnitGridShape.fromShapefile and infer the extent and /or admin regions from there. Also, have a way to retrieve that user data as part of the project. 

* the area extent can be also uploaded by the user but is only a geometry that represents a custom extent.

And yes the same logic should go in after the update, + a check to invalidate everything if this is triggered. 
And also we must do all this checks/validations before we trigger this.

But as you mention this looks like work for a next pr :D (i will put this exact thing in jira so we dont forget about all things that need to happen)

docker-compose.yml
```diff
@@ -91,18 +95,28 @@ services:
...
+  # Admin gui for redis
+  redis-commander:
+    image: rediscommander/redis-commander
+    hostname: redis-commander
+    container_name: marxan-redis-commander
+    restart: always
+    environment:
+      - REDIS_HOSTS=local:marxan-redis:6379
+    ports:
+      - '9000:8081'
```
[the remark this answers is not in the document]
aagm: You are right, I did add the commander may be in a rush mode in order to have a glimpse at the Redis instance, I will fix it setting an env with the port as the rest of the services :D

api/src/modules/planning-units/planning-units.service.ts
```diff
@@ -0,0 +1,53 @@
...
+ **/
+@Injectable()
+export class PlanningUnitsService {
+  public readonly queueName: string = 'planning-units';
+  private readonly logger: Logger = new Logger(
+    `${this.queueName}-queue-publisher`,
+  );
+  public readonly planningUnitsQueue: Queue = new Queue(this.queueName, {
+    ...config.get('redisApi'),
+    defaultJobOptions: {
```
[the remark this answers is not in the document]
aagm: I left it open but agree that this can be outed in the config default, moving it there.

geoprocessing/src/modules/planning-units/planning-units.job.ts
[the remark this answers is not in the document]
aagm: ok that any was for the tests, a lack of knowledge on my end :S as for testing, it complains that is not a full job definition.

geoprocessing/src/modules/planning-units/planning-units.job.ts
```diff
@@ -0,0 +1,72 @@
...
+        }
+        if (job.data.adminAreaLevel2Id?.length) {
+          filterSQL.push(`gid_2 = '${job.data.adminAreaLevel2Id}'`);
+        }
+        subquery = `(SELECT (${gridShape[job.data.planningUnitGridShape]}(${
+          Math.sqrt(job.data.planningUnitAreakm2) * 1000
+        },
+                            ST_Transform(a.the_geom, 3857))).*
+                    FROM admin_regions a
+                    WHERE ${filterSQL.join(' AND ')} ) grid`;
```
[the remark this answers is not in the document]
aagm: Yes you are right, that validation should happen in the API on the `actionAfterCreate / actionAfterUpdate` in `project.service` (I have added a small guard there but probably is far from perfect)
And I agree that this is far from being complete and perfect, as I'm not taking into account uploaded files.

geoprocessing/src/modules/planning-units/planning-units.job.ts
```diff
@@ -0,0 +1,72 @@
...
+      let subquery: string;
+      const gridShape = {
+        square: 'ST_SquareGrid',
+        hexagon: 'ST_HexagonGrid',
+      };
+      if (job.data.extent) {
+        subquery = `(SELECT (${gridShape[job.data.planningUnitGridShape]}(${
+          Math.sqrt(job.data.planningUnitAreakm2) * 1000
+        },
+                            ST_Transform(ST_GeomFromGeoJSON('${JSON.stringify(
```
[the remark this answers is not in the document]
aagm: ok I have tested and so far, if the users try to pass a negative value the DB on project creation throws an error (i love enforcing uint in the DB table) so from there we should be safe. but I had also add at the job level a check and an error trow just in case

api/test/planning-units.e2e-spec.ts
aagm: i know but i was pending that the new ones are in order to incorporate them there. i can leave a todo note and we can refactor this tests later

geoprocessing/src/modules/planning-units/planning-units.job.ts
hotzevzl: sorry, this is confusing, I just realized this now: it was an initial suggestion, then the PR review page froze and when I reloaded it this comment wasn't shown, and in the meanwhile I had checked it locally and realised that naming the export was leading to the function not being called, so I left this alone thinking that the comment had been lost, but no, the part about naming the export should be ignored (and the part about the type signature of the job parameter could be considered, if you want). sorry about the confusion!

geoprocessing/test/planning-units-processor.e2e-spec.ts
aagm: ok i think this is solved now

geoprocessing/test/planning-units-processor.e2e-spec.ts
aagm: :+1:

geoprocessing/test/planning-units-processor.e2e-spec.ts
aagm: +1

geoprocessing/src/modules/planning-units/planning-units.job.ts
aagm: thank you for this :D

hotzevzl (approved):
@aagm thanks, I think we can merge 🚀 if you are happy.

My only remaining concerns are around unescaped content from user input in `geoprocessing/src/modules/planning-units/planning-units.job.ts`. This shouldn't be an issue when API requests come from the frontend, but in general we should avoid at all costs unescaped user input to end up in queries - it's too easy to overlook some validation and things can slip through. We can revisit this in a successive PR but let's definitely not forget about this.

Also, I'd suggest to stick to using error objects: line 77 in the same file - I'd suggest to `throw new Error('Without extent or valid administrative level a regular pu area cannot be created');` there.

PR #96, opened 2021-03-10
Feature/project item delete [MARXAN-101]

mbarrenechea:
### Overview

This PR adds:
- Click on the delete button shows a prompt that allows you to delete a project


pablopareja (commented):
I'm not sure if it's something expected at this stage of the implementation process but, it seems that projects that have scenarios associated to them cannot be removed 👇 
image
image

mbarrenechea:
> I'm not sure if it's something expected at this stage of the implementation process but, it seems that projects that have scenarios associated to them cannot be removed 👇
> image
> image
Yes. it's something we are aware of => https://vizzuality.atlassian.net/browse/MARXAN-101

PR #99, opened 2021-03-11
configure allowed included entities + add way to document this in OpenAPI docs [MARXAN-208]

hotzevzl:
See https://vizzuality.atlassian.net/browse/MARXAN-208


aagm (approved):
The only thing that i miss is a way for the api user to know which are the available include relationships on each module. Apart from that i've been able to properly fetch relationships in projects related users and scenarios

hotzevzl:
> The only thing that i miss is a way for the api user to know which are the available include relationships on each module. Apart from that i've been able to properly fetch relationships in projects related users and scenarios.

We do have support for this!

Screenshot_2021-03-18 Swagger UI

The OpenAPI output is configured here:
https://github.com/Vizzuality/marxan-cloud/pull/99/files#diff-8004824776c1bfc398aac2199f95ff630b5396ffcf741c1376fa9ecb8d10d26eR18

And the actual includes are configured here (for example in the `OrganizationsModule`):
https://github.com/Vizzuality/marxan-cloud/pull/99/files#diff-e91240c8879096abe6a97f2b8844c2334218db945b3ffb8fcec867f608578c6eR13
and
https://github.com/Vizzuality/marxan-cloud/pull/99/files#diff-31c1831c70abddff8ac7ae1d5cf8e41d9fcc3fdf16676b41ff622909c41a35e5R60

What I don't super like is that for the serializer configuration we have strong typing for an entity's own attributes:
https://github.com/Vizzuality/marxan-cloud/pull/99/files#diff-2b5d0b429abc297338bc221e8ead6e8a79e22113a0c29eb0050afddeb3b29f1cR37
but not for those of included entities:
https://github.com/Vizzuality/marxan-cloud/pull/99/files#diff-2b5d0b429abc297338bc221e8ead6e8a79e22113a0c29eb0050afddeb3b29f1cR40

But this is a bit more complicated...

PR #101, opened 2021-03-16
event module [MARXAN-124]

hotzevzl:
See https://vizzuality.atlassian.net/browse/MARXAN-124


aagm (approved):
The changelog is the only thing I miss, but apart from that, the test runs as expected and I was able to play with the events module. I think this is ready to be merge!

hotzevzl:
@aagm thanks, changelog is now updated, going to merge! 🚀

PR #107, opened 2021-03-23
allow to specify includes in singular get endpoints; add base filters for scenarios, projects and organizations [MARXAN-212] [MARXAN-213]

hotzevzl:
See https://vizzuality.atlassian.net/browse/MARXAN-212 and https://vizzuality.atlassian.net/browse/MARXAN-213

### DIFF
--- a/api/CHANGELOG.md
+++ b/api/CHANGELOG.md
@@ -6,9 +6,27 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic
 Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.2.0]
+
+## [0.3.0]
 
 Unreleased
+
+### Added
+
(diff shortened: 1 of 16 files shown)


aagm (commented):
Ok includes for the specific relations works for me; said that users retrieve me an empty array; not sure if that should be the case. Apart from that :tada:

hotzevzl:
> Ok includes for the specific relations works for me; said that users retrieve me an empty array; not sure if that should be the case. Apart from that tada

yes and no - the array should be empty because we don't actually have any code that associates users directly to other entities right now, except a few `INSERTS` in the test data seed that gives users roles on seed organizations and projects...

In this phase we'll only support granting users roles on organizations, but my plan is to create implicit roles on projects (maybe not scenarios) that are part of each organization, so this property may eventually get populated for projects too. For now it is expected for the `users` prop to always be empty.

PR #118, opened 2021-04-07
docker build for app

hotzevzl:
see https://vizzuality.atlassian.net/browse/MARXAN-207


hotzevzl:
@mbarrenechea this is as far as I got earlier this week.

The multi-stage Docker build lifted pretty much verbatim from the nextjs docs works fine, but I am missing something as the built tree in the Docker image doesn't have all I expect, and in fact when starting the app via `docker-compose up --build app` I see the error message below which to my n00b eye sounds like there's something my build setup is not building correctly.

```
marxan-app            | internal/modules/cjs/loader.js:818
marxan-app            |   throw err;
marxan-app            |   ^
marxan-app            |
marxan-app            | Error: Cannot find module '/app/start'
marxan-app            |     at Function.Module._resolveFilename (internal/modules/cjs/loader.js:815:15)
marxan-app            |     at Function.Module._load (internal/modules/cjs/loader.js:667:27)
marxan-app            |     at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:60:12)
marxan-app            |     at internal/main/run_main_module.js:17:47 {
marxan-app            |   code: 'MODULE_NOT_FOUND',
marxan-app            |   requireStack: []
marxan-app            | }
marxan-app exited with code 1
```

PR #123, opened 2021-04-14
GeoFeaturesModule [MARXAN-155]

hotzevzl:
See https://vizzuality.atlassian.net/browse/MARXAN-155


geoprocessing/src/migrations/geoprocessing/1618853324000-ExtractDistictPropsFromFeatures.ts
```diff
@@ -0,0 +1,50 @@
...
+    all_properties_across_feature_rows := (select properties from features_data where feature_id = input limit 1);
+    FOR _key, _value IN
+      SELECT * FROM jsonb_each_text(all_properties_across_feature_rows)
+    LOOP
+      result := result || jsonb_build_object(_key, array(select distinct properties -> _key from features_data where feature_id = input));
+    END LOOP;
+
+    RETURN result;
+  END
+$f$;
```
hotzevzl, line 30: @aagm as much as I love pg functions, I also tend to agree with folks who warn about their source control to be a bit of a parallel world to the git context of the other components, because of the indirection layer of migrations. For a few, rarely changing (I guess... 😅) functions this all looks fine to me, in any case.
However, would you have suggestions about adding some brief documentation to these pg functions? Should we add anything to the TS source in the migration? or as SQL comments within the SQL string that we pass to `queryRunner.query()`?
Again, not really a big deal here as the function is very small and rather plain and hopefully understandable at a glance, but I am open to suggestions.

aagm (approved):
we will merge the pr as is but we have agree on the next improvements:

- add the geographical filter on the geoprocessing query, and drop the view
- move the properties keys for each feature at the ingestion level creating a new jsonb column on the features table
- create the bbox field at project level with its require triggers.
- rework a bit the procedure so we can simplify the operations perform now that the property keys will be calculated at ingestion level.

------
Apart from this we have agree on a feature recipe functionality and how is that going to be define at job level arising also some issues for the future of the project  related huge ammount of data ingested that we do expect.

PR #129, opened 2021-04-21
Feature/api/marxan 90 add tests for data shape error responses

alexeh:
### Overview

E2E Test Coverage to check error responses's shape follows JSON:API Specification

Unit Test Coverage for `all-exceptions.exception.filter.ts`
### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


api/test/json-api.e2e-spec.ts
```diff
@@ -0,0 +1,44 @@
...
+      },
+    };
+    const response = await request(app.getHttpServer())
+      .post('/auth/sign-in')
+      .send({
+        username: 'fakeuser@example.com',
+        password: 'fakePassword',
+      });
+    response.body.errors.forEach((err: any) => {
+      expect(Object.keys(err)).toEqual(Object.keys(jsonApiErrorResponse));
```
hotzevzl, line 38: All looks good to me in this PR, thanks! my only question would be about this bit specifically.

Here we check that the root props of the error response object(s) match the list of three props in the `jsonApiErrorResponse` const above - which is fine. Since the JSON:API spec also allows to include `id`, `links`, `code`, `detail` and `source`, do we want to "whitelist" them in this test? This way, if we added allowed props later on to our response, the test will still pass (and correctly so) without needing to change it.

I am ok with the current implementation, just checking what your opinion would be on this.

api/test/json-api.e2e-spec.ts
```diff
@@ -0,0 +1,44 @@
...
+      expect(Object.keys(err.meta)).toEqual(
```
hotzevzl, line 39: One other bit that we may want to test (I would be fine with *not* testing it) is whether we keep removing (omitting) the `meta.rawError` and `meta.stack` props *if* `NODE_ENV !== 'development' - aka we don't make changes that accidentally break what we have in place to not leak this data outside of development environments.

hotzevzl (approved):
well done, thank you @alexeh!

I am happy for you to merge this as is - I have left two notes for your consideration.

Also, `yarn test` will currently show one failure because of an empty spec (`src/app.controller.spec.ts`) - this is unrelated to the aim of this PR, but consider if you want to remove that empty spec file altogether to get all green in Jest - I don't think we'll ever need it (and we could probably even remove `src/app.controller.ts` altogether), but consider if you want to do this small cleanup as part of this PR.

api/test/json-api.e2e-spec.ts
alexeh: Yes, I totally agree on this one. I am not sure if Jest allow us to set NODE_ENV safely from test to test (it will always set it to 'test' by default when setting its universe up, but we can set a custom one in a conf file), because I think that forcing process.env from one test to another it is not a good idea, but I will have a look at it!

api/test/json-api.e2e-spec.ts
alexeh: Yes I see your point. And I'll be happy to add this whitelist, but I didn't at the first place and here is the reasoning behind. I think that testing needs to be accurate to what logic/functionality we are covering with it, so if our production code does not send said properties I we shouldnt test that something 'might exist or not', but in the other hand, we are dealing with a standar specification and not an open interpretation of what we want to send, and suddenly does not seem that bad.
I'll add this one, thanks!

api/test/json-api.e2e-spec.ts
hotzevzl: sorry, I had misread this comment - what I meant was to actually do test both conditions by fiddling with the env var, but I am actually not sure this works as intended if done in a `beforeAll()` for a `describe()` set.

I have done this in another project, but I think I ended up relying on a different strategy and not setting this to different values within the test (it was not `NODE_ENV`, just an app-specific var).

Be just mindful that if running these tests with something like `docker-compose run -e NODE_ENV=production api yarn test:e2e json` the test will fail:

```
 FAIL  test/json-api.e2e-spec.ts (12.634 s)
  JSON API Specs (e2e)
    ✕ should return a response shaped as JSON:API Error spec (68 ms)

  ● JSON API Specs (e2e) › should return a response shaped as JSON:API Error spec

    expect(received).toEqual(expected) // deep equality

    Expected: ArrayContaining ["timestamp", "path", "type", "rawError", "stack"]
    Received: ["status", "title", "meta"]

      47 |       Logger.debug(`NODE_ENV: ${process.env.NODE_ENV}`);
      48 |       if (process.env.NODE_ENV != 'development') {
    > 49 |         expect(Object.keys(err)).toEqual(
         |                                  ^
      50 |           expect.arrayContaining(Object.keys(jsonApiErrorResponse.meta)),
      51 |         );
      52 |       }

      at json-api.e2e-spec.ts:49:34
          at Array.forEach (<anonymous>)
      at Object.<anonymous> (json-api.e2e-spec.ts:43:26)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 1 total
Snapshots:   0 total
Time:        12.734 s
Ran all test suites matching /json/i.
```

hotzevzl: I was hungry and in a rush, I should have checked before suggesting running things with different `NODE_ENV` settings without checking 🤦🏼 

I have pushed a discardable branch with a very quick&dirty suggestion of a possible way to do this.

To be clear, all of this is really minor - I was happy with your original PR!

I agree that in general we shouldn't tie test behaviour to variables out of our control (such as the `NODE_ENV` set on the parent process). For the specific bit "we may not want to disclose stack traces and other details in environments other than `development`, my idea is however really to *take control* of the variable and make sure that if we run in something else than `development`, we trim out the error details we don't want to leak.

Don't feel you have to do this, and feel free to disagree of course, in any case 😅 .

I would say though that the failure I highlighted in my comment above from earlier today should be reviewed.

hotzevzl (approved):

api/test/json-api.e2e-spec.ts
hotzevzl: One last note after we've now merged this PR 🚀 - going back to my past work involving envvars, I think the setup I had suggested in my suggestions branch would not work in any case and that the way to go would be to use https://jestjs.io/docs/configuration#setupfiles-array and then in one such setupfile set the envvars that are needed.

IIRC from my past work on this, however one fiddles with the envvars within the tests themselves (even running `[...].createNestApplication()` again after fiddling with the envvar), the `NODE_ENV` set when Jest is started is the only one that gets used. The issue then is how to *conditionally set* the envvar in a setupfile, but well, this can be a topic for another day if we ever need to do that 😅

PR #130, opened 2021-04-26
marxan 345 add feature class name aliases filters

alexeh:
### Overview

Allow user to add a **q** query parameter to filter by **featureClassName** and **alias** 

### Designs

### Feature relevant tickets

--- 

- [ ] Update CHANGELOG file


api/test/geo-features.e2e-spec.ts
```diff
@@ -49,6 +50,10 @@ describe('GeoFeaturesModule (e2e)', () => {
   const country = 'NAM';
   const l1AdminArea = 'NAM.13_1';
   const l2AdminArea = 'NAM.13.5_1';
+  const geoFeaturesFilters = {
+    cheeta: { featureClassName: 'iucn_acinonyxjubatus', alias: 'cheetah' },
```
kgajowy, line 54: is this mapping specific / well known? Where `featureClassName` comes from?

hotzevzl:
@alexeh I'm taking the liberty to move here the conversation you started in the project channel:

>  I'd like to to do something like this:
> ```
> @Get(':projectId/features')
>   async findAllGeoFeaturesForProject(
>     @ProcessFetchSpecification() fetchSpecification: FetchSpecification,
>     @Param() params: { projectId: string },
>     @Query('q') queries: { featureAliasFilter: string },
>   ): Promise<GeoFeatureResult> {
>     const results = await this.geoFeaturesService.findAllPaginated(
>       fetchSpecification,
>       { params, queries },
>     );
>     return this.geoFeaturesService.serialize(results.data, results.metadata);
>  }
> ```
> But this is not possible due to typing, and it will need to either update `nestjs-base-service` or add a new type, which not sure if > worth the trouble. Thoughts?

So, firstly - what is the intent here? I think we should handle the legacy of the `InfoDTO` that was bestowed upon us by wise elders and adapt it to our needs (or discard it if not useful). My guess of the general intent is that it would have been used for endpoint or module-specific params and "other stuff", rather than things like FetchSpecification which are common to most endpoints.

I think we may need a few more of these "partial match over one or more fields" parameters in this project, but I'd wait until we need it at least a second time in practice to make it into a common pattern.

Said that, I think this should work:

```
  @Get(':projectId/features')
  async findAllGeoFeaturesForProject(
    @ProcessFetchSpecification() fetchSpecification: FetchSpecification,
    @Param() params: { projectId: string; featureClassAndAliasFilter?: string },
    @Query() featureClassAndAliasFilter: { q: string},
  ): Promise<GeoFeatureResult> {
    /**
     * @debt Add proper typing in nestjs-base-service to allow adding query parameters in InfoDTO
     */

    params.featureClassAndAliasFilter = featureClassAndAliasFilter.q;
    const results = await this.geoFeaturesService.findAllPaginated(
      fetchSpecification,
      { params },
    );

    return this.geoFeaturesService.serialize(results.data, results.metadata);
  }
```

Or you can extend AppInfoDTO via type intersection (e.g. `export type AppInfoDTO = InfoDTO<User> & { otherStuff: ... };`), but I'd go here later, if we want to have a common property for this kind of usage - but then, we should probably try to move it to nestjs-base-service, I suppose.

api/test/geo-features.e2e-spec.ts
hotzevzl: this may come from source data from the ETL pipelines; its format and content may depend on the data source

api/test/geo-features.e2e-spec.ts
```diff
@@ -168,6 +173,62 @@ describe('GeoFeaturesModule (e2e)', () => {
       test.todo(
         'As a user, when I upload feature shapefiles, I should see the related features in the list of those available within a project',
       );
+
+      test('should return a single result of geo-features which its className matches a given filter', async () => {
```
hotzevzl, line 177:
```suggestion
      test('should return a single result of geo-features whose className property matches a given filter', async () => {
```

api/test/geo-features.e2e-spec.ts
```diff
@@ -168,6 +173,62 @@ describe('GeoFeaturesModule (e2e)', () => {
...
+
+        console.log(response.body.data);
+
+        expect(response.body.data).toHaveLength(1);
+        expect(response.body.data[0].attributes.featureClassName).toEqual(
+          geoFeaturesFilters.cheeta.featureClassName,
+        );
+      });
+
+      test('should return a single result of geo-features which its alias matches a given filter', async () => {
```
hotzevzl, line 193: same as above regarding test description and `console.log()`

api/test/geo-features.e2e-spec.ts
```diff
@@ -168,6 +173,62 @@ describe('GeoFeaturesModule (e2e)', () => {
...
+          .expect(HttpStatus.OK);
+
+        console.log(response.body.data);
+
+        expect(response.body.data).toHaveLength(1);
+        expect(response.body.data[0].attributes.alias).toEqual(
+          geoFeaturesFilters.cheeta.alias,
+        );
+      });
+      test('should return a list of geo-features which its alias matches a given filter', async () => {
```
hotzevzl, line 208:
```suggestion
      test('should return a list of geo-features whose alias or featureClassName properties match a given substring', async () => {
```

api/test/geo-features.e2e-spec.ts
```diff
@@ -168,6 +173,62 @@ describe('GeoFeaturesModule (e2e)', () => {
...
+            `/api/v1/projects/${aProjectWithCountryAsPlanningArea.id}/features?q=${geoFeaturesFilters.partialMatches.us}`,
+          )
+          .set('Authorization', `Bearer ${jwtToken}`)
+          .expect(HttpStatus.OK);
+
+        /**
+         * Here we just assert the result length since result match can come either from aliases or
+         * featureClassNames, and we have control over stored test data
+         */
+        expect(response.body.data).toHaveLength(7);
```
hotzevzl, line 220: we may need to update this count if we cannot get any matches on the `alias` column with the current test data

api/test/geo-features.e2e-spec.ts
```diff
@@ -168,6 +173,62 @@ describe('GeoFeaturesModule (e2e)', () => {
...
+      });
+      test('should return all available features if query param has no value', async () => {
+        const response = await request(app.getHttpServer())
+          .get(
+            `/api/v1/projects/${aProjectWithCountryAsPlanningArea.id}/features?q=`,
+          )
+          .set('Authorization', `Bearer ${jwtToken}`)
+          .expect(HttpStatus.OK);
+
+        expect(response.body.data).toHaveLength(9);
```
hotzevzl, line 230: same as above

api/src/modules/projects/projects.controller.ts
```diff
@@ -68,12 +69,19 @@ export class ProjectsController {
...
+    /**
+     * @debt Add proper typing in nestjs-base-service to allow adding query parameters in InfoDTO
+     */
+
+    params.featureClassAndAliasFilter = featureClassAndAliasFilter;
     const results = await this.geoFeaturesService.findAllPaginated(
       fetchSpecification,
       { params },
     );
+
```
hotzevzl, line 84: for this, see comments by myself and Kamil in this PR's general discussion if you want to refactor this a bit.

alexeh: So I decided to let that query param as is for simplicity and just add it to params object applying immutability according to @kgajowy 's suggestion

api/test/geo-features.e2e-spec.ts
alexeh: Count updated

api/test/geo-features.e2e-spec.ts
alexeh: No need for updating this one since an empty query param will return all available features

PR #135, opened 2021-04-27
chore(api): fix pending issues of strict settings

kgajowy:
### Overview

Please pay close attention during review of fields, especially those within `Entities`, as I may not have a full knowledge of the context/scope. I tried to match `!`/`?` according to existing swagger definitions.

Added `geojson` types, however keep in mind that the input is far from being validated (yet) when consumer hits the API. It works fine however within code:
image

Regarding `!` operator for DTO classes - it may look bad but there aren't any other ways to get this issue down. Usually DTO/Entities aren't created by hand anyway.

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


hotzevzl:
@kgajowy thank you. I'm going to be cross-checking the fields as you asked, other than this the PR looks good to me. I am not too bothered by the definite assignment assertions even when writing entities and DTOs by hand, and it's straightforward to add  however, do you think we gain something valuable here vs having `"strictPropertyInitialization": false` in `tsconfig.json`? Either way we're blanket-silencing the property initialization checker, so we're removing a safety net one way or another, but most importantly, I can't think of cases where not having this safety net so far has been an issue in practice.

api/src/modules/projects/dto/create.project.dto.ts
```diff
@@ -54,18 +56,20 @@ export class CreateProjectDTO {
...
   @IsOptional()
   @IsObject()
-  extent?: Record<string, unknown>;
+  extent?: MultiPolygon | Polygon;
 
   @ApiPropertyOptional()
   @IsOptional()
   @IsObject()
-  metadata?: Record<string, unknown>;
+  metadata?: Record<string, string>;
```
hotzevzl, line 74: I think this could stay as `Record<string, unknown>` for the moment (we're not using it IIRC and if we do start using it we may want to have more precise typing for this prop's values, but just in case, I'd leave this open for the time being).

api/src/modules/projects/project.api.entity.ts
```diff
@@ -119,7 +119,7 @@ export class Project extends TimeUserEntityMetadata {
    */
   @ApiPropertyOptional()
   @Column('geometry')
-  extent: Record<string, unknown> | null;
+  extent?: Record<string, unknown> | null;
```
hotzevzl, line 122:
```suggestion
  extent?: Record<string, unknown>;
```

api/src/modules/projects/dto/create.project.dto.ts
hotzevzl:
```suggestion
  metadata?: Record<string, unknown>;
```

kgajowy:
> @kgajowy thank you. I'm going to be cross-checking the fields as you asked, other than this the PR looks good to me. I am not too bothered by the definite assignment assertions even when writing entities and DTOs by hand, and it's straightforward to add however, do you think we gain something valuable here vs having `"strictPropertyInitialization": false` in `tsconfig.json`? Either way we're blanket-silencing the property initialization checker, so we're removing a safety net one way or another, but most importantly, I can't think of cases where not having this safety net so far has been an issue in practice.

DTO/Entities are kind of "special" where, in most cases, are being initialized outside of our "common" flow (for example, by TypeORM).
However we may have "normal" classes, even common injectable services, where we can "forgot" to assign class field. Therefore, enabling `"strictPropertyInitialization": false` in `tsconfig` may give us troubles. If this can be covered by eslint for only files with ext. of `.dto` and `.entity`, such disable would be ok!

api/src/modules/projects/project.api.entity.ts
hotzevzl: taking the chance to remove older cruft 😱

api/src/modules/projects/project.api.entity.ts
```diff
@@ -128,40 +128,40 @@ export class Project extends TimeUserEntityMetadata {
    */
   @ApiPropertyOptional()
   @Column('jsonb')
-  metadata: Dictionary<string>;
+  metadata?: Dictionary<string>;
```
hotzevzl, line 131: again, older cruft
```suggestion
  metadata?: Record<string, unknown>;
```

api/src/modules/users/user.api.entity.ts
```diff
@@ -25,23 +25,23 @@ export const userResource: BaseServiceResource = {
...
 
   @ApiProperty()
   @Column('character varying')
-  email: string;
+  email!: string;
 
   @ApiPropertyOptional()
   @Column('character varying', { name: 'display_name' })
-  displayName: string | null;
+  displayName?: string | null;
```
hotzevzl, line 36:
```suggestion
  displayName?: string;
```

api/src/modules/users/user.api.entity.ts
```diff
@@ -25,23 +25,23 @@ export const userResource: BaseServiceResource = {
...
 
   @ApiPropertyOptional()
   @Column('character varying')
-  fname: string | null;
+  fname?: string | null;
```
hotzevzl, line 40:
```suggestion
  fname?: string;
```

api/src/modules/users/user.api.entity.ts
```diff
@@ -25,23 +25,23 @@ export const userResource: BaseServiceResource = {
...
 
   @ApiPropertyOptional()
   @Column('character varying')
-  lname: string | null;
+  lname?: string | null;
```
hotzevzl, line 44:
```suggestion
  lname?: string;
```

api/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "compilerOptions": {
     "module": "commonjs",
+    "strict": true,
```
hotzevzl, line 4: we can remove `noImplicityAny` and `strictNullChecks`

hotzevzl (commented):
@kgajowy thanks! I have only suggested a few changes mostly to remove older cruft while we are touching these props

api/src/modules/users/user.api.entity.ts
VA-DominikOstrowski: AFAIR TypeORM returns nulls

Dyostiq: AFAIR TypeORM returns nulls

hotzevzl:
@kgajowy ok, let's keep strict property initialization - see if you want to include any of the cleanup changes I suggested, and we're then good to merge, thanks!

kgajowy:
> @kgajowy ok, let's keep strict property initialization - see if you want to include any of the cleanup changes I suggested, and we're then good to merge, thanks!

sure, will do soon amend some of them - except `| null`:

I suppose we shouldn't do that.

If there isn't any value, TypeORM returns null (like NULL value from column) thus such change could lead to unpredicted issues

https://github.com/typeorm/typeorm/issues/5368
https://github.com/typeorm/typeorm/issues/4139

and possibly others.

api/tsconfig.json
kgajowy: Fixed in https://github.com/Vizzuality/marxan-cloud/pull/135/commits/d8ef02663ae149b88043f29404e4f763f1ad337e

kgajowy:
@hotzevzl current comments addressed, if everything else if fine, we can proceed to merge from my side (can't do it by myself, CI blocking)

PR #136, opened 2021-04-28
Fix issue with TypeORM loading entities from the dist folder

tiagojsag:
In some scenarios, when running the api natively, TypeORM will load the "compiled" entities (.js file from the dist folder) and thus will fail to load some entities.

### DIFF
--- a/api/src/ormconfig.ts
+++ b/api/src/ormconfig.ts
@@ -25,7 +25,7 @@ export const apiConnections: {
     type: 'postgres',
     url: AppConfig.get('postgresApi.url'),
     ssl: false,
-    entities: [__dirname + '/modules/**/*.api.entity.ts'],
+    entities: [__dirname + '/modules/**/*.api.entity.{ts,js}'],
     // Logging may be: ['query', 'error', 'schema', 'warn', 'info', 'log'] Use
     // 'query' if needing to see the actual generated SQL statements (this should
     // be limited to `NODE_ENV=development`). Use 'error' for least verbose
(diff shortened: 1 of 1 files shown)


kgajowy (approved):

hotzevzl (approved):
@tiagojsag I am fine with this if it works - all still seems to work ok within Docker too, but I don't understand how running this natively (`yarn start`? `yarn start:dev`?) ends up loading entities from the `dist` folder, nor how the fix helps since it should basically be broadening the range of files from which entities may be loaded rather than narrowing it, or looking in the correct base folder to start with, since this seems to be the core of the issue.

tiagojsag:
I also don't know why it magically loads things from /dist instead of the "normal" folder. as to why this works, the /dist folder will have a similar structure to the "normal" folder, just that is has a .d.ts and a .js file for each .ts file. This is why the change works.

This sort of issue with typeorm is quite ubiquitous in stackoverflow and the relevant gh repos, with multiple people reporting this solution as one that works for them. I understand this is an issue only I've experienced so far, but I also don't see any downside to the solution I'm providing, which is why I'm pitching it. If you see any inconvenient, I'm happy to drop this

hotzevzl:
Aaaah ok, I understand now - I had misread your original problem statement as if the problem was that transpiled entities were actually being loaded and you wanted to avoid that. Ok, sure. Now I understand the fix :)

And it will actually be needed if we run the Marxan API from transpiled source - in other projects we settled on using `ts-node` after drinking the kool aid that it doesn't introduce noticeable overhead, but it's good to keep options open.

PR #141, opened 2021-04-29
fix missing meta prop for pagination

hotzevzl:
See https://vizzuality.atlassian.net/browse/MARXAN-374

### DIFF
--- a/api/src/utils/app-base.service.ts
+++ b/api/src/utils/app-base.service.ts
@@ -83,7 +83,7 @@ export abstract class AppBaseService<
   ): Promise<any> {
     const serializer = new JSONAPISerializer.Serializer(this.pluralAlias, {
       ...this.serializerConfig,
-      paginationMeta,
+      meta: paginationMeta,
     });
 
     return serializer.serialize(entities);


alexeh (approved):

kgajowy (approved):
We miss you typings! 😅

hotzevzl:
> We miss you typings! sweat_smile

Totally. I'm fully guilty here of having added `@types/jsonapi-serializer` only to zap it myself. The reason is explained in `api/src/types/jsonapi-serializer.stub.d.ts` - I didn't have a chance back at that time to think if what was missing from the `@types/jsonapi-serializer` declarations could be achieved easily (probably contributing this back as a PR to the definitelytyped repo).

If you have thoughts on that, I'd welcome your opinion 😃

kgajowy:
> > We miss you typings! sweat_smile
> 
> Totally. I'm fully guilty here of having added `@types/jsonapi-serializer` only to zap it myself. The reason is explained in `api/src/types/jsonapi-serializer.stub.d.ts` - I didn't have a chance back at that time to think if what was missing from the `@types/jsonapi-serializer` declarations could be achieved easily (probably contributing this back as a PR to the definitelytyped repo).
> 
> If you have thoughts on that, I'd welcome your opinion 😃

@hotzevzl I can take a quick look at it (previously I did some locally-extending existing types) if only you could provide some example code what we expect to work

PR #142, opened 2021-04-29
make time/user metadata nullable

hotzevzl:
See https://vizzuality.atlassian.net/browse/MARXAN-342

### DIFF
--- a/api/src/migrations/api/1619711501000-DropNotNullsOnTimeUserMetadataColumns.ts
+++ b/api/src/migrations/api/1619711501000-DropNotNullsOnTimeUserMetadataColumns.ts
@@ -7,8 +7,8 @@ export class DropNotNullsOnTimeUserMetadataColumns1619711501000
 
   public async up(queryRunner: QueryRunner): Promise<void> {
     await Promise.all(
-      this.tablesToAlter.map(async (table) => {
-        await queryRunner.query(`
+      this.tablesToAlter.map(table => {
+        queryRunner.query(`
 ALTER TABLE ${table}
   ALTER COLUMN created_at DROP NOT NULL,
   ALTER COLUMN last_modified_at DROP NOT NULL,
--- a/api/src/migrations/api/1619711501000-DropNotNullsOnTimeUserMetadataColumns.ts
+++ b/api/src/migrations/api/1619711501000-DropNotNullsOnTimeUserMetadataColumns.ts
@@ -1,0 +1,50 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+import { v4 } from 'uuid';
+
+export class DropNotNullsOnTimeUserMetadataColumns1619711501000
+  implements MigrationInterface {
+  private tablesToAlter = ['organizations', 'projects', 'scenarios'];
+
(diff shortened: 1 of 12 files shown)


api/src/migrations/api/1619711501000-DropNotNullsOnTimeUserMetadataColumns.ts
```diff
@@ -0,0 +1,50 @@
...
+INSERT INTO users
+  (email, password_hash, display_name, is_active)
+VALUES
+  (${dummyUserEmailAddress}, crypt('${v4()}', gen_salt('bf')), 'Default owner for data imported via ETL jobs', false);
+`);
+    await Promise.all(
+      this.tablesToAlter.map(async (table) => {
+        await queryRunner.query(`
+UPDATE ${table}
+  SET created_by = (SELECT id FROM users WHERE email = '${dummyUserEmailAddress}')
```
kgajowy, line 39: Shouldn't it set `created_at` and `last_modified_at` as well? (might be default now() as well).

api/src/migrations/api/1619711501000-DropNotNullsOnTimeUserMetadataColumns.ts
```diff
@@ -0,0 +1,50 @@
...
+    await Promise.all(
+      this.tablesToAlter.map(async (table) => {
+        await queryRunner.query(`
+ALTER TABLE ${table}
+  ALTER COLUMN created_at DROP NOT NULL,
+  ALTER COLUMN last_modified_at DROP NOT NULL,
+  ALTER COLUMN created_by DROP NOT NULL;
+`);
+      }),
+    );
```
kgajowy, line 18: [minor/style]
```
await Promise.all(
      this.tablesToAlter.map((table) =>
        queryRunner.query(`
ALTER TABLE ${table}
ALTER COLUMN created_at DROP NOT NULL,
ALTER COLUMN last_modified_at DROP NOT NULL,
ALTER COLUMN created_by DROP NOT NULL;`),
      ),
    );
```

api/src/migrations/api/1619711501000-DropNotNullsOnTimeUserMetadataColumns.ts
hotzevzl: yep, the down side of this migration doesn't make fully sense in any case as it makes up data (owner and timestamps) that we cannot have if we haven't been gathering them - timestamps will default to `now()`. I don't envisage this side of the migration to have more than a formal value - if needing to roll back, we won't have SQL errors, but owner/timestamps data may be arbitrary.

kgajowy: all good, I wasn't sure if we have `default` set 👍

kgajowy (approved):

api/src/migrations/api/1619711501000-DropNotNullsOnTimeUserMetadataColumns.ts
hotzevzl: done in ac8f10379e509cce7ad6a530755ad19711d5e2e9

PR #143, opened 2021-04-30
marxan 89 add test post patch data shapes

alexeh:

**Add tests for:**

- Checking POST & PAST request follows the JSON:API Specs
- Check Pagination responses actually includes pagination metadata 

### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/test/json-api.e2e-spec.ts
+++ b/api/test/json-api.e2e-spec.ts
@@ -4,10 +4,21 @@ import * as request from 'supertest';
 import { AppModule } from '../src/app.module';
 import { omit } from 'lodash';
 import { E2E_CONFIG } from './e2e.config';
+import { ProjectsTestUtils } from './utils/projects.test.utils';
+import { Organization } from 'modules/organizations/organization.api.entity';
+import { OrganizationsTestUtils } from './utils/organizations.test.utils';
+import * as JSONAPISerializer from 'jsonapi-serializer';
+import { Project } from 'modules/projects/project.api.entity';
 
 describe('JSON API Specs (e2e)', () => {
   let app: INestApplication;
   let jwtToken: string;
+  let fakeCountry: string = 'ESP';
+  let fakeOrganization: Organization;
+  let fakeProject: Project;
+  const Deserializer = new JSONAPISerializer.Deserializer({
+    keyForAttribute: 'camelCase',
+  });
 
   beforeAll(async () => {
     const moduleFixture: TestingModule = await Test.createTestingModule({
@@ -16,6 +27,10 @@ describe('JSON API Specs (e2e)', () => {
 
     app = moduleFixture.createNestApplication();
     await app.init();
+
+    /**
+     * Login User
+     */
 
     const response = await request(app.getHttpServer())
       .post('/auth/sign-in')
@@ -26,6 +41,23 @@ describe('JSON API Specs (e2e)', () => {
       .expect(201);
 
     jwtToken = response.body.accessToken;
+
+    /**
+     * Create Scenario
+     */
+
+    fakeOrganization = await OrganizationsTestUtils.createOrganization(
+      app,
+      jwtToken,
+      E2E_CONFIG.organizations.valid.minimal(),
+    ).then(async (response) => await Deserializer.deserialize(response));
+
+    fakeProject = await ProjectsTestUtils.createProject(app, jwtToken, {
+      ...E2E_CONFIG.projects.valid.minimalInGivenAdminArea({
+        countryCode: fakeCountry,
+      }),
+      organizationId: fakeOrganization.id,
+    }).then(async (response) => await Deserializer.deserialize(response));
   });
 
   afterAll(async () => {
@@ -48,9 +80,10 @@ describe('JSON API Specs (e2e)', () => {
         stack: null,
       },
     };
-    const response = await request(app.getHttpServer()).get(
-      '/api/v1/projects/fakeProject/features?q=fakeFeature',
-    );
+
+    const response = await request(app.getHttpServer())
+      .get('/api/v1/projects/fakeProject/features?q=fakeFeature')
+      .set('Authorization', `Bearer ${jwtToken}`);
 
     response.body.errors.forEach((err: any) => {
       expect(Object.keys(jsonApiErrorResponse)).toEqual(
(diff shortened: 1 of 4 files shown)


api/test/json-api.e2e-spec.ts
```diff
@@ -3,9 +3,22 @@ import { INestApplication } from '@nestjs/common';
...
+import { ProjectsTestUtils } from './utils/projects.test.utils';
+import { Organization } from 'modules/organizations/organization.api.entity';
+import { OrganizationsTestUtils } from './utils/organizations.test.utils';
+import * as JSONAPISerializer from 'jsonapi-serializer';
+import { Project } from 'modules/projects/project.api.entity';
 
 describe('JSON API Specs (e2e)', () => {
   let app: INestApplication;
+  let jwtToken: string;
+  let fakeCountry: string = 'ESP';
```
kgajowy, line 16: `const fakeCountry = 'ESP'`. TS compiler complains ; )

I think we should enable "build" step in CI, otherwise we may end up in broken `develop` (in fact this is e2e so it isn't that important there but in general, we should)

api/test/json-api.e2e-spec.ts
```diff
@@ -64,4 +107,57 @@ describe('JSON API Specs (e2e)', () => {
       }
```
kgajowy, line 107: Not really related to this PR but I believe this isn't a reliable test. CI runs in a particular environment and any `if`s (especially there, that it may omit them at all) may cause false-positive test.
If a CI runs with `prod` for example as NODE_ENV, it will skip any assertions.

In particular, having those `if`s imply that we never check the second scenario

api/test/json-api.e2e-spec.ts
```diff
@@ -14,6 +27,37 @@ describe('JSON API Specs (e2e)', () => {
...
+      jwtToken,
+      E2E_CONFIG.organizations.valid.minimal(),
+    ).then(async (response) => await Deserializer.deserialize(response));
+
+    fakeProject = await ProjectsTestUtils.createProject(app, jwtToken, {
+      ...E2E_CONFIG.projects.valid.minimalInGivenAdminArea({
+        countryCode: fakeCountry,
+      }),
+      organizationId: fakeOrganization.id,
+    }).then(async (response) => await Deserializer.deserialize(response));
```
kgajowy, line 60: Would be great to make setups (Assets) clean as much as possible, thus I would recommend moving it to standalone function.

api/test/json-api.e2e-spec.ts
```diff
@@ -36,12 +80,11 @@ describe('JSON API Specs (e2e)', () => {
...
       },
     };
+
     const response = await request(app.getHttpServer())
-      .post('/auth/sign-in')
-      .send({
-        username: 'fakeuser@example.com',
-        password: 'fakePassword',
-      });
+      .get('/api/v1/projects/fakeProject/features?q=fakeFeature')
```
kgajowy, line 85: any clue why:
image

is logged to output?

api/test/json-api.e2e-spec.ts
```diff
@@ -64,4 +107,57 @@ describe('JSON API Specs (e2e)', () => {
...
+        ...E2E_CONFIG.projects.valid.minimalInGivenAdminArea({
+          countryCode: fakeCountry,
+        }),
+        organizationId: fakeOrganization.id,
+      });
+
+    expect(typeof response.body).toBe('object');
+    expect(Object.keys(response.body)).toHaveLength(1);
+    expect(response.body.hasOwnProperty('data')).toBe(true);
+    expect(typeof response.body.data).toBe('object');
```
kgajowy, line 125: What do you think about using custom matcher? We should actually test it for each endpoint, right? On top of that, it currently duplicates in a few places.

https://everyday.codes/javascript/create-custom-jest-matchers-to-test-like-a-pro/

api/test/json-api.e2e-spec.ts
```diff
@@ -64,4 +107,57 @@ describe('JSON API Specs (e2e)', () => {
...
+        organizationId: fakeOrganization.id,
+      });
+
+    expect(response.body.hasOwnProperty('meta')).toBe(true);
+    expect(Object.keys(response.body.meta)).toEqual([
+      'totalItems',
+      'totalPages',
+      'size',
+      'page',
+    ]);
```
kgajowy, line 161: How about some simplification?
```
expect(response.body).toMatchObject({
      meta: {
        totalItems: expect.any(Number),
        totalPages: expect.any(Number),
        size: expect.any(Number),
        page: expect.any(Number),
      },
    });
```

api/test/json-api.e2e-spec.ts
alexeh: Thanks for pointing out. It does not complain to me, but I'll add this change.
Maybe we should start using WebStorm...? xD

api/test/json-api.e2e-spec.ts
alexeh: Have to say that I am agree at this point. And I am personally not a fan of testing what an app does depending on enviroment, I think we should focus quality to production, and I'd rather check this logics in unit testing. But we talked about doing this, and having separated test for different enviroments required to stop, change env and re-run the test-app. 
Any better ways to achieve this?

api/test/json-api.e2e-spec.ts
alexeh: I do agree on this as well, but I believe this is needs to be talked about and reach to an agreement. It is not a big deal to refactor it later (I think)

api/test/json-api.e2e-spec.ts
alexeh: I see your point. My reason to do this is that I like to be accurate to the test-case. The task does not require to actually see the type of the data, but that contains pagination metadata info, thats why I just check the keys in it.
But maybe I should update the requirements and Acceptance Test to add more cases, since I feel @hotzevzl is really busy. I'd like to bring this to the table, but I believe now is not the time to do so

api/test/json-api.e2e-spec.ts
hotzevzl: yep, this could be a nice improvement - except for non-core stuff like pagination and error responses this has not been a problem in practice though, so I'd suggest to create a task for this but only get to it once we're well ahead with stories that unblock frontend work

api/test/json-api.e2e-spec.ts
hotzevzl: yep, I'd say while on this, good to also check the content as it's a small change.

api/test/json-api.e2e-spec.ts
hotzevzl: yep, what we have is already an improvement above the initial implementation (which is still used in a lot of early tests), which in turn was a great improvement above what we had in two previous projects (no e2e tests, and in the earlier no tests at all 😱) but there are better ways.
I suggest to introduce standalone functions bit by bit for new tests, and if/when we have time, update older tests.

api/test/json-api.e2e-spec.ts
alexeh: Oh, all right, is actually ESLint complaining, and I know why didn't complain to be. My bad here

api/test/json-api.e2e-spec.ts
kgajowy: @alexeh it only validates a type (which rather should be a number ;) You can write `expect.anything()` as well.
My point there is that writing all assertions and transforming response is just more unnecessary code.

kgajowy (approved):

hotzevzl (approved):

PR #144, opened 2021-04-30
add features and protected areas to proxy

elpamart:
Add protected areas and features endpoint in proxy.

### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/src/modules/proxy/proxy.controller.ts
+++ b/api/src/modules/proxy/proxy.controller.ts
@@ -101,11 +101,11 @@ export class ProxyController {
     required: true,
   })
   @ApiQuery({
-    name: 'wdpaid',
+    name: 'id',
     description: 'Id of WDPA area',
-    type: Number,
+    type: String,
     required: false,
-    example: '5555697751',
+    example: 'e5c3b978-908c-49d3-b1e3-89727e9f999c',
   })
   @Get('/protected-areas/preview/tiles/:z/:x/:y.mvt')
   async proxyProtectedAreaTile(
@@ -151,10 +151,11 @@ export class ProxyController {
     required: true,
   })
   @ApiQuery({
-    name: 'extentId',
-    description: 'Extent of the project',
-    type: Number,
+    name: 'bbox',
+    description: 'Bounding box of the project',
+    type: Array,
     required: false,
+    example: [-1, 40, 1, 42],
   })
   @Get('/features/:id/preview/tiles/:z/:x/:y.mvt')
   async proxyFeaturesTile(@Req() request: Request, @Res() response: Response) {
--- a/api/src/modules/proxy/proxy.controller.ts
+++ b/api/src/modules/proxy/proxy.controller.ts
@@ -16,10 +16,9 @@ import { Request, Response } from 'express';
 
 @UseGuards(JwtAuthGuard)
 @ApiBearerAuth()
-@ApiTags('AdminArea vector tile proxy')
+@ApiTags('Vector tile proxy')
 @Controller(`${apiGlobalPrefixes.v1}`)
 export class ProxyController {
-
   constructor(private readonly proxyService: ProxyService) {}
 
   @ApiOperation({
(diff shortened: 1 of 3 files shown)


api/src/modules/proxy/proxy.controller.ts
```diff
@@ -16,10 +16,9 @@ import { Request, Response } from 'express';
 
 @UseGuards(JwtAuthGuard)
 @ApiBearerAuth()
-@ApiTags('AdminArea vector tile proxy')
+@ApiTags('Vector tile proxy')
```
hotzevzl, line 19: good point 😄 

if we add more/unrelated (to tiles) proxy endpoints in the future we may want to *then* split things across this and another controller, all using the same service and grouped under the same module, but just a thought for the future - maybe we won't need any of this 😅

api/src/modules/proxy/proxy.controller.ts
```diff
@@ -67,13 +66,98 @@ export class ProxyController {
...
+    type: Number,
+    required: true,
+  })
+  @ApiQuery({
+    name: 'wdpaid',
+    description: 'Id of WDPA area',
+    type: Number,
+    required: false,
+    example: '5555697751',
+  })
```
hotzevzl, line 109: which id are we after here? My understanding is that we may not have data in the `wdpa.wdpaid` column in the geodb for protected areas that are uploaded by users - should we use the `wdpa.id` UUID column instead? (moreover, right now `wdpaid` is nullable, whereas `id` is a PK)

api/src/modules/proxy/proxy.controller.ts
```diff
@@ -67,13 +66,98 @@ export class ProxyController {
...
+    description: 'Specific id of the feature',
+    type: String,
+    required: true,
+  })
+  @ApiQuery({
+    name: 'extentId',
+    description: 'Extent of the project',
+    type: Number,
+    required: false,
+  })
```
hotzevzl, line 158: is this correct? what does extentId map to (in our data model)?

geoprocessing/src/modules/protected-areas/protected-areas.controller.ts
```diff
@@ -47,10 +47,10 @@ export class ProtectedAreasController<T> {
   })
   @ApiQuery({
     name: 'wdpaid',
-    description: 'Parent country of administrative areas in ISO code',
-    type: String,
+    description: 'Id of WDPA area',
```
hotzevzl, line 50: same as above in the corresponding controller in the API service

hotzevzl (commented):
This is basically an approval of the PR. please see my main question about which IDs to use for protected areas and (if appropriate) update that - other than this, LGTM, thanks! 🚀

api/src/modules/proxy/proxy.controller.ts
elpamart: haha yes, I though we could add all the tiles endpoints under the same proxy and potentially split things if we add something not tile related.

api/src/modules/proxy/proxy.controller.ts
elpamart: that was @aagm suggestion and based on what we have in our endpoint definition doc. I guess if we are going to have the case of `wdpa.wdpaid` with no data for the areas uploaded by the user, I'm happy to change this param to `wdpa.id`.

@aagm what do you think?

api/src/modules/proxy/proxy.controller.ts
elpamart: yes, this is what we defined on the [endpoints definition doc](https://docs.google.com/document/d/1T1Gp9RM6J-CoZoyIL7XBeUU1_w2ZrvN7rCSN_XAnBSg/edit#). If I remember well was the` features.id` but I cannot check on my local as the features import didn't work. @aagm could confirm

api/src/modules/proxy/proxy.controller.ts
aagm: yes i was willing to use based on wdpaid, but i didn't though on user uploaded data, so changing it based on the db based id wdpa.id is ok from my point of view!

api/src/modules/proxy/proxy.controller.ts
aagm: This looks like a slip-on in my mind regarding how will we access the extent. I'm thinking if indeed we should refer to bbox here and filter by it.  But I think that based on the designs this is more related to the shape of the study area (or adm or extent, so I would refer it to a project id, and we can generate the specific logic from there.)

PR #147, opened 2021-04-30
various fixes/improvements to CI workflow

hotzevzl:
- Allow to mute all logging irrespective of Node env and other settings that would affect Logger verbosity, via `API_LOGGING_MUTE_ALL` envvar

e.g. `docker-compose run -e API_LOGGING_MUTE_ALL=true api yarn test:e2e`

- Backport (manually for now) today's updates to `seed-geoapi-with-test-data` Makefile recipe to local and ci e2e recipes

Missing (moving them to future PRs):

- [ ] ability to run only specific test suites when running the full local CI pipeline (`make test-e2e-api <pattern>`)
- [ ] make `--watch` work
- [ ] run unit tests in CI
- [ ] avoid setting logger configuration in each service

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,6 +116,9 @@ test-e2e-api:
 		echo "appending data for $${table_name} with id $${featureid}"; \
 		sed -e "s/\$$feature_id/$$featureid/g" api/test/fixtures/features/$${table_name}.sql | docker-compose -f docker-compose-test-e2e.yml -f docker-compose-test-e2e.local.yml --env-file .env-test-e2e exec -T test-e2e-postgresql-geo-api psql -U "${GEO_POSTGRES_USER}"; \
 		done;
+	SCENARIOID=$(shell docker-compose -f docker-compose-test-e2e.yml -f docker-compose-test-e2e.local.yml --env-file .env-test-e2e exec -T test-e2e-postgresql-api psql -X -A -t -U "${API_POSTGRES_USER}" -c "select id from scenarios where name = 'Example scenario 1 Project 1 Org 1'"); \
+	echo "appending data for scenario with id $${SCENARIOID}"; \
+	sed -e "s/\$$user/00000000-0000-0000-0000-000000000000/g" -e "s/\$$scenario/$$SCENARIOID/g" api/test/fixtures/test-geodata.sql | docker-compose -f docker-compose-test-e2e.yml -f docker-compose-test-e2e.local.yml --env-file .env-test-e2e exec -T test-e2e-postgresql-geo-api psql -U "${GEO_POSTGRES_USER}";
 	# run tests
 	docker-compose -f docker-compose-test-e2e.yml -f docker-compose-test-e2e.local.yml --env-file .env-test-e2e up --abort-on-container-exit --exit-code-from api api
 	# teardown
--- a/Makefile-ci
+++ b/Makefile-ci
@@ -25,7 +25,10 @@ test-e2e-api:
 		echo "appending data for $${table_name} with id $${featureid}"; \
 		sed -e "s/\$$feature_id/$$featureid/g" api/test/fixtures/features/$${table_name}.sql | docker-compose -f docker-compose-test-e2e.yml -f docker-compose-test-e2e.ci.yml exec -T test-e2e-postgresql-geo-api psql -U "${GEO_POSTGRES_USER}"; \
 		done;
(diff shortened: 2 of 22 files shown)


hotzevzl:
@kgajowy this PR should improve the CI situation (at least for `make test-e2e-api` and when running tests via docker-compose), in terms of verbosity of the output and backporting new seed work from Alicia's work to the e2e workflows.

Setting of response headers when sending errors via exception filter also makes the three failing tests from the vector tiles e2e suite pass now.

I have added a list of items still to do in the PR description - though I may tackle them in a separate PR to avoid growing this one too much.

I am unhappy about the way my current PR code is passing on the `logging.muteAll` flag to the base service - in a way, we now have a service logger automatically in each service (vs manually injecting this into each service as we had started doing) but the `logging.muteAll` setting should be set at the application level and used verbatim in the constructor of each service rather than reading the relevant envvar via AppConfig in each constructor, but I didn't want to spend more time on this before the weekend.

There's also an PostGIS error that happens only in the local CI workflow (and very likely in the GH actions one if this was used) - I'll need to check next week:

```
echo "appending data for scenario with id ${SCENARIOID}"; \
sed -e "s/\$user/00000000-0000-0000-0000-000000000000/g" -e "s/\$scenario/$SCENARIOID/g" api/test/fixtures/test-geodata.sql | docker-compose -f docker-compose-test-e2e.yml -f docker-compose-test-e2e.local.yml --e
nv-file .env-test-e2e exec -T test-e2e-postgresql-geo-api psql -U "marxan-geo-api";
appending data for scenario with id 7be47aaf-0e71-45ad-91f9-e80b3e56447e
ERROR:  function st_squaregrid(integer, geometry) does not exist
LINE 4: (SELECT (ST_SquareGrid(1000, ST_Transform(ST_GeomFromGeoJSON...
                 ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

Overall this should get us quite closer to a newly polished CI pipeline, but there's still some work I need to do.

kgajowy:
> the `logging.muteAll` setting should be set at the application level and used verbatim in the constructor of each service rather than reading the relevant envvar via AppConfig in each constructor,

I would say it would be best to not re-define this setting each time in every single service but at the whole application level - at least as our long-term goal. 

Looks great for now! ✅

hotzevzl:
> > the `logging.muteAll` setting should be set at the application level and used verbatim in the constructor of each service rather than reading the relevant envvar via AppConfig in each constructor,
> 
> I would say it would be best to not re-define this setting each time in every single service but at the whole application level - at least as our long-term goal.

yep, that's the intention. I did run out of ideas on how to do this without explicitly passing it in each service's constructor, but at the very least the setting should be read **once** via `AppConfig.get()` and then reused. Suggestions on how to pass this to the base class without fiddling with the constructors are welcome 😃

kgajowy:
> > > the `logging.muteAll` setting should be set at the application level and used verbatim in the constructor of each service rather than reading the relevant envvar via AppConfig in each constructor,
> > 
> > 
> > I would say it would be best to not re-define this setting each time in every single service but at the whole application level - at least as our long-term goal.
> 
> yep, that's the intention. I did run out of ideas on how to do this without explicitly passing it in each service's constructor, but at the very least the setting should be read **once** via `AppConfig.get()` and then reused. Suggestions on how to pass this to the base class without fiddling with the constructors are welcome 😃

How about using [custom](https://docs.nestjs.com/techniques/logger#dependency-injection) Logger? 

```
app.useLogger(app.get(MyLogger));
```
Then we could inject config once within `MyLogger` implementation.

kgajowy (approved):

hotzevzl:
> > yep, that's the intention. I did run out of ideas on how to do this without explicitly passing it in each service's constructor, but at the very least the setting should be read **once** via `AppConfig.get()` and then reused. Suggestions on how to pass this to the base class without fiddling with the constructors are welcome smiley
> 
> How about using [custom](https://docs.nestjs.com/techniques/logger#dependency-injection) Logger?
> 
> ```
> app.useLogger(app.get(MyLogger));
> ```
> 
> Then we could inject config once within `MyLogger` implementation.

Sounds good. I am actually tempted to just remove all logging output from nestjs-base-service and call it a day 😱 

Ok, to avoid keeping things moving too long, I'll merge this PR and then address the missing bits including this logging setup in a separate PR (or more)

PR #148, opened 2021-05-04
chore(api): scenarios-planning-units base service

kgajowy:
### Overview

integration-e2e was mostly added to verify mapping(transform)

image

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/src/modules/scenarios-planning-unit/entities/scenarios-planning-unit.geo.entity.ts
+++ b/api/src/modules/scenarios-planning-unit/entities/scenarios-planning-unit.geo.entity.ts
@@ -1,0 +1,72 @@
+import { Column, Entity, PrimaryGeneratedColumn } from 'typeorm';
+import { LockStatus } from '../lock-status.enum';
+
+export const scenariosPuDataEntityName = 'scenarios_pu_data';
+
+const toLockEnum: Record<1 | 2, LockStatus> = Object.freeze({
+  1: LockStatus.LockedIn,
+  2: LockStatus.LockedOut,
+});
+
+const fromLockEnum: Record<LockStatus, 1 | 2 | null> = Object.freeze({
+  [LockStatus.Unknown]: null,
+  [LockStatus.LockedIn]: 1,
+  [LockStatus.LockedOut]: 2,
+});
+
+@Entity(scenariosPuDataEntityName)
+export class ScenariosPlanningUnitGeoEntity {
+  @PrimaryGeneratedColumn('uuid')
+  id!: string;
+
+  /**
+   * missing FK
+   */
+  @Column({
+    type: 'uuid',
+    name: 'pu_geom_id',
+  })
+  puGeometryId!: string;
+
+  /**
+   * missing FK
+   */
+  @Column({
+    type: 'uuid',
+    name: 'scenario_id',
+  })
+  scenarioId!: string;
+
+  /**
+   * missing FK
+   */
+  @Column({
+    type: 'int',
+    name: 'puid',
+  })
+  projectId!: number;
+
+  @Column({
+    type: 'int',
+    name: 'lockin_status',
+    transformer: {
+      from(value: number | null): LockStatus {
+        if (value !== null && (value === 1 || value === 2)) {
+          return toLockEnum[value];
+        }
+        return LockStatus.Unknown;
+      },
+      to(value: LockStatus): 1 | 2 | null {
+        return fromLockEnum[value];
+      },
+    },
+  })
+  /**
+   * null - not touched
+   * 1 - locked-in (included)
+   * 2 - locked-out (excluded)
+   */
+  lockStatus!: LockStatus;
+
+  // relations not defined yet until it is necessary
+}
--- a/api/src/modules/scenarios-planning-unit/entities/scenarios-planning-unit.geo.entity.ts
+++ b/api/src/modules/scenarios-planning-unit/entities/scenarios-planning-unit.geo.entity.ts
@@ -3,13 +3,14 @@ import { LockStatus } from '../lock-status.enum';
 
 export const scenariosPuDataEntityName = 'scenarios_pu_data';
 
-const toLockEnum: Record<1 | 2, LockStatus> = Object.freeze({
+const toLockEnum: Record<0 | 1 | 2, LockStatus> = Object.freeze({
+  0: LockStatus.Unstated,
   1: LockStatus.LockedIn,
   2: LockStatus.LockedOut,
 });
 
-const fromLockEnum: Record<LockStatus, 1 | 2 | null> = Object.freeze({
-  [LockStatus.Unknown]: null,
+const fromLockEnum: Record<LockStatus, null | 1 | 2> = Object.freeze({
+  [LockStatus.Unstated]: null,
   [LockStatus.LockedIn]: 1,
   [LockStatus.LockedOut]: 2,
 });
@@ -44,8 +45,14 @@ export class ScenariosPlanningUnitGeoEntity {
     type: 'int',
     name: 'puid',
   })
-  projectId!: number;
+  /**
+   * numeric id for each PU (unique within each scenario. It's called puid in the db as it maps directly to the same-named variable in Marxan dat files
+   */
+  planningUnitMarxanId!: number;
 
+  // TODO: debt: either 0|1|2|3 or null|1|2|3; currently DB allows nulls
+  // @Min(0)
+  // @Max(3)
   @Column({
     type: 'int',
     name: 'lockin_status',
@@ -54,9 +61,9 @@ export class ScenariosPlanningUnitGeoEntity {
         if (value !== null && (value === 1 || value === 2)) {
           return toLockEnum[value];
         }
-        return LockStatus.Unknown;
+        return LockStatus.Unstated;
       },
-      to(value: LockStatus): 1 | 2 | null {
+      to(value: LockStatus): null | 1 | 2 {
         return fromLockEnum[value];
       },
     },
--- a/api/test/scenarios-pu-data/steps/given-scenario-pu-data-exists.ts
+++ b/api/test/scenarios-pu-data/steps/given-scenario-pu-data-exists.ts
@@ -1,0 +1,47 @@
+import { INestApplication } from '@nestjs/common';
+import { Repository } from 'typeorm';
+import { getRepositoryToken } from '@nestjs/typeorm';
+import { v4 } from 'uuid';
+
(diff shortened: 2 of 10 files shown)


api/src/modules/scenarios-planning-unit/entities/scenarios-planning-unit.geo.entity.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export const scenariosPuDataEntityName = 'scenarios_pu_data';
+
+const toLockEnum: Record<1 | 2, LockStatus> = Object.freeze({
+  1: LockStatus.LockedIn,
+  2: LockStatus.LockedOut,
+});
+
+const fromLockEnum: Record<LockStatus, 1 | 2 | null> = Object.freeze({
+  [LockStatus.Unknown]: null,
```
hotzevzl, line 12: I think we should also explicitly handle the value `0` in the _from_ transform as this is something we'll normally want to set a PU `lockin_status` to for any PUs which are not explicitly included or excluded (and we then use this value of `0` verbatim when generating Marxan .dat files - ofc we could as well just store `1 | 2 | null` in the column and only output the `0`s when generating .dat files, but I'd keep the initial design on this).

Status `0` thus becomes something that may be labelled "not locked" or "unstated" (or something similar - I think NotLocked is good enough).

api/src/modules/scenarios-planning-unit/entities/scenarios-planning-unit.geo.entity.ts
```diff
@@ -0,0 +1,72 @@
...
+  scenarioId!: string;
+
+  /**
+   * missing FK
+   */
+  @Column({
+    type: 'int',
+    name: 'puid',
+  })
+  projectId!: number;
```
hotzevzl, line 47: this column is only a numeric id for each PU (unique within each scenario, I think we would need a constraint for this). It's called `puid` in the db as it maps directly to the same-named variable in Marxan dat files: given this, I'd suggest to name the property itself `puid` too, or if wanting to be more explicit `planningUnitId`, though this may even cause more confusion (because we have PU ids as in UUIDs and then again an id of each `scenarios_pu_data` row.

api/test/scenarios-pu-data/steps/given-scenario-pu-data-exists.ts
```diff
@@ -0,0 +1,47 @@
...
+  scenarioId = v4(),
+): Promise<{
+  scenarioId: string;
+  projectId: number;
+  rows: ScenariosPlanningUnitGeoEntity[];
+}> => {
+  const repo: Repository<ScenariosPlanningUnitGeoEntity> = await app.get(
+    getRepositoryToken(ScenariosPlanningUnitGeoEntity, remoteConnectionName),
+  );
+  const projectId = 1234;
```
hotzevzl, line 21: Given my other note about the role of the `puid` column, I suggest instead to make this a numeric sequence (and ideally to also delete the parent scenario in the `afterEach()` step of the test above, just for general cleanup as we won't need to  deal with scenario id clashes due to using `v4()` to generate these.

hotzevzl (changes requested):
thanks @kgajowy, looks great.

I have suggested a couple of minor amendments, please have a look and this PR should be done then 🚀

api/src/modules/scenarios-planning-unit/entities/scenarios-planning-unit.geo.entity.ts
kgajowy: @hotzevzl sure, I just assumed we don't use `0`
image
(will fix!)

api/test/scenarios-pu-data/steps/given-scenario-pu-data-exists.ts
kgajowy: There aren't any `FKs` so it is possible to create rows without any relations (and thus, we don't need scenario there at all, nor its removal)

api/src/modules/scenarios-planning-unit/entities/scenarios-planning-unit.geo.entity.ts
kgajowy: @hotzevzl I would keep using `unknown` we already have there (their meaning is the same, isn't it?)

api/src/modules/scenarios-planning-unit/entities/scenarios-planning-unit.geo.entity.ts
kgajowy: @hotzevzl how about `planningUnitMarxanId` in this case to better expose its nature? The reason I am afraid to leave "original" names within code is that their... meaning is unclear.

api/src/modules/scenarios-planning-unit/entities/scenarios-planning-unit.geo.entity.ts
kgajowy: on the other hand, if we want to keep `null` instead of `0` (`null` is possible to provide) this would require a bit of mess (i.e. `lockStatus!: LockStatus | null;`)

hotzevzl: I think we should either allow null and restrict the non-null values to 1, 2 or 3 (we don't use 3 in the app, but it's a legitimate value for Marxan data, and we will work on importing legacy Marxan projects later on, which could use this value, in which case we'd map 0 in the input to null in the db), *or* make the column non-nullable with default 0 and possible values restricted to 0, 1, 2 and 3.
I'd defer this decision to Alicia though, as she will have other implications of any such choices in her plans.

The name is really just minor semantics, and since we don't surface this to the users it's largely immaterial. In practice, 0 in Marxan means that the PU is neither locked in nor locked out, that means it's available for a decision on its final status (included or excluded from the proposed planning solution) by the Marxan CLI's algorithm, so strictly speaking is "not locked" (whereas a PU which is locked in is always guaranteed to be tagged as included in the planning solution).

api/src/modules/scenarios-planning-unit/entities/scenarios-planning-unit.geo.entity.ts
hotzevzl: @kgajowy I would go with your suggestion, thanks.
I am ok with using the original names too - that "only" forces each dev to memorize the 133 pages of the Marxan manual 😬 , but in general we don't need to save every character like in a 1980s MSDOS dat file, so descriptive names are ok too (and leaves us with the pleasure of dealing with one of the two complex problems in computer science, i.e. naming things 😅 )

api/test/scenarios-pu-data/steps/given-scenario-pu-data-exists.ts
hotzevzl: you're totally right - I missed that we're only inserting scenarios_pu_data rows here.

still best to clean these rows up (by the v4()-generated scenarioId?) on teardown to avoid keeping test data around if this is run against a volume-backed db (rather than the ephemeral dbs in the CI pipeline)?

api/src/modules/scenarios-planning-unit/entities/scenarios-planning-unit.geo.entity.ts
kgajowy: updated in https://github.com/Vizzuality/marxan-cloud/pull/148/commits/6396c14d671e06fbc11259fc539499d536454a85

Decided to keep "current" values (null,0,1) until we do a proper migration according to Alicia's decision.

I also put jsdoc for few mentioned values for future reference.

api/src/modules/scenarios-planning-unit/entities/scenarios-planning-unit.geo.entity.ts
kgajowy: fixed in https://github.com/Vizzuality/marxan-cloud/pull/148/commits/6396c14d671e06fbc11259fc539499d536454a85

api/test/scenarios-pu-data/steps/given-scenario-pu-data-exists.ts
kgajowy: `id` fixed in https://github.com/Vizzuality/marxan-cloud/pull/148/commits/6396c14d671e06fbc11259fc539499d536454a85

kgajowy: `cleanup` fixed in https://github.com/Vizzuality/marxan-cloud/pull/148/commits/e7c84aeaad8622b4c1682b9fd953add6f7395bbb 
Refactored steps a bit.

api/src/modules/scenarios-planning-unit/entities/scenarios-planning-unit.geo.entity.ts
aagm: First of all, let me put here the values that the lock-in column should be:

```
{
                       0: 'The PU is not guaranteed to be in the initial reserve',
                       1: 'The PU will be included in the initial reserve',
                       2: 'The PU is fixed in the reserve system (“locked-in”). starts in the initial reserve system and cannot be removed.',
                       3: 'The PU is fixed outside the reserve system (“locked-out”). It is not included in the initial reserve system and cannot be added.'
                      }
```

So my recommendation: In this column, we should make the column non-nullable with default 0 and possible values restricted to 0, 1, 2, and 3 as Andrea suggested. Making them nullable will require us to make a transformation at `.dat ` creation or project import that I think we can avoid. Most of the projects will go or 0 or 2 or 3 in very rare cases.

kgajowy: @aagm the issue is that it is currently allowed to put `null`s. I am fine to adjust them but this may require adjustments on GeoService side. How about making it ina separate PR and touch all places where it is needed?

aagm: sounds good to me, lets migrate this on a separate pr

kgajowy: https://vizzuality.atlassian.net/browse/MARXAN-398

api/test/scenarios-pu-data/steps/given-scenario-pu-data-exists.ts
aagm: Ok I didn't constrain this PU because we need to also allow User uploads with their own ids. This ids are also generated at insert time so we can control it but a unique constrain based on scenario id should be added. Task for a next pr

hotzevzl: I have added this to MARXAN-398

hotzevzl (approved):

PR #149, opened 2021-05-04
chore(api): analysis module for incl/excl feature

kgajowy:
## Analysis module application 

Implementation of the core/coordinator of change PU include/exclude things. To be used from controller.

### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


api/src/modules/analysis/analysis.service.ts
```diff
@@ -0,0 +1,52 @@
...
+import { ArePuidsAllowedPort } from './are-puids-allowed.port';
+import { RequestJobPort } from './request-job.port';
+
+import { AnalysisInput } from './analysis-input';
+import { JobStatusPort } from './job-status.port';
+
+type Success = true;
+
+@Injectable()
+export class AnalysisService {
```
kgajowy, line 13: (?) AdjustPlanningUnitsService

api/src/modules/analysis/analysis.service.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    await this.jobRequester.queue({
+      scenarioId,
+      ...constraints,
+    });
+
+    return true;
+  }
+
+  async getJobStatus(scenarioId: string): Promise<AsyncJob> {
```
kgajowy, line 49: Separate intentions from queries

api/src/modules/analysis/analysis.service.ts
```diff
@@ -0,0 +1,52 @@
...
+import { AsyncJob } from './async-job';
+import { ArePuidsAllowedPort } from './are-puids-allowed.port';
+import { RequestJobPort } from './request-job.port';
+
+import { AnalysisInput } from './analysis-input';
+import { JobStatusPort } from './job-status.port';
+
+type Success = true;
+
+@Injectable()
```
kgajowy, line 12:
```
import { FeatureCollection, GeoJSON, Geometry } from 'geojson';

// shapefile -> GeoJson

export class CostSurfaceInputDto {
  geo!: FeatureCollection<
    Geometry,
    {
      cost: number;
      planningUnitId: string;
    }
  >;
}
```
input for CostSurfaceService

api/src/modules/analysis/providers/planning-units/adapters/are-puids-allowed-adapter.ts
```diff
@@ -0,0 +1,19 @@
+import { Injectable } from '@nestjs/common';
+import { ArePuidsAllowedPort } from '../are-puids-allowed.port';
+
+import { ScenariosPlanningUnitService } from '../../../../scenarios-planning-unit/scenarios-planning-unit.service';
```
hotzevzl, line 4: this is very much cosmetic and any IDE can deal with this, but can we stick to paths relative to `src`, except within modules where it makes sense to use paths relative to the current file?

api/src/modules/analysis/providers/planning-units/adapters/are-puids-allowed-adapter.ts
```diff
@@ -0,0 +1,19 @@
...
+
+@Injectable()
+export class ArePuidsAllowedAdapter
+  extends ScenariosPlanningUnitService
+  implements ArePuidsAllowedPort {
+  async validate(
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
hotzevzl, line 11: just cosmetic - you can avoid this by prepending `_` to var names (at least with our eslint settings) - of course this means renaming the variable later on, so up to you - I don't have an IDE shortcut set up for this so I tend to prepend `_` 😄

api/src/modules/analysis/providers/planning-units/update-planning-units.service.ts
```diff
@@ -0,0 +1,42 @@
...
+      ...(constraints.include?.pu ?? []),
+      ...(constraints.exclude?.pu ?? []),
+    ];
+    if (targetPuIds.length > 0) {
+      const { errors } = await this.puUuidValidator.validate(
+        scenarioId,
+        targetPuIds,
+      );
+      if (errors.length > 0) {
+        throw new Error('Given PU ids are not reachable in this context.');
```
hotzevzl, line 31: Should we treat this as a `BadRequestException`?

Also, I suggest to make the error message a bit clearer, for example

```suggestion
        throw new Error('One or more of the planning units provided for exclusion or inclusion does not match any planning unit of the present scenario.');
```

hotzevzl (approved):
@kgajowy thank you, and sorry for taking so long to review this - I guess I need to learn how to lower the cognitive overhead when dealing with the structure of a feature separately from the way this is implemented.

All looks good - I have left a couple of cosmetic comments and a suggestion to improve an error message.

api/src/modules/analysis/providers/planning-units/adapters/are-puids-allowed-adapter.ts
kgajowy: @hotzevzl to be honest, I would love us to move to `paths` (see tsconfig option, similar to https://docs.nestjs.com/cli/libraries#using-libraries) rather than relative paths. Using `/src...` or `/module...`  breaks `jest`s unit tests.

api/src/modules/analysis/providers/planning-units/adapters/are-puids-allowed-adapter.ts
kgajowy: It is soon to be removed anyway (there https://github.com/Vizzuality/marxan-cloud/pull/165). As you said, this is a personal preference but leaving it silenced with `_prefix` leaves eye-state as "its all fine" while disable comment... stings the eyes.  I could as well just omit them but why? (: It does not matter much tho but if want to stick to project's style guide I am fine with that (as long as we note it down :)

hotzevzl: nah, i'd leave this to devs, either way it's clear

api/src/modules/analysis/providers/planning-units/adapters/are-puids-allowed-adapter.ts
kgajowy: @hotzevzl you may be correct. Will investigate this topic a bit more after/during nestjs-monorepo.

PR #150, opened 2021-05-05
chore: add scenario-pu-request entity definition

kgajowy:
## `scenario-pu-request` entity & service

Please keep in mind that migration **was fully generated** from `.entity` code. 

Things to keep in mind: validating if given `(ex)includedPlantingUnits` are `uuids` isn't that straightforward (even with using class-validator and `@BeforeSave/Update` hooks). As we have a task to validate if those are allowed anyway, decided to drop the validation out of the entity model.

#### Fail on inserting invalid MultiPolygon (db constraint check)
```
[102.0, 2.0],
[104.0, 5.0],
[103.0, 3.0],
```
image

#### Fail on non MultiPolygon
image

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


kgajowy:
Decided that we will not store config - just proxy it directly for job.

PR #152, opened 2021-05-06
(chore) improve jest ci workflow part 2

hotzevzl:
### Overview

I have looked far and wide, but failed to understand why setting envvars from docker output (when we run `SELECT id...` via psql) works flakily when running `make test-e2e-api` locally and never works in GH Actions.

Since this blocks being able to run CI pipelines, I am proposing here to move the steps of a makefile recipe through which we inject data into a shell script that we run **within the api db container** (`api/test/fixtures/seed-all.sh`).

This also requires installing some APK packages - I'm currently doing this in the script itself.

In my view, this feels a bit hacky, but so far I couldn't think of a different (and not overengineered) way to make these tests work.

Data injection now works *locally* - I would suggest to review this PR and then merge it to see how things work in GH Actions.

We'll then need to fix/update some tests that are now failing, but this will be for a distinct PR.

### Testing instructions

`make test-e2e-api` should run migrations and inject seed data (no errors shown). Some tests will fail (see above - this is ok for now).

---

- [ ] Update CHANGELOG file


hotzevzl:
Closing, in favour of Alicia's Makefile work of art! 💓

PR #153, opened 2021-05-06
feat(api): protection-levels: queue & handling application

kgajowy:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


hotzevzl:
@kgajowy looks good to me, thank you.

the part on how queues are used is clear, but I am not clear about how we'd trigger the command - from a different service where the calculations are needed?

also, just a thought about splitting things up - it's ok to keep modules simple and focused, but in practice some bits (calculations, updates to data, etc.) do fit together within a "parent" module - I would not go out of our way to implement degrees of flexibility that we may realistically not need, with the cost of more code and/or abstractions to maintain over time. As often, I think it's a matter of striking an optimal balance.

I am not sure also about circular dependencies - besides some nasty messages shouted by NestJS occasionally when we forget to handle a circular dependency, I have not run into problems with these in practice, but I am of course open to opinions on this 😄 

A final note - since you mentioned e2e tests in the Slack message pointing to this PR, let's discuss separately how to best handle e2e tests for async/queue-based jobs so that we can settle on a single framework for this and use it consistently (while avoiding to end up kind of testing bullmq and redis themselves, ofc, which I imagine is your concern here).

kgajowy:
@hotzevzl I suppose we can push this forward (see PR's description for details - nothing new there tbh, just need a confirmation to get next piece of puzzle ;)

hotzevzl (approved):

PR #155, opened 2021-05-06
add planning unit vector tile preview wip

elpamart:

Planning units preview view - vector tile service. The pr contains the implementation of the endpoint on the proxy module.

Todo: need to implement the query params

### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


geoprocessing/test/index.html
```diff
@@ -80,54 +88,65 @@
             'minzoom': 1,
             'maxzoom': 22
         },
+        // {
```
kgajowy, line 91: Commented code can be removed from source. If this piece may be usable in future, we can always link to particular SHA in codebase on wiki/gist etc.

geoprocessing/src/modules/planning-units/planning-units.geo.entity.ts
```diff
@@ -0,0 +1,22 @@
...
+  @ApiProperty()
+  @PrimaryColumn('uuid', { name: 'id' })
+  id: string;
+
+  /**
+   * geometry column.
+   */
+  @ApiProperty()
+  @Column('geometry', { name: 'the_geom', select: false })
+  theGeom: any;
```
kgajowy, line 21: We should avoid using `any` whenever possible. 

Take a look [there](https://github.com/Vizzuality/marxan-cloud/pull/150/files#diff-e860777d8791413043fe2da6becf6ed22d1471e696f4f5a13bd78460f1d69f2eR46) for proper typing on `Geometry`

geoprocessing/src/modules/planning-units/planning-units.controller.ts
```diff
@@ -0,0 +1,83 @@
...
+  @Header('Content-Disposition', 'attachment')
+  @Header('Access-Control-Allow-Origin', '*')
+  @Header('Content-Encoding', 'gzip')
+  async getTile(
+    @Param() tileRequest: TileRequest,
+    @Query('bbox') bbox: number[],
+    @Query('planningUnitGridShape') planningUnitGridShape: string,
+    @Query('planningUnitAreakm2') planningUnitAreakm2: number,
+    @Res() response: Response,
+  ): Promise<Object> {
```
kgajowy, line 79: so actually it sends `Buffer`, not `Object`, right?

geoprocessing/src/modules/planning-units/planning-units.controller.ts
```diff
@@ -0,0 +1,83 @@
...
+  @Header('Content-Type', 'application/x-protobuf')
+  @Header('Content-Disposition', 'attachment')
+  @Header('Access-Control-Allow-Origin', '*')
+  @Header('Content-Encoding', 'gzip')
+  async getTile(
+    @Param() tileRequest: TileRequest,
+    @Query('bbox') bbox: number[],
+    @Query('planningUnitGridShape') planningUnitGridShape: string,
+    @Query('planningUnitAreakm2') planningUnitAreakm2: number,
+    @Res() response: Response,
```
kgajowy, line 78: Injecting `Response` is not necessary, NestJS handles it on its own, so you should be able to safely return `tile` below and remove `@Res()`

geoprocessing/src/modules/planning-units/planning-units.controller.ts
```diff
@@ -0,0 +1,83 @@
...
+  @ApiBadRequestResponse()
+  @Header('Content-Type', 'application/x-protobuf')
+  @Header('Content-Disposition', 'attachment')
+  @Header('Access-Control-Allow-Origin', '*')
+  @Header('Content-Encoding', 'gzip')
+  async getTile(
+    @Param() tileRequest: TileRequest,
+    @Query('bbox') bbox: number[],
+    @Query('planningUnitGridShape') planningUnitGridShape: string,
+    @Query('planningUnitAreakm2') planningUnitAreakm2: number,
```
kgajowy, line 77: It is possible to create a `dto` file with all those definitions and have Nest include wanted params for you:
`@Query() inputs: YourDtoClass`.
You could then also move Api-Doc decorators to that `dto` file.

On top of that - currently there is no validation so you can receive `bbox` as `string` or `null` and so on.

geoprocessing/src/modules/planning-units/planning-units.controller.ts
```diff
@@ -0,0 +1,83 @@
...
+  ApiParam,
+  ApiQuery,
+  ApiBadRequestResponse,
+} from '@nestjs/swagger';
+import { TileRequest } from 'src/modules/tile/tile.service';
+
+import { Response } from 'express';
+
+@Controller(`${apiGlobalPrefixes.v1}`)
+export class PlanningUnitsController<T> {
```
kgajowy, line 23: Wondering if there is reason we need generic type there?

geoprocessing/src/modules/planning-units/planning-units.controller.ts
```diff
@@ -0,0 +1,83 @@
...
+  private readonly logger: Logger = new Logger(PlanningUnitsController.name);
```
kgajowy, line 24: Seems to be not used anywhere.

api/src/modules/proxy/proxy.controller.ts
```diff
@@ -161,4 +161,60 @@ export class ProxyController {
...
+    type: String,
+    required: false,
+  })
+  @ApiQuery({
+    name: 'planningUnitAreakm2',
+    description: 'Planning unit area in km2',
+    type: Number,
+    required: false,
+  })
+  @Get('/planning-units/preview/tiles/:z/:x/:y.mvt')
```
hotzevzl, line 213: may I suggest to start moving these endpoints to the controller they logically belong to?

In practice, we should get rid of this ProxyController, after all, I think - unless we have proxied requests that don't belong elsewhere. So this specific new endpoint in this PR could be moved to a (new in this case) `PlanningUnitsController` within the (existing) `PlanningUnitsModule`.

If you agree with moving these endpoints, feel free to do this in a separate PR, of course.

The main benefit from my point of view is that we would be grouping them together with related endpoints (so the features tiles endpoint could go to the existing GeoFeaturesController, and so on), which also helps with the generated Swagger documentation.

hotzevzl:
Superceded by https://github.com/Vizzuality/marxan-cloud/pull/182

PR #156, opened 2021-05-07
chore(api): integration-or-e2e tests with queues

kgajowy:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/src/modules/planning-units-protection-level/calculate-planning-units-protection-level.handler.ts
+++ b/api/src/modules/planning-units-protection-level/calculate-planning-units-protection-level.handler.ts
@@ -17,7 +17,11 @@ export class CalculatePlanningUnitsProtectionLevelHandler
   async execute(
     command: CalculatePlanningUnitsProtectionLevel,
   ): Promise<CalculatePlanningUnitsProtectionLevelResult> {
-    await this.queueService.queue.add(this.queueService.name, command, {});
+    await this.queueService.queue.add(
+      `calculate-planning-units-protection-level-${command.scenarioId}`,
+      command,
+      {},
+    );
 
     return true;
   }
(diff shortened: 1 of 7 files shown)


hotzevzl (commented):
@kgajowy this looks good to me, and I agree on the scope of what we test.

This is basically the same we've done in the tests for the first module where we've added BullMQ jobs (`planning-units.e2e-spec.ts`), but with the addition here of the uniform abstraction over queues.

kgajowy:
> @kgajowy firstly - are we talking about the same thing?
> 
> my understanding is that your core suggestion is that _for workflows that include orchestrating async jobs through a queue_, we should avoid checking that "Redis works" (and BullMQ as well) - if so, I agree with this. Redis and BullMQ have (or should have) their own tests, and there's no point in running more tests on top of the upstream ones, in my opinion.
> 
> We can check that jobs are put on a queue, and forget about them.
> 
> I would still include in our e2e tests requests that are reverse-proxied from API to geoprocessing service but are processed synchronously (such as tile generation right now, though this may be refactored to async with cache, in which case we should omit these from e2e tests).
> 
> We also however want to test that when we supply some kinds of inputs, upload some specific files, and so on, the async jobs _eventually_ get us the kind of data we expect to receive - is this what you mean with product e2e tests?
> 
> If so, I think we should tackle this together with FE and science, and do this through a scripted browser session.

yes, exactly:
```
We can check that jobs are put on a queue, and forget about them.
```
&
```
is this what you mean with product e2e tests
```

About the point with `synchronous` processing - that could be useful, however it is a bit artificial - due to changing how we process job, not sure what exact value it has. If we test two "end points" (Geo & Api) and leave the middle/glue (bull), we mostly covered the use case - in other words, testing bull with a different mode (not to mention different setup) is like testing library/tool itself - and IMO it would be enough to test it via "product testing"

PR #158, opened 2021-05-07
api(ci): enable unit tests

kgajowy:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/.github/workflows/api-e2e-tests.yml
+++ b/.github/workflows/api-e2e-tests.yml
@@ -26,5 +26,7 @@ jobs:
     timeout-minutes: 15
     steps:
     - uses: actions/checkout@v2
+    - name: Unit Tests
+      run: npm run test:ci
     - name: Run CI tests via make task
       run: make -f Makefile-ci test-e2e-api
--- a/api/src/filters/all-exceptions.exception.filter.spec.ts
+++ b/api/src/filters/all-exceptions.exception.filter.spec.ts
@@ -1,65 +1,66 @@
 import { AllExceptionsFilter } from './all-exceptions.exception.filter';
-import { HttpException, ArgumentsHost, HttpStatus } from '@nestjs/common';
+import { ArgumentsHost, NotFoundException } from '@nestjs/common';
+import { Test } from '@nestjs/testing';
 
 describe('AllExceptionFilter class spec (unit)', () => {
-  const allExceptionsFilter = new AllExceptionsFilter();
+  let filter: AllExceptionsFilter;
+  let argumentMock: jest.Mocked<ArgumentsHost>;
+  let jsonMock: jest.Mock;
 
-  /**
-   * Common Mocks
-   */
-  const responseMock = {
-    status: jest.fn().mockReturnThis(),
-    json: jest.fn(),
-    url: 'fakeUrl',
-  };
-
-  const hostMock = {
-    switchToHttp: () => {
-      return {
-        getResponse: () => responseMock,
-        getRequest: () => {
-          return { url: 1 };
-        },
-      };
-    },
-  } as ArgumentsHost;
-  afterEach(() => {
-    jest.clearAllMocks();
+  beforeEach(async () => {
+    const sandbox = await Test.createTestingModule({
+      providers: [AllExceptionsFilter],
+    }).compile();
+    filter = sandbox.get(AllExceptionsFilter);
+    jsonMock = jest.fn();
+    argumentMock = mockArgumentsHost(jsonMock);
   });
 
   it('should be defined', () => {
-    expect(new AllExceptionsFilter()).toBeDefined();
+    expect(filter).toBeDefined();
   });
 
-  it('should get HTTP status through exception Object', () => {
-    const exceptionObjectMock = ({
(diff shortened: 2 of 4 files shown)


kgajowy:
@hotzevzl Need your help to see why the job didn't trigger there (and hopefully make it work)

.github/workflows/api-e2e-tests.yml
```diff
@@ -26,5 +26,7 @@ jobs:
     timeout-minutes: 15
     steps:
     - uses: actions/checkout@v2
+    - name: Unit Tests
+      run: npm run test:ci
```
hotzevzl, line 30: I would use the same strategy we use for e2e, namely running these within a Docker container... It's tricky to test this in GH actions (because one needs to first merge to develop in any case) - I'll take care of this once Alicia is done with her own checks for the e2e pipeline.

api/src/filters/all-exceptions.exception.filter.spec.ts
```diff
@@ -1,65 +1,66 @@
 import { AllExceptionsFilter } from './all-exceptions.exception.filter';
-import { HttpException, ArgumentsHost, HttpStatus } from '@nestjs/common';
+import { ArgumentsHost, NotFoundException } from '@nestjs/common';
```
hotzevzl, line 2: what happened to this test file?! 😅

.github/workflows/api-e2e-tests.yml
kgajowy: @hotzevzl are you sure? What would be the benefits of it? I suspect they would only run slower - unit tests don't need any setup/ecosystem up & running.

api/src/filters/all-exceptions.exception.filter.spec.ts
kgajowy: @hotzevzl it mainly received some heavy lifting:
1) changed it to actually use NestJS ecosystem (i.e. is injected in DI instead manually constructed)
2) updated some assertions to be reliable (`expect(responseMock.json).toHaveBeenCalled();`, `expect(new AllExceptionsFilter()).toBeDefined();`)

Sorry, couldn't resist :(

.github/workflows/api-e2e-tests.yml
hotzevzl: tests themselves should run more or less the same - we're only executing Jest inside of a container, so the overhead should be negligible. Most of the time spent building the image is `yarn install` time, which would need to happen in any case even if running this outside of a container, and then there's ~20 seconds spent transpiling, which is not needed (we could use a different Dockerfile without this step).
You are right that there's not need for much here - the main benefit I see is that we can keep using yarn, and hence `yarn.lock`.
I'm going to push to a separate suggestion branch my local changes in a few minutes - I think there's no clash with the CI work Alicia is doing so we could test the unit tests setup as soon as we are happy with it.

kgajowy: @hotzevzl thats valid point. I am fine with how it runs on CI.

hotzevzl: See my tentative suggestions here (tenative as in, all works locally, need to merge to `develop` to check if this works in GH actions 😱) https://github.com/Vizzuality/marxan-cloud/tree/api/ci-unit-tests%2Bandrea-suggestions

hotzevzl: And I realized this will indeed require a small change once Alicia's work on the e2e pipeline is merged, as she got everything working without using a second makefile (`Makefile-ci`), but that's a matter of updating the GH actions config and removing the `Makefile-ci` (the main `Makefile` already has a recipe to run unit tests in my branch above)

hotzevzl: @kgajowy have a look at my suggested changes and once we're happy with some solution we can merge and see what GH Actions tells us.

There's a neat project that allows to simulate the GH Actions runners locally, but when I tried to use it a while ago I run into a couple of edge situations and it turned out to be much simpler to just merge some attempts into `develop` until all worked 🙄

hotzevzl: pfff... actually I can already adapt my suggestions to Alicia's work as we merged it on Friday, and she's only working on updating some bits - I will rebase my branch on `develop` if we want to go with these suggestions.

hotzevzl: and well, TIR (today I realized) that changes to GH actions configs are indeed picked up in branches - not sure how I even ever got the idea that these would need merging into the main branch 🤷🏼‍♂️ 
anyway, so this works: https://github.com/Vizzuality/marxan-cloud/runs/2545099663
but I forgot to use the silent flag - adding it now, and rebasing my branch at the same time

hotzevzl: I have created a PR with my suggestions here: https://github.com/Vizzuality/marxan-cloud/pull/164

kgajowy:
Closing in favor of improved https://github.com/Vizzuality/marxan-cloud/pull/164

PR #160, opened 2021-05-07
add unique constraint on admin_regions table

hotzevzl:
See https://vizzuality.atlassian.net/browse/MARXAN-146

### DIFF
--- a/geoprocessing/src/migrations/geoprocessing/1620408879000-AddUniqueConstraintsToAdminRegionsTable.ts
+++ b/geoprocessing/src/migrations/geoprocessing/1620408879000-AddUniqueConstraintsToAdminRegionsTable.ts
@@ -7,6 +7,7 @@ export class AddUniqueConstraintsToAdminRegionsTable1620408879000
 CREATE UNIQUE INDEX unique_l0_regions ON admin_regions (gid_0, level) where level = 'country';
 CREATE UNIQUE INDEX unique_l1_regions ON admin_regions (gid_1, level) where level = 'adm_1';
 CREATE UNIQUE INDEX unique_l2_regions ON admin_regions (gid_2, level) where level = 'adm_2';
+CREATE INDEX admin_regions_l012_ids ON admin_regions (gid_0, gid_1, gid_2);
     `);
   }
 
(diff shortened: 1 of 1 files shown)


geoprocessing/src/migrations/geoprocessing/1620408879000-AddUniqueConstraintsToAdminRegionsTable.ts
```diff
@@ -0,0 +1,20 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddUniqueConstraintsToAdminRegionsTable1620408879000
+  implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
```
kgajowy, line 6: Judging from the usage, we may want to consider adding composite index:

image

geoprocessing/src/migrations/geoprocessing/1620408879000-AddUniqueConstraintsToAdminRegionsTable.ts
```diff
@@ -0,0 +1,20 @@
...
+CREATE UNIQUE INDEX unique_l0_regions ON admin_regions (gid_0, level) where level = 'country';
+CREATE UNIQUE INDEX unique_l1_regions ON admin_regions (gid_1, level) where level = 'adm_1';
+CREATE UNIQUE INDEX unique_l2_regions ON admin_regions (gid_2, level) where level = 'adm_2';
```
kgajowy, line 9: As far as I remember, indexed value has a length limit (of `varchar(2712)`) thus we could enforce max length as well to avoid unpredicted fails on inserts.

hotzevzl:
> As far as I remember, indexed value has a length limit (of `varchar(2712)`) thus we could enforce max length as well to avoid unpredicted fails on inserts. 

I don't think we will need this in practice: we don't let users add admin regions as we only import this data from the GADM data source, and these `gid_*` columns should by design be always < ~15chars

kgajowy:
> > As far as I remember, indexed value has a length limit (of `varchar(2712)`) thus we could enforce max length as well to avoid unpredicted fails on inserts.
> 
> I don't think we will need this in practice: we don't let users add admin regions as we only import this data from the GADM data source, and these `gid_*` columns should by design be always < ~15chars

It could be even better, indexes will take less space if we strictly defined max size. A minor improvement tho.

hotzevzl:
> Judging from the usage, we may want to consider adding composite index:

:+1: done in 4e8cb74598e3100e5d710d8ecf588c18cb44a177

kgajowy (approved):

PR #161, opened 2021-05-10
Feature/api/marxan 294 upload shapefiles

alexeh:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


alexeh:
E2E Test related to this func. should be implemented after 
https://vizzuality.atlassian.net/browse/MARXAN-399
Is done

Thoughts?

geoprocessing/src/utils/file.utils.ts
```diff
@@ -0,0 +1,33 @@
...
+ * File Utils
+ */
+
+import { MulterOptions } from '@nestjs/platform-express/multer/interfaces/multer-options.interface';
+import { createReadStream, readdirSync } from 'fs';
+
+/**
+ * ES5 import to avoid TS complaining
+ */
+const multer = require('multer');
```
kgajowy, line 11: https://www.npmjs.com/package/@types/multer

geoprocessing/src/utils/file.utils.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Options for Multer
+ */
+export const uploadOptions: MulterOptions = {
+  storage: multer.diskStorage({
+    destination: function (
+      req: any,
+      file: any,
```
kgajowy, line 20: Consider providing proper types (with types package it should be easier)

geoprocessing/src/utils/file.utils.ts
```diff
@@ -0,0 +1,33 @@
...
+      cb: (error: Error | null, destination: string) => void,
+    ) {
+      cb(null, '/../tmp/');
```
kgajowy, line 23: It should be OS-agnostic. Consider using https://www.npmjs.com/package/temp-dir

geoprocessing/src/utils/file.utils.ts
```diff
@@ -0,0 +1,33 @@
...
+    },
+    filename: function (
+      req: any,
+      file: any,
+      cb: (error: Error | null, fieldname: string) => void,
+    ) {
+      cb(null, Date.now() + '-' + file.originalname);
```
kgajowy, line 30: It will cause troubles on some OS (`:` `/` `\` characters). Consider using `Date.now().getTime()` instead.

geoprocessing/src/utils/file.utils.ts
```diff
@@ -0,0 +1,33 @@
...
+import { createReadStream, readdirSync } from 'fs';
+
+/**
+ * ES5 import to avoid TS complaining
+ */
+const multer = require('multer');
+
+/**
+ * Options for Multer
+ */
```
kgajowy, line 15: JSDoc isn't any helpful there. Consider removing it or providing description on its purpose.

geoprocessing/src/utils/file.utils.ts
```diff
@@ -0,0 +1,33 @@
...
+export const uploadOptions: MulterOptions = {
```
kgajowy, line 16: Consider using `Object.freeze(...)` to avoid mutations (respective type of `Readonly<MulterOptions>` would be helpful as well)

geoprocessing/src/types/mapshaper.d.ts.bck
```diff
@@ -0,0 +1 @@
+declare module 'mapshaper';
```
kgajowy, line 1: not sure if it works with `.bck` extension, what is the purpose of this?

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -0,0 +1,56 @@
+import { Injectable, Logger } from '@nestjs/common';
+
+import { createReadStream, rmdirSync } from 'fs';
+const mapshaper = require('mapshaper');
+const unzipper = require('unzipper');
```
kgajowy, line 5: https://www.npmjs.com/package/@types/unzipper

we should aim at using TS whenever possible.

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -0,0 +1,56 @@
...
+
+@Injectable()
+export class ShapeFileService {
+  private readonly logger: Logger = new Logger(ShapeFileService.name);
+  constructor() {}
+
+  private _unzipShapefile(fileInfo: Express.Multer.File): Promise<string> {
```
kgajowy, line 12: 1) `_` is not necessary as the function is already `private`. 
2) it actually unzips any file, not only `shapefiles` - consider name change
3) shouldn't this method return `path` to unzipped file?

I would also consider moving `unzip` logic to standalone service (as well as handling removal of leftovers)

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -0,0 +1,56 @@
...
+        .on('close', () =>
+          resolve(`${fileInfo.filename} extracted succesfully`),
+        )
+        .on('error', () =>
+          reject(new Error(`${fileInfo.filename} could not been extracted`)),
+        );
+    });
+  }
+
+  private async _shapeFileToGeoJson(fileInfo: Express.Multer.File) {
```
kgajowy, line 29: `_` is not necessary, `private` modifier is already there.

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -0,0 +1,56 @@
...
+    );
+
+    return JSON.parse(_geoJson[outputKey].toString('utf-8'));
+  }
+
+  private wipeOutFolder(path: string): void {
+    rmdirSync(path, { recursive: true });
+  }
+
+  async getGeoJson(shapeFile: Express.Multer.File) {
```
kgajowy, line 42: Strongly recommending to add return type (`Promise<Geometry>` ? not sure, please verify)

Not sure as well if shapefile validates GeoJSON as well.

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -0,0 +1,56 @@
...
+      this.logger.log(await this._unzipShapefile(shapeFile));
+    } catch (err) {
+      this.logger.error(err);
+    }
+    const geoJson = await this._shapeFileToGeoJson(shapeFile).then(
+      (geoJson) => {
+        this.wipeOutFolder(shapeFile.path.replace('.zip', ''));
+        return geoJson;
+      },
+    );
```
kgajowy, line 53: as above, consider moving unzipping/cleaning to standalone, reusable service.

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -0,0 +1,56 @@
...
+        .on('error', () =>
+          reject(new Error(`${fileInfo.filename} could not been extracted`)),
+        );
+    });
+  }
+
+  private async _shapeFileToGeoJson(fileInfo: Express.Multer.File) {
+    const outputKey = `shapefile-${new Date().getTime()}.gejson`;
+    const _geoJson = await mapshaper.applyCommands(
+      `-i ${fileInfo.path.replace('.zip', '')}/*.shp -o ${outputKey}`,
```
kgajowy, line 32: Would be great to provide unit tests on sample file. I believe it does not fully parse shapefile (coordinates are missing)

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -0,0 +1,56 @@
+import { Injectable, Logger } from '@nestjs/common';
+
+import { createReadStream, rmdirSync } from 'fs';
+const mapshaper = require('mapshaper');
+const unzipper = require('unzipper');
+
+@Injectable()
+export class ShapeFileService {
+  private readonly logger: Logger = new Logger(ShapeFileService.name);
+  constructor() {}
```
kgajowy, line 10: Consider injecting the logger, otherwise it won't leverage global settings (like muting logs)

geoprocessing/src/modules/planning-units/planning-units.controller.ts
```diff
@@ -0,0 +1,31 @@
...
+  constructor(public shapefileService: ShapeFileService) {}
+  @UseInterceptors(FileInterceptor('file', uploadOptions))
+  @Post('/:id/planning-unit-shapefile')
+  async getShapeFile(
+    @Param('id') scenarioId: string,
+    @UploadedFile() file: Express.Multer.File,
+  ) {
+    const geoJson = this.shapefileService.getGeoJson(file);
+
+    return geoJson;
```
kgajowy, line 29: can just `return this.shapefileService.getGeoJson(file)`

geoprocessing/src/modules/planning-units/planning-units.controller.ts
```diff
@@ -0,0 +1,31 @@
...
+import { ShapeFileService } from '../shapefiles/shapefiles.service';
+
+@Controller(`${apiGlobalPrefixes.v1}/planning-units`)
+export class PlanningUnitsController {
+  private readonly logger: Logger = new Logger(PlanningUnitsController.name);
+  constructor(public shapefileService: ShapeFileService) {}
+  @UseInterceptors(FileInterceptor('file', uploadOptions))
+  @Post('/:id/planning-unit-shapefile')
+  async getShapeFile(
+    @Param('id') scenarioId: string,
```
kgajowy, line 24: Shouldn't it include `ParseUUID` pipe?

geoprocessing/src/modules/planning-units/planning-units.controller.ts
```diff
@@ -0,0 +1,31 @@
...
+import { uploadOptions } from 'src/utils/file.utils';
+import { ShapeFileService } from '../shapefiles/shapefiles.service';
+
+@Controller(`${apiGlobalPrefixes.v1}/planning-units`)
+export class PlanningUnitsController {
+  private readonly logger: Logger = new Logger(PlanningUnitsController.name);
+  constructor(public shapefileService: ShapeFileService) {}
+  @UseInterceptors(FileInterceptor('file', uploadOptions))
+  @Post('/:id/planning-unit-shapefile')
+  async getShapeFile(
```
kgajowy, line 23: Would be great to provide OpenAPI definitions for swagger for both input and output.

geoprocessing/src/modules/planning-units/planning-units.controller.ts
```diff
@@ -0,0 +1,31 @@
...
+import { FileInterceptor } from '@nestjs/platform-express';
+
+import { apiGlobalPrefixes } from 'src/api.config';
+import { uploadOptions } from 'src/utils/file.utils';
+import { ShapeFileService } from '../shapefiles/shapefiles.service';
+
+@Controller(`${apiGlobalPrefixes.v1}/planning-units`)
+export class PlanningUnitsController {
+  private readonly logger: Logger = new Logger(PlanningUnitsController.name);
+  constructor(public shapefileService: ShapeFileService) {}
```
kgajowy, line 20: Consider injecting the logger, otherwise it won't leverage global settings (like muting logs)

geoprocessing/package.json
```diff
@@ -66,8 +68,10 @@
     "@types/geojson": "^7946.0.7",
     "@types/jest": "^26.0.15",
     "@types/lodash": "^4.14.168",
+    "@types/multer": "^1.4.5",
     "@types/node": "^14.14.6",
     "@types/supertest": "^2.0.10",
+    "@types/unzipper": "^0.10.3",
```
kgajowy, line 74: We should aim at pinning dependencies ([reading](https://docs.renovatebot.com/dependency-pinning/))

api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -105,6 +114,34 @@ export class ScenariosController {
       await this.service.create(dto, { authenticatedUser: req.user }),
     );
   }
+  // TODO wrap in custom decorator
+  // TODO add Validations
+  @ApiConsumes('multipart/form-data')
```
kgajowy, line 119: TODOs from my patch ; )

kgajowy: also, those are missing in other endpoints (and outputs for Api Doc)

api/src/modules/proxy/proxy.service.ts
```diff
@@ -9,6 +9,9 @@ import * as Server from 'http-proxy';
 
 @Injectable()
 export class ProxyService {
+  private geoprocessingServiceUrl: string = AppConfig.get(
+    'geoprocessing.url',
+  ) as string;
```
kgajowy, line 14: Similar to logger stuff (injecting). Current approach makes it really hard to test it in an unit way. 

([one of the example reading, you can find plenty of those](https://ducmanhphan.github.io/2020-01-15-Understanding-about-SOLID-part-5/))

kgajowy:
> E2E Test related to this func. should be implemented after
> https://vizzuality.atlassian.net/browse/MARXAN-399
> Is done
> 
> Thoughts?

As mentioned somewhere in comments, we should test the parts we already can to ensure the currently implemented pieces are working as intended.

`e2e` test is also fine (we receive the GeoJson, don't we?)

geoprocessing/src/utils/file.utils.ts
alexeh: @kgajowy @types/multer are already present as dev dependecies tho
How can I solve this?

kgajowy:
```

import { diskStorage } from 'multer';

/**
 * Options for Multer
 */
export const uploadOptions: MulterOptions = {
  storage: diskStorage({
    destination: function (
      req,
      file,
      cb: (error: Error | null, destination: string) => void,
    ) {
      cb(null, '/../tmp/');
    },
    filename: (req, file, cb: (error: Error | null, _: string) => void) =>
      cb(null, Date.now() + '-' + file.originalname),
  }),
};

```

kgajowy: image

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -0,0 +1,56 @@
...
+          resolve(`${fileInfo.filename} extracted succesfully`),
+        )
+        .on('error', () =>
+          reject(new Error(`${fileInfo.filename} could not been extracted`)),
+        );
+    });
+  }
+
+  private async _shapeFileToGeoJson(fileInfo: Express.Multer.File) {
+    const outputKey = `shapefile-${new Date().getTime()}.gejson`;
```
hotzevzl, line 30: watch the spelling of the extension

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -0,0 +1,56 @@
...
+    const _geoJson = await mapshaper.applyCommands(
+      `-i ${fileInfo.path.replace('.zip', '')}/*.shp -o ${outputKey}`,
+    );
+
+    return JSON.parse(_geoJson[outputKey].toString('utf-8'));
+  }
+
+  private wipeOutFolder(path: string): void {
+    rmdirSync(path, { recursive: true });
```
hotzevzl, line 39: I am always a bit nervous about `rm -rf` without checks on the path - I would either throw an error if path doesn't start with `/tmp` plus something else, or accept the *name* (not the path) of a temp folder, and prepend `/tmp/whatever` here.

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
hotzevzl: I would suggest to reconsider this approach - we can keep it for this PR, but reading a potentially large file and passing it around forces us to keep it in memory, and depending on context deal with serialization/deserialization/etc - I would consider an approach where we only pass around the path to the geojson file written for us by mapshaper on disk, and then fopen *and stream* that file's contents directly into the HTTP response.

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -0,0 +1,56 @@
...
+  private async _shapeFileToGeoJson(fileInfo: Express.Multer.File) {
+    const outputKey = `shapefile-${new Date().getTime()}.gejson`;
+    const _geoJson = await mapshaper.applyCommands(
+      `-i ${fileInfo.path.replace('.zip', '')}/*.shp -o ${outputKey}`,
+    );
+
+    return JSON.parse(_geoJson[outputKey].toString('utf-8'));
+  }
+
+  private wipeOutFolder(path: string): void {
```
hotzevzl, line 38: very tiny note on naming (as usual from me :grin:) - is this wipe-out folder, or wipe the output folder? I would just use "remove", "delete" or "unlink" folder.

but besides naming (which you can ignore - I just commented because wipe triggers images of bit flipping to remove traces of whatever was written to disk!) - this function is basically a thin wrapper around `rmdirSync()`, so also taking into account my other comment below, I would suggest to make this a sort of shared/utility function for "removing folder with temporary artifacts for job X", assuming that we can stick to a consistent handling of paths for various kinds of file-related jobs and therefore the temp directories we may create to write source files, unzip them where applicable, process them, read results from, etc.

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
kgajowy: @hotzevzl not sure if this is possible to reach the disk. The file is created on `GeoService` module, which would introduce limitations how we deploy units (Api, Geo).  Currently this sends the whole thing over HTTP to API and back to consumer. Probably it would require putting the file on some S3-like storage with short lifespan (to be auto-removed) and provide the link to it back to consumer (or maybe even delegate streaming the file to some cloud function).

Not sure about actual limit (of file size) we have in mind but looks like that different (than current) approach would introduce a lot of things to implement now.

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
alexeh: @hotzevzl I see the point. In this case multer will add the destination folder according to what we decided to pass in `uploadOptions`
I'll add yout first suggestion

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
hotzevzl: @kgajowy to clarify my suggestion: I am thinking about something like this: https://github.com/mbloch/mapshaper/wiki/Using-mapshaper-programmatically#runcommands instead of `applyCommands()`, and then open the GeoJSON file created by mapshaper on disk and stream that back to the response from geoprocessing to api, which will then reverse-proxy this back to the API client.
Using a storage bucket and only returning the signed URL of a short-lived file on a bucket could also be an option, but we'd still need to get the file written to local disk first (if wanting to avoid keeping the whole GeoJSON document in memory) and then use an Azure storage client module to write it to a bucket, sign it, and so on.

In practice I am happy with the current implementation for this PR, of course - no need to add scope to it now!

geoprocessing/src/modules/planning-units/planning-units.controller.ts
alexeh: Following the project's approach. Should we consider DI-n all the loggers in a refactor task?

geoprocessing/src/utils/file.utils.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * File Utils
+ */
+
+import { MulterOptions } from '@nestjs/platform-express/multer/interfaces/multer-options.interface';
+import { Request } from 'express';
+import { diskStorage } from 'multer';
+const tempDirectory = require('temp-dir');
```
kgajowy, line 8: why not import?

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  private async shapeFileToGeoJson(fileInfo: Express.Multer.File) {
+    const outputKey = `shapefile-${new Date().getTime()}.geojson`;
+
+    const _geoJson = await mapshaper.applyCommands(
+      `-i ${fileInfo.path.replace(
+        '.zip',
+        '',
+      )}/*.shp -info -o ${outputKey} -verbose `,
```
kgajowy, line 43: I believe it should support `mute_all_logs` setting.

geoprocessing/src/modules/planning-units/planning-units.controller.ts
kgajowy: We don't have to postpone this very piece until "refactor" and starting doing it "right" now ; )

hotzevzl: @alexeh for context, Kamil and I discussed this in a different PR and converged on injecting this. In general, your is a valid concern - we certainly don't want to be introducing a different way of doing many small things for each piece of code/service, so it is always a good idea to plan an alignment of existing code when we change how we do something in a new component. In this case, you would not have been aware of the DI discussion for loggers, of course.
Given this, in this case it's ok to start injecting the logger here since it's a small change, and we'll take care of aligning the rest in a separate PR (feel free to create a task and take it or bid for takers!).

api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -105,6 +113,36 @@ export class ScenariosController {
...
+        file: {
+          type: 'string',
+          format: 'binary',
+        },
+      },
+    },
+  })
+  @ApiOperation({
+    description: 'Upload Zip file containing .shp, .dbj, .prj and .shx files',
+  })
```
kgajowy, line 132: Pssst 
https://docs.nestjs.com/custom-decorators#decorator-composition

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -0,0 +1,70 @@
...
+        .pipe(
+          Extract({
+            path: `${fileInfo.destination}/${fileInfo.filename.replace(
+              '.zip',
+              '',
+            )}`,
+          }),
+        )
+        .on('close', () =>
+          resolve(`${fileInfo.filename} extracted succesfully`),
```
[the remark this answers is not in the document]
aagm: We were having a problem-related zip extraction, I guess this is part of the debugging work that Alex did.

geoprocessing/src/modules/planning-units/planning-units.controller.ts
alexeh: Changes added in [here](https://github.com/Vizzuality/marxan-cloud/pull/161/commits/11a63a68962ecb3e0793d6fc77c7e223f91b1b94)

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  private unzipShapefile(fileInfo: Express.Multer.File): Promise<string> {
+    return new Promise((resolve, reject) => {
+      createReadStream(fileInfo.path)
+        .pipe(
+          Extract({
+            path: `${fileInfo.destination}/${fileInfo.filename.replace(
+              '.zip',
+              '',
+            )}`,
```
Dyostiq, line 20: to be platform agnostic you may use `import * as path from 'path'` and `path.join(fileInfo.destination, path.basename(fileInfo.filename))`

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -0,0 +1,70 @@
...
+      )}/*.shp -info -o ${outputKey} -verbose `,
+    );
+
+    return JSON.parse(_geoJson[outputKey].toString('utf-8'));
+  }
+
+  private deleteShapefileData(path: string): void {
+    if (path.startsWith('/tmp')) {
+      unlinkSync(path);
+      rmdirSync(path.replace('.zip', ''), { recursive: true });
```
Dyostiq, line 52: if you don't want to block the event loop, you may consider using `unlink` and `rmdir` from `fs/promises`

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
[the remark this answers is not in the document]
alexeh: @kgajowy Actually the message is being logged by the logger in the public method. I added it as part of promisifying the stream in order to be able to `await` for the 'close' event of the stream and then call for other private method instead of using a callback in the same one

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
alexeh: Changes applied [here](https://github.com/Vizzuality/marxan-cloud/pull/161/commits/1cf47128646b7c495378b79d8e2dc76ba563eb83)

geoprocessing/src/modules/planning-units/planning-units.controller.ts
alexeh: Changes applied [here](https://github.com/Vizzuality/marxan-cloud/pull/161/commits/1cf47128646b7c495378b79d8e2dc76ba563eb83)

geoprocessing/src/utils/file.utils.ts
alexeh: Fixed [here](https://github.com/Vizzuality/marxan-cloud/pull/161/commits/47b6937375bcb9d4aeef04854d80b4d8be26ad29)

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
alexeh: Deleted verbose flag [here](https://github.com/Vizzuality/marxan-cloud/pull/161/commits/8dee9241dc41373ccd309bbd3d104b72f93e10ea)

geoprocessing/src/utils/file.utils.ts
alexeh: This was nice.

Fixed [here](https://github.com/Vizzuality/marxan-cloud/pull/161/commits/05d64f5ac08039cd5cef8d5b626f8418762d840d)

hotzevzl:
> E2E Test related to this func. should be implemented after 
> https://vizzuality.atlassian.net/browse/MARXAN-399
> Is done
> 
> Thoughts? 

I would recommend to implement e2e tests as part of 399 (can be a PR just for tests followed by the PR for the implementation of validations, or following it). Could this work?

PR #162, opened 2021-05-10
feat(api): scenarios: wdpa calculation trigger

kgajowy:
## Trigger WDPA calculations on create/update of a Scenario according to conditions

Based on:
```
2º IF wdpaIUCNCategories / customProtectedAreasIds and wdpaThershold has a value (created or updated) we do trigger the next job that executes the wdpa area calculation 
```

### Overview

"TODO" should be filled with [command from this PR](https://github.com/Vizzuality/marxan-cloud/pull/153 )

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/src/modules/scenarios/wdpa-arena-calculation.service.ts
+++ b/api/src/modules/scenarios/wdpa-arena-calculation.service.ts
@@ -21,22 +21,13 @@ export class WdpaArenaCalculationService {
    * and new entity state consists (i.e. are present) of all of the below
    */
   private readonly requiredToTriggerChange: (keyof Scenario)[] = [
-    'wdpaIucnCategories',
-    'protectedAreaFilterByIds',
     'wdpaThreshold',
   ];
 
   /**
-   * 2º IF (x) has a value
-   * - wdpaIUCNCategories
-   * - customProtectedAreasIds
-   * - wdpaThershold
-   *
-   * we do trigger the next job
-   * @private
+   * Every post update that affects this 3 elements (when threshold is not null)
    */
   shouldTrigger(scenario: Scenario, changeSet: InputChange): boolean {
-    // verify if any of the "wanted" value was provided/changed
     if (!this.intendsToChangeWatchedProperty(changeSet)) {
       return false;
     }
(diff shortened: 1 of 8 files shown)


api/src/modules/scenarios/wdpa-arena-calculation.service.ts
```diff
@@ -0,0 +1,49 @@
+import { Injectable } from '@nestjs/common';
+import { pick } from 'lodash';
+import { CreateScenarioDTO } from './dto/create.scenario.dto';
+import { UpdateScenarioDTO } from './dto/update.scenario.dto';
+import { Scenario } from './scenario.api.entity';
+
+type InputChange = CreateScenarioDTO | UpdateScenarioDTO;
+
+@Injectable()
+export class WdpaArenaCalculationService {
```
hotzevzl, line 10: after having squinted blankly at this class name for a while, I think I was trying to read some meaning which is not there - I suppose this is a gladiator misspelling for the name below, right?

```suggestion
export class WdpaAreaCalculationService {
```

api/src/modules/scenarios/wdpa-arena-calculation.service.ts
```diff
@@ -0,0 +1,49 @@
...
+    return Object.entries(pick(changeSet, this.watchedChangeProperties)).some(
+      ([, value]) => value,
+    );
+  }
+
+  private areRequiredFieldsAvailable(scenario: Scenario): boolean {
+    return Object.entries(pick(scenario, this.requiredToTriggerChange)).every(
+      ([, value]) => value,
+    );
+  }
```
hotzevzl, line 48: these seem like good candidates to move to utility functions (when we need these kinds of checks for a second context) - just a note, nothing to do now

api/src/modules/scenarios/wdpa-arena-calculation.service.ts
kgajowy: fixed in https://github.com/Vizzuality/marxan-cloud/pull/162/commits/87c647950b6545a6abaff37d4f154b5347f6344a

PR #174, opened 2021-05-13
chore(api): scenario cost surfaces shell

kgajowy:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

https://vizzuality.atlassian.net/browse/MARXAN-291

---

- [ ] Update CHANGELOG file


api/src/decorators/shapefile.decorator.ts
```diff
@@ -8,6 +8,26 @@ import {
 
 import { ShapefileGeoJSONResponseDTO } from 'modules/scenarios/dto/shapefile.geojson.response.dto';
 
+export function ApiConsumesShape() {
```
hotzevzl, line 11: If the only current and expected difference from `@ApiConsumesShapefile()` is that there would be no geojson in the response here, could we make this a parameter, or change `@ApiConsumesShapefile()` to be a combination of this other decorator and `@ApiOkResponse()`?

Either way, if we want to have two distinct decorators, we should clarify their name a bit or someone looking at them (even ourselves when applying them to new endpoints) will need to dig into the inner `applyDecorators()` to understand what is the difference.

api/src/modules/scenarios/cost-surface/cost-surface.facade.ts
```diff
@@ -0,0 +1,40 @@
...
+    private readonly proxyService: ProxyService,
+  ) {}
+
+  /**
+   * non blocking - will do job in "background"
+   *
+   * @param scenarioId
+   * @param request
+   */
+  convert(scenarioId: string, request: Request) {
```
hotzevzl, line 19: Is the intention to make this `async` but only placing it on the event loop without awaiting for it in the controller?

api/src/modules/scenarios/cost-surface/cost-surface.facade.ts
```diff
@@ -0,0 +1,40 @@
...
+    // TODO #1 Call Proxy Service to get GeoJSON
+    // this.proxyService.... - modify this to send back data, not act on Response
+
+    // TODO #2 Call Analysis-module with scenario id & GeoJson from above
+    this.adjustCostSurfaceService
+      .update(scenarioId, {
+        geo: {
+          features: [],
+          type: 'FeatureCollection',
+        },
```
hotzevzl, line 31: this could probably be a plain list of ids and associated cost value

hotzevzl (commented):
@kgajowy thank you. The shell looks ok. I would still prefer to have geo-jobs be self-contained in the geoprocessing service, especially where we don't need to send back anything to the API client, but between ApiEvents and other bits, it's true that there is quite some code we'd need to share or (heavens forbid) duplicate between API and geoprocessing service.

In this case, maybe we could simplify the process a bit by handling just a bit more than shp->geojson transformation in the geoprocessing service: we'll only need the id of each planning unit and the associated cost, so we can get the geoprocessing service to return that, and get back a (potentially) much smaller payload

api/src/modules/scenarios/cost-surface/cost-surface.facade.ts
kgajowy: yes, assuming we will do some processing there.

api/src/modules/scenarios/cost-surface/cost-surface.facade.ts
kgajowy: I would refrain from putting more logic there. We already have `analysis module` which handles validation of integrity and persistence - the facade is "only" connecting those two pieces together (i.e. rest->shapefile->geo->process)

api/src/decorators/shapefile.decorator.ts
kgajowy: Adjusted in https://github.com/Vizzuality/marxan-cloud/pull/174/commits/2a3b01804529a4fd1af49ecc3a703616d8547247

api/src/modules/scenarios/cost-surface/cost-surface.facade.ts
kgajowy: i.e. there is no point in waiting within a controller as this is a "schedule job"-alike and we will process a file in an async way (but we cannot guarantee any reliable waiting time)

api/src/modules/scenarios/cost-surface/cost-surface.facade.ts
kgajowy: Revisited in https://github.com/Vizzuality/marxan-cloud/pull/174/commits/1a12fbbe7d4bdf4df095a88d5dd69ced127a860c

hotzevzl (approved):

PR #180, opened 2021-05-13
API development workflow proposal

hotzevzl:
Replaces https://github.com/Vizzuality/marxan-cloud/pull/168


hotzevzl:
@aagm merging this as is, and I have created a task for myself to add the pointers to examples as you suggested: https://vizzuality.atlassian.net/browse/MARXAN-402

PR #182, opened 2021-05-14
add planning unit vector tile preview - rebased

hotzevzl:
Supercedes https://github.com/Vizzuality/marxan-cloud/pull/155

Rebased on latest `develop` after some history rewrite.

### DIFF
--- a/api/src/modules/proxy/proxy.service.ts
+++ b/api/src/modules/proxy/proxy.service.ts
@@ -23,11 +23,6 @@ export class ProxyService {
    */
 
   proxyTileRequest(request: Request, response: Response) {
-    // Use `originalUrl` here, as the `/api/v1/geoprocessing` is otherwise stripped
-    // away, but we need this because the backend microservice expects that as
-    // part of its routes
-    request.url = request.originalUrl.replace('api/v1/', 'api/v1/geodata/');
-
     return this.server.web(
       request,
       response,
@@ -49,7 +44,7 @@ export class ProxyService {
   ): Promise<any> {
     request.url = request.originalUrl.replace(
       'api/v1/scenarios',
-      'api/v1/geodata/planning-units',
+      'api/v1/planning-units',
     );
 
     return this.server.web(
--- a/geoprocessing/src/api.config.ts
+++ b/geoprocessing/src/api.config.ts
@@ -5,5 +5,5 @@
  * the relevant controllers.
  */
 export const apiGlobalPrefixes = {
-  v1: '/api/v1/geodata',
+  v1: '/api/v1',
 };
--- a/geoprocessing/src/filters/all-exceptions.exception.filter.ts
+++ b/geoprocessing/src/filters/all-exceptions.exception.filter.ts
@@ -67,12 +67,13 @@ export class AllExceptionsFilter implements ExceptionFilter {
         ? omit(errorData, ['meta.rawError', 'meta.stack'])
         : errorData,
     );
-
+  
     response.removeHeader('content-encoding');
     response
       .status(status)
       .header('Content-Type', 'application/json')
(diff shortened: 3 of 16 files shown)


geoprocessing/src/api.config.ts
```diff
@@ -5,5 +5,5 @@
  * the relevant controllers.
  */
 export const apiGlobalPrefixes = {
-  v1: '/api/v1/geodata',
```
elpamart, line 8: are we removing the 'geodata' from the geoprocessing endpoints? I guess this is the reason why you have updated the replace from the proxy.

api/src/modules/proxy/proxy.service.ts
```diff
@@ -23,11 +23,6 @@ export class ProxyService {
    */
 
   proxyTileRequest(request: Request, response: Response) {
-    // Use `originalUrl` here, as the `/api/v1/geoprocessing` is otherwise stripped
-    // away, but we need this because the backend microservice expects that as
-    // part of its routes
-    request.url = request.originalUrl.replace('api/v1/', 'api/v1/geodata/');
-
     return this.server.web(
```
elpamart, line 26: these changes are different from the ones I had in my PR. I guess is a change due to the decision of removing 'geodata' from the geoprocessing endpoints?

geoprocessing/src/filters/all-exceptions.exception.filter.ts
```diff
@@ -67,11 +67,11 @@ export class AllExceptionsFilter implements ExceptionFilter {
         ? omit(errorData, ['meta.rawError', 'meta.stack'])
         : errorData,
     );
-
+    response.removeHeader('content-encoding');
     response
       .status(status)
       .header('Content-Type', 'application/json')
-      .header('Content-Disposition', 'inline')
+      .header('content-disposition', 'inline')
```
elpamart, line 74: not sure if the uppercase affects here (e.g 'Content-disposition vs content-disposition). We have this in uppercase in all the controllers endpoints.

elpamart (approved):
I see some small changes but are improvements from the original code. We have also combined in this PR the shapefile upload that Alex did but I guess should be ok. From my side everything looks good!

geoprocessing/src/filters/all-exceptions.exception.filter.ts
hotzevzl: I found this change in the commits of the old branch (was authored by Alicia IIRC), in practice headers are case-insensitive so I left it as it was, but I'll change to stick to our usual convention.

api/src/modules/proxy/proxy.service.ts
hotzevzl: yep, this is my suggestion to drop the prefix (which IIRC was my suggestion to use to start with, sorry about the doing and undoing 😬) from the geoprocessing service and to use basically the same paths as we do on the API "proxy-side".

Originally I wasn't too sure how we'd use the geoprocessing service, so my idea for the `geodata` part of the URLs was to group some endpoints together, but after all everything is pretty much related to geodata here, so I'd suggest to drop it and use identical paths in the API and in the geoprocessing service (unless we need to use different paths for specific future endpoints).

elpamart: makes sense to me!

hotzevzl: ok, thanks, and sorry about my change of mind 😱

geoprocessing/src/filters/all-exceptions.exception.filter.ts
hotzevzl: fixed in 857d81a86f28224b0bbb737ebd979bdafb37698d

PR #183, opened 2021-05-14
feat(api): protect api-events with api-key

kgajowy:
### Overview

It could be necessary to provide relevant envs on CI/staging/production

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


hotzevzl (approved):
@kgajowy thank you, looks great.

hotzevzl:
actually, may it make sense to get this prop from env only once at app startup, rather than per-request in the guard?

kgajowy:
> actually, may it make sense to get this prop from env only once at app startup, rather than per-request in the guard?

Could be, if we are fine to restart the deployed unit if we need to change the secret ;)

fixed in https://github.com/Vizzuality/marxan-cloud/pull/183/commits/700096169e0e73b90f090d90c402bfe14bc27708

hotzevzl:
> > actually, may it make sense to get this prop from env only once at app startup, rather than per-request in the guard?
> 
> Could be, if we are fine to restart the deployed unit if we need to change the secret ;)
> 
> fixed in [7000961](https://github.com/Vizzuality/marxan-cloud/commit/700096169e0e73b90f090d90c402bfe14bc27708)

thanks, and yes, no problem in having to restart - this will rarely change in most cases.

PR #185, opened 2021-05-14
Legend types and maxHeight

mbarrenechea:
### Overview

This PR adds:
- Legend types

  - Choropleth
  - Basic
  - Gradient

- Max height

### DIFF
--- a/app/components/map/legend/component.stories.tsx
+++ b/app/components/map/legend/component.stories.tsx
@@ -3,6 +3,11 @@ import { Story } from '@storybook/react/types-6-0';
 
 import Legend, { LegendProps } from './component';
 import LegendItem from './item';
+
+import LegendTypeBasic from './types/basic';
+import LegendTypeChoropleth from './types/choropleth';
+import LegendTypeGradient from './types/gradient';
+
 import ITEMS from './mock';
 
 export default {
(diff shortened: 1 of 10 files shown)


app/components/map/legend/component.stories.tsx
```diff
@@ -27,14 +32,22 @@ const Template: Story<LegendProps> = (args) => {
...
+          <LegendItem
+            key={i.id}
+            {...i}
+          >
+            {type === 'basic' && <LegendTypeBasic className="text-sm text-gray-300" items={items} />}
+            {type === 'choropleth' && <LegendTypeChoropleth className="text-sm text-gray-300" items={items} />}
+            {type === 'gradient' && <LegendTypeGradient className="text-sm text-gray-300" items={items} />}
+          </LegendItem>
+        );
+      })}
```
kgajowy, line 50:
```
// pseudocode
{sortedItems.map(({type, items, id, ...rest}) => <LegendItem
            key={i.id} // wouldn't  type be better?
            {...rest}
          >
            {type === 'basic' && <LegendTypeBasic className="text-sm text-gray-300" items={items} />}
            {type === 'choropleth' && <LegendTypeChoropleth className="text-sm text-gray-300" items={items} />}
            {type === 'gradient' && <LegendTypeGradient className="text-sm text-gray-300" items={items} />}
          </LegendItem>
)}
```

app/components/map/legend/types/gradient/component.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import cx from 'classnames';
+
+export interface LegendTypeGradientProps {
+  className?: string;
+  items: Array<{
+    value: string;
+    color: string;
+  }>;
+}
```
kgajowy, line 10: Any reason why legends cannot share the same interface?

app/components/map/legend/component.stories.tsx
mbarrenechea: _**"wouldn't  type be better?"**_
Not at all. If you have legend items of the same type you will have the same keys for different items. Not desirable

app/components/map/legend/types/gradient/component.tsx
mbarrenechea: It depends on the type. It's true that now they have the same interface, but I could imagine basic type having their own items. For example, you could define the icon for basic ones. That's why I want to keep them separate.

mbarrenechea:
@kgajowy thanks for reviewing!!

app/components/map/legend/component.stories.tsx
kgajowy: Ah indeed, I wrongly assumed that only those 3 will be present (i.e. it couldn't have multiple `basic` types)

app/components/map/legend/types/gradient/component.tsx
kgajowy: @mbarrenechea sure thing, they can just extend base later on - but they still have the same base ; ) Even tho they have different name, TS consider them to be exactly the same.

PR #187, opened 2021-05-14
feat(geoprocessing): api-events wrapper

kgajowy:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/geoprocessing/src/modules/api-events/api-events.service.ts
+++ b/geoprocessing/src/modules/api-events/api-events.service.ts
@@ -1,0 +1,42 @@
+import { HttpService, Injectable } from '@nestjs/common';
+import { API_EVENT_KINDS } from './events.enum';
+import { AppConfig } from '../../utils/config.utils';
+
+@Injectable()
+export class ApiEventsService {
+  readonly #secret: string;
+  readonly #apiUrl: string;
+
+  constructor(private readonly http: HttpService) {
+    // TODO debt: config shall be injected (nestjs/config); it isn't really unit-testable
+    // will throw if not provided
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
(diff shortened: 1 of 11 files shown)


geoprocessing/src/modules/api-events/api-events.service.ts
```diff
@@ -0,0 +1,42 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    this.#apiUrl = AppConfig.get<string>('api.url')!;
+  }
+
+  async create<T>(
+    resourceId: string,
+    kind: API_EVENT_KINDS,
+    data?: T,
+  ): Promise<void> {
+    // TODO what if it failed? (currently validateStatus "swallows" the error)
```
hotzevzl, line 24: I would suggest to let `validateStatus()` propagate an error for responses >= 400 and then further propagate this up from `create()` here, so that the caller can decide whether to go ahead or stop/rollback/etc. as appropriate. Can be for a future PR.

hotzevzl (approved):
thanks @kgajowy, I think we can add propagation of errors from the ApiEvents on the API side later - feel free to suggest something different from my own suggestion which I left inline in the review.

I'll add configuration of the endpoint URL in a later PR too.

geoprocessing/src/modules/api-events/api-events.service.ts
kgajowy: @hotzevzl the point is that I am not really sure if we should propagate the whole error in its natural form. Why would consumers need to know details of some endpoint?

kgajowy: created https://vizzuality.atlassian.net/browse/MARXAN-405

PR #192, opened 2021-05-17
Refactior File Service

alexeh:
Refactior File Service

This PR introduces a refactor to create stand-alone File Service.

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -1,35 +1,17 @@
...
-          ),
-        );
-    });
+export class ShapefileService {
+  constructor(
+    private readonly logger: Logger,
+    private fileService: FileService,
+  ) {
+    this.logger.setContext(ShapefileService.name);
+    this.fileService = new FileService();
```
kgajowy, line 14: Why are we bypassing NestJS's DI ?

geoprocessing/src/utils/file.utils.ts
```diff
@@ -25,3 +29,36 @@ export const uploadOptions: Readonly<MulterOptions> = {
     },
   }),
 };
+
+export class FileService {
```
kgajowy, line 33: Looks like it should be `Injectable()` and wrapped in some module instead

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -42,23 +24,14 @@ export class ShapeFileService {
...
-      this.logger.log(await this.unzipShapefile(shapeFile));
+      this.logger.log(await this.fileService.unzipFile(shapeFile));
     } catch (err) {
       this.logger.error(err);
     }
     const geoJson = await this.shapeFileToGeoJson(shapeFile).then((geoJson) => {
-      this.deleteShapefileData(shapeFile.path);
+      this.fileService.deleteDataFromFS(shapeFile.path);
       return geoJson;
     });
```
kgajowy, line 36:
```
const geoJson = await this.shapeFileToGeoJson(shapeFile)
/** await ? */ this.fileService.deleteDataFromFS(shapeFile.path)
return geoJson
```

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -42,23 +24,14 @@ export class ShapeFileService {
...
-  private deleteShapefileData(path: string): void {
-    if (path.startsWith('/tmp')) {
-      unlinkSync(path);
-      rmdirSync(path.replace('.zip', ''), { recursive: true });
-    } else {
-      throw new Error(`Could not complete deletion: ${path} is not in /tmp`);
-    }
-  }
-
   async getGeoJson(shapeFile: Express.Multer.File) {
```
kgajowy, line 27: Would be great to see some unit tests with "simple" shapefile (both valid and invalid)

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -1,35 +1,17 @@
...
-            new Error(`${fileInfo.filename} could not be extracted: ` + error),
-          ),
-        );
-    });
+export class ShapefileService {
+  constructor(
+    private readonly logger: Logger,
+    private fileService: FileService,
+  ) {
+    this.logger.setContext(ShapefileService.name);
```
Dyostiq, line 13: As the NestJS injects the logger as a singleton (in `PlanningUnitsModule` given as a `ClassProvider`) – then it changes the `context` of the logger in all services of the module.

geoprocessing/src/utils/file.utils.ts
```diff
@@ -25,3 +29,36 @@ export const uploadOptions: Readonly<MulterOptions> = {
...
+            new Error(`${fileInfo.filename} could not be extracted: ` + error),
+          ),
+        );
+    });
+  }
+
+  deleteDataFromFS(path: string): void {
+    if (path.startsWith('/tmp')) {
+      unlink(path);
+      rmdir(path.replace('.zip', ''), { recursive: true });
```
Dyostiq, line 59: it should be awaited, and usages of `deleteDataFromFS` should at least handle rejections

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
alexeh: @kgajowy 1d15b56c1ef8c605a970984b80f2616a35a65f64

geoprocessing/src/utils/file.utils.ts
alexeh: @kgajowy 1d15b56c1ef8c605a970984b80f2616a35a65f64

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
kgajowy: @alexeh 
```
 @Inject(FileService)    private fileService: FileService
```

Should be enough to `private readonly fileService: FileService` as long as you import the FileModule

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
alexeh: @Dyostiq Totally right, I just decided to delete contextsetting and let the controller override it, 1105fe35e9b4ce2eeb8265172f46496f055d8978

Let me know yout thoughts, pls 😃

geoprocessing/src/utils/file.utils.ts
alexeh: @Dyostiq 1105fe35e9b4ce2eeb8265172f46496f055d8978

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
alexeh: @kgajowy 1105fe35e9b4ce2eeb8265172f46496f055d8978

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
alexeh: @kgajowy 1105fe35e9b4ce2eeb8265172f46496f055d8978

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -42,25 +19,20 @@ export class ShapeFileService {
...
-      return geoJson;
-    });
+    const geoJson = await this.shapeFileToGeoJson(shapeFile).then(
+      async (geoJson) => {
+        await this.fileService
+          .deleteDataFromFS(shapeFile.path)
+          .catch((error) => error);
+        return geoJson;
+      },
+    );
```
Dyostiq, line 35: it's a bit hard to read and seems to be unnecessarily nested, I'd suggest this:
```suggestion
    const geoJson = await this.shapeFileToGeoJson(shapeFile);
    await this.fileService
      .deleteDataFromFS(shapeFile.path)
      .catch((error) => error);
```

Also, `catch((error) => error);` is not the best option, as we will miss issues with file storage. I would at least try to log it, preferably with an `error` log level with some context of the error.

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
Dyostiq: A controller overwriting a context is a bit surprising and very implicit. It will get worse with multiple controllers in a module. Honestly, I would stay with `private readonly logger = new Logger(this.constructor.name)` in every class. [Muting loggers is then still possible, and NestJS handles it as soon as we don't use static methods]. With a logger as a singleton, we should not set the `context` manually, and provide if with every log. If you insist injecting it, the solution is to change the provider to provide a constructor or a factory for the logger, not the instance.

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
alexeh: I agree 7c1a76eee6b4e7692467d485d9cae6b487d3d8b0

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
alexeh: @Dyostiq 01859c0bba0333574d75839c6849ff5a3dd27a4b

kgajowy (approved):

aagm (commented):
The conversion of shp to geojson works as a charm, even for very large and complex shapefiles, congratulations!. 
Said this; I'm seeing a lot of problems with invalid shapefiles. For some reason is not taking into account missing files like .proj which leads to an erroneous projection output.
is not validating file upload sizes, and geojson allowed types or correctness of geometry (yet right?).

PR #193, opened 2021-05-18
feat(api): analysis: cost-surface: validate pu-ids

kgajowy:
* chore(api): analysis: move are-puuids-allowed to shared provider
* feat(api): analysis: cost-surface: validate pu-ids


aagm (approved):
validation looks right, i will suggest merge

PR #194, opened 2021-05-18
fix(geo): e2e specs

kgajowy:
* In the end, closing db/redis connections manually still keeps something under the hood. Judging from pending handlers it still looks like Redis from queues. Possibly could be revisited later.
* Fixed a few tests.

- [x] `test/api-events.service.integration.e2e-spec.ts` needs env in CI environment to work correctly
- [ ] needs a fix to exit with error if any test fails

### DIFF
--- a/.github/workflows/api-e2e-tests.yml
+++ b/.github/workflows/api-e2e-tests.yml
@@ -11,6 +11,7 @@ jobs:
     timeout-minutes: 15
     env:
       API_SERVICE_PORT: 3130
+      API_SERVICE_URL: http://api:3000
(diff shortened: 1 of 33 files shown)


hotzevzl (approved):
@kgajowy works great, thanks! I have added the `API_SERVICE_URL` envvar and fixed another bug (`NODE_ENV=test` would not have the correct redis host name configured).

I would merge, and get on the issue of propagating a failure in tests upstream in a separate PR. It is not hard at all, but requires a few steps and caveats to make sure we still clean things up properly in case of failure when running locally (in GH Actions we can avoid teardown as the runner gets destroyed upstream in any case)

PR #196, opened 2021-05-18
feat(api): cost-surface: application logic

kgajowy:
### Overview

Part of `Users should be able to upload shapefiles of scenario cost surfaces` - subtask https://vizzuality.atlassian.net/browse/MARXAN-413

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/src/modules/scenarios/cost-surface/__mocks__/cost-surface-events-fake.ts
+++ b/api/src/modules/scenarios/cost-surface/__mocks__/cost-surface-events-fake.ts
@@ -1,0 +1,20 @@
+import { Injectable, Scope } from '@nestjs/common';
+import {
+  CostSurfaceEventsPort,
+  CostSurfaceState,
+} from '../cost-surface-events.port';
+
+@Injectable()
+export class CostSurfaceEventsFake implements CostSurfaceEventsPort {
+  mock = jest.fn();
+  events: [string, CostSurfaceState][] = [];
+
+  async event(
+    scenarioId: string,
+    state: CostSurfaceState,
+    context?: Record<string, unknown> | Error,
+  ): Promise<void> {
+    this.events.push([scenarioId, state]);
+    return this.mock(scenarioId, state, context);
+  }
+}
--- a/api/src/modules/scenarios/cost-surface/cost-surface.facade.spec.ts
+++ b/api/src/modules/scenarios/cost-surface/cost-surface.facade.spec.ts
@@ -1,0 +1,141 @@
+import { CostSurfaceFacade } from './cost-surface.facade';
+import { Test } from '@nestjs/testing';
+import { Request } from 'express';
+import { AdjustCostSurface } from '../../analysis/entry-points/adjust-cost-surface';
+import { AdjustCostServiceFake } from './__mocks__/adjust-cost-service-fake';
+import { ResolvePuWithCost } from './resolve-pu-with-cost';
+import { ShapefileConverterFake } from './__mocks__/shapefile-converter-fake';
+import { CostSurfaceEventsPort } from './cost-surface-events.port';
+import { CostSurfaceEventsFake } from './__mocks__/cost-surface-events-fake';
+import { getValidSurfaceCost } from './__mocks__/surface-cost.data';
+
+let sut: CostSurfaceFacade;
+
+let costService: AdjustCostServiceFake;
+let fileConverter: ShapefileConverterFake;
+let events: CostSurfaceEventsFake;
+
+beforeEach(async () => {
+  const sandbox = await Test.createTestingModule({
+    providers: [
+      {
+        provide: AdjustCostSurface,
+        useClass: AdjustCostServiceFake,
+      },
+      {
+        provide: ResolvePuWithCost,
+        useClass: ShapefileConverterFake,
+      },
+      {
+        provide: CostSurfaceEventsPort,
+        useClass: CostSurfaceEventsFake,
+      },
+      CostSurfaceFacade,
+    ],
+  }).compile();
+
+  sut = sandbox.get(CostSurfaceFacade);
+  costService = sandbox.get(AdjustCostSurface);
+  fileConverter = sandbox.get(ResolvePuWithCost);
+  events = sandbox.get(CostSurfaceEventsPort);
+});
+
+const scenarioId = 'scenarioId';
+const request: Request = Object.freeze({
+  file: ({ fakeFile: 1 } as unknown) as Express.Multer.File,
+} as unknown) as Request;
+
+describe(`when file couldn't be converted`, () => {
+  beforeEach(async () => {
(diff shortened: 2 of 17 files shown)


api/src/modules/scenarios/cost-surface/cost-surface.facade.ts
```diff
@@ -1,42 +1,44 @@
...
-  convert(scenarioId: string, request: Request) {
-    // TODO #0 Generate & Dispatch Api Event (some wrapping service for /dummy/"terminating" if already running)
+  async convert(scenarioId: string, request: Request) {
+    await this.events.event(scenarioId, CostSurfaceState.Submitted);
+    let costSurface: CostSurfaceInputDto;
 
-    // TODO #1 Call Proxy Service to get Planning Units and their surface cost
-    // this.proxyService.... - modify this to send back data, not act on Response
+    try {
+      costSurface = await this.shapefileConverter.fromShapefile(request.file);
```
Dyostiq, line 27: The comment says it is not blocking (I understand it as the user should not await the returned promise). Actually, it is a bit – event dispatching is awaited (as well as a conversion), only the update below is not. Also, the inferred type of the method is `Promise`, and it's what a linter may complain about, and be misleading.

Theoretically, it may also result in an unhandled promise rejection.

What I would do is to just use `void` as a return type, and the entire `async` code I would put into `Promise.resolve().then(async () => {/*...*/}).catch(() => {/* swallowing/logging exception */})` or an async IIFE (`(async () => {/*...*/})().catch(() => {/* swallowing/logging exception */})`). It would allow to use normal `await`s while updating.

Dyostiq: or make it a normal async method and let the user of the method decide how to handle it

api/src/modules/scenarios/cost-surface/cost-surface.facade.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+    });
+
+    it(`proxies file to port`, () => {
+      expect(fileConverter.mock.mock.calls[0][0]).toEqual(request.file);
+    });
+
+    it(`proxies port output to update service`, () => {
+      expect(costService.mock.mock.calls[0][0]).toEqual(scenarioId);
+      expect(costService.mock.mock.calls[0][1]).toEqual(getValidSurfaceCost());
+    });
```
Dyostiq, line 120:
```suggestion
    it(`updates the const with a converted cost`, () => {
      expect(costService.mock.mock.calls[0][0]).toEqual(scenarioId);
      expect(costService.mock.mock.calls[0][1]).toEqual(getValidSurfaceCost());
    });
```

the second assertion already covers that the converter is used – otherwise, it wouldn't get `getValidSurfaceCost`.

kgajowy: @Dyostiq indeed it covers usage of the converter - but the assertion you suggest to remove checks if the file was passed as expected 🤔

api/src/modules/scenarios/cost-surface/__mocks__/cost-surface-events-fake.ts
```diff
@@ -0,0 +1,20 @@
+import { Injectable, Scope } from '@nestjs/common';
+import {
+  CostSurfaceEventsPort,
+  CostSurfaceState,
+} from '../cost-surface-events.port';
+
+@Injectable()
```
alexeh, line 7: Little clarifications. The reason to making Mock services @Injectable is no get them as provider when you create sandboxes or SUT's?

kgajowy: @alexeh yes - to let them be initialized by NestJS DI.

api/src/modules/scenarios/cost-surface/cost-surface.facade.ts
kgajowy: Idea was to "you can, but you don't have to and won't get anything in return", thanks for pointing out missing piece!

api/src/modules/scenarios/cost-surface/cost-surface.facade.spec.ts
Dyostiq: oh, ok

I'd modify the mock then, to return `getValidSurfaceCost` only for the file given, or keep it as is now

api/src/modules/scenarios/cost-surface/cost-surface.facade.ts
kgajowy: @Dyostiq https://github.com/Vizzuality/marxan-cloud/pull/196/commits/fef830aa4a16397bc40f76209dd3bf945f7db2c1

Dyostiq (approved):

api/src/modules/scenarios/cost-surface/cost-surface.facade.ts
```diff
@@ -1,42 +1,44 @@
...
+  CostSurfaceState,
+} from './cost-surface-events.port';
+import { CostSurfaceInputDto } from '../../analysis/entry-points/adjust-cost-surface-input';
 
 @Injectable()
 export class CostSurfaceFacade {
   constructor(
     private readonly adjustCostSurfaceService: AdjustCostSurface,
-    private readonly proxyService: ProxyService,
+    private readonly shapefileConverter: ResolvePuWithCost,
```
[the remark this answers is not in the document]
alexeh: @kgajowy Sorry, I forgot to click 'Start review...'

aagm (approved):

api/src/modules/scenarios/cost-surface/resolve-pu-with-cost.ts
```diff
@@ -0,0 +1,7 @@
+import { CostSurfaceInputDto } from '../../analysis/entry-points/adjust-cost-surface-input';
+
+export abstract class ResolvePuWithCost {
+  abstract fromShapefile(
```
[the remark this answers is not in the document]
kgajowy: Its just an interface what main logic will use.

api/src/modules/scenarios/cost-surface/cost-surface.facade.ts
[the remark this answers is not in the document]
kgajowy: @alexeh its input is a shapefile and returns base PU ids with cost. Suitable names are always appreciated ;)

api/src/modules/scenarios/cost-surface/resolve-pu-with-cost.ts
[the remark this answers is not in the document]
kgajowy: `Dependency Inversion Principle`

PR #198, opened 2021-05-19
feat(api): product testing with cypress

kgajowy:
Exact steps and their implementation is subject to change once those are provided. Current ones should be considered only as an example.


e2e-product-testing/cypress/integration/LogIn.feature
```diff
@@ -0,0 +1,10 @@
+Feature: LogIn
+
+  I want to log in
+
+  Scenario: Logging in
```
alexeh, line 5: Even tho I love Gherkin/Cucumber, I'm not sure if it makes sense adding these, because would imply following a BDD approach, and that would be a team decision which involves both BE/FE

e2e-product-testing/cypress/integration/common/pages.ts
```diff
@@ -0,0 +1,9 @@
+import { Given } from "cypress-cucumber-preprocessor/steps";
+
+Given("I am on a main page", () => {
+  cy.visit("https://marxan.vercel.app/", {
```
alexeh, line 4: I'm personally not a fan on relying e2e tests that are triggered as part of a CI on an instance deployed somewhere else
Also, if FE (or BE) adds a new e2e test, that would fail until a new develop is deployed (which is happening manually right know I think), so I think is not a good option (yet) to have it as a CI stage

e2e-product-testing/cypress/integration/LogIn.feature
kgajowy: @alexeh 
https://vizzuality.atlassian.net/browse/MARXAN-415#icft=MARXAN-415
https://vizzuality.atlassian.net/browse/MARXAN-408#icft=MARXAN-408

We can keep it w/o Gherkin but I believe it adds a great value. 

Product-testing does not imply following BDD approach in underlying parts of the systen.

e2e-product-testing/cypress/integration/common/pages.ts
kgajowy: This is a typical black-box testing of deployed product as a whole.

It is a "golden target" to write requirements in this form and complete feature (merge it) when its fully done. We may not require this test to be present on develop/production until feature is ready (or mark it as "can fail").

What we had in mind is to ensure that most business-critical features *works*. Writing such for future/upcoming features is a different story.

e2e-product-testing/cypress/integration/LogIn.feature
alexeh: @kgajowy Product testing as is does not imply following BDD, **using Gherkin it does**, and It's my opinion that we should be consistent.
Product/e2e tests need to be maintained, and as we go adding more features, or more use cases that we want to be covered by these test, we should all follow the same path.
As I said, I personally like this approach, but I think we should agree as a team this approach, because implies using Gherkin and Cucumber.

But this is a very personal opinion, and since I'm a junior newcomer, I'd let FE/BE people decide

e2e-product-testing/cypress/integration/common/pages.ts
alexeh: `This is a typical black-box testing of deployed product as a whole.`

The project every time gets built is the product as a whole.

I just think that since this is a part of the project, it would make more sense to be targeted to the project itself, as part of a CI stage. Otherwise I think it would make more sense having a standalone project that runs against a deployed instance, and just take care of having them in sync.

Very personal opinion tho

e2e-product-testing/cypress/integration/LogIn.feature
hotzevzl: my personal opinion - product testing seems to me to be a context where Gherkin syntax can greatly help to create scenarios with consistent structure and clarity.

As long as there are no concerns from developers and scientists about using this syntax for specs of product tests, I suggest to stick to this.

@aagm @mbarrenechea @tamaramegan @Dyostiq what do you think?

This could also serve as a testbed for a possible switch to Gherking syntax in API/geoprocessing e2e tests if we wanted to consider that. I would not hold back using this in a new context such as the one started with this PR because we are not consistent across test types - even if we did not want or were not able (time, etc.) to Gherkinize e2e test, even just the new ones forgetting about the already-written ones, I believe using this for this new area of testing would add a lot of value.

e2e-product-testing/cypress/integration/common/pages.ts
hotzevzl: I think we could start with what we have, and adjust as we change our deployment workflow.

We should actually be quite close to being able to run the whole product on GH Actions (https://github.com/Vizzuality/marxan-cloud/pull/118, which I need to resurrect from the recent mass-closure of PRs), at which stage we could target the local (GH Actions) app.

If we want to stick with the initial idea of running these tests against a staging environment, though, that would still be separate from GH Actions, and I would still like to use GH Actions to run these tests to avoid having to set up a runner for these Cypress tests...

I am not sure how much "full seed data" may make sense for running product tests (probably we can avoid, but who knows what we may want to test at some point soon), in which case we would not be able to populate ephemeral test instances such as those on GH Actions runners with full data as this requires lots of time - just as a side note.

e2e-product-testing/cypress/integration/Project/on-project-page.ts
```diff
@@ -0,0 +1,7 @@
+import { Given } from "cypress-cucumber-preprocessor/steps";
+
+Given(`I am on Projects page`, () => {
+  cy.visit(`https://marxan.vercel.app/projects`, {
```
hotzevzl, line 4: I suggest to generate these URLs via a URL factory that reads the base URL from an env var and appends paths according to which part of the app we are testing (or at least allows to set the base URL via env var).
Can be in a separate PR, here we're dealing with the initial PoC so we can merge with hardcoded URLs if we wanted.

e2e-product-testing/cypress/integration/Project/on-project-page.ts
```diff
@@ -0,0 +1,7 @@
...
+    timeout: 2000,
```
hotzevzl: I suggest to use some constant (enum?) for these timeouts, logically labeled in a way that makes it easier to choose which timeout to use - mostly to avoid having to take decisions on the exact timeout in each context.

e.g. something like

```
export enum WaitingTimesForResponses {
  subsecond = 1000,
  short = 2000,
  medium = 10000,
  long = 20000,
  giveMeABreak = 60000
};
```

hotzevzl (approved):
thanks @kgajowy I like this - I would wait for feedback from the other devs I mentioned in the discussion about gherkin, and if there are no strong objections (or no responses), I suggest to merge.

alexeh (approved):

e2e-product-testing/cypress/integration/common/pages.ts
```diff
@@ -0,0 +1,9 @@
...
+    timeout: 10000,
+  });
+  // Should be ensure given step was executed?
```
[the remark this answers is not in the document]
kgajowy: I agree about too much verbosity and more assertions to write - lets avoid it for now and see if we will have troubles with finding out what failed

e2e-product-testing/cypress/integration/Project/on-project-page.ts
kgajowy: I agree on separate PR - as per requirements we want the minimal setup to work with ;)

e2e-product-testing/cypress/integration/common/pages.ts
kgajowy: I am not sure about any strong value of testing "local" setup with product test - we want to test the real infrastructure (mostly) - with all the downsizes and exactly how user will experience it. (thus long timeouts are a nice indicator that page works slowly from UX perspective).

We can have them as an addition, of course.

> GH Actions runners with full data as this requires lots of time
we can prepare ready docker images with all the data already seeded and strong cache ; )

PR #199, opened 2021-05-19
Feature/api/marxan 196 migrate for bbox

aagm:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/src/migrations/api/1621439031072-AddBboxToProjects.ts
+++ b/api/src/migrations/api/1621439031072-AddBboxToProjects.ts
@@ -1,0 +1,48 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddBboxToProjects1621439031072
+  implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+    -----------------------------------------
+    -- Generates the new column we need
+    -----------------------------------------
+    ALTER TABLE projects ADD COLUMN bbox jsonb;
+
+    UPDATE projects SET bbox = jsonb_build_array(ST_XMax(extent), ST_XMin(extent), ST_YMax(extent), ST_YMin(extent));
+
+    -----------------------------------------
+    -- tr_GetBbox()
+    -- Utility func to populate bbox
+    -----------------------------------------
+    CREATE OR REPLACE FUNCTION tr_GetBbox()
+    RETURNS trigger AS $BODY$
+    BEGIN
+      IF NEW.extent IS NOT NULL THEN
+        NEW.bbox := jsonb_build_array(ST_XMax(NEW.extent), ST_XMin(NEW.extent), ST_YMax(NEW.extent), ST_YMin(NEW.extent));
+      ELSE
+        NEW.bbox := NULL;
+      END IF;
+      RETURN NEW;
+    END;
+    $BODY$ LANGUAGE plpgsql;
+    -----------------------------------------
+    -- Creates the trigger
+    -----------------------------------------
+    DROP TRIGGER IF EXISTS tr_projects_extent ON projects;
+
+    CREATE TRIGGER tr_projects_extent
+    BEFORE INSERT or UPDATE ON projects
+    FOR EACH ROW EXECUTE
+    PROCEDURE tr_GetBbox();
+    `);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+     await queryRunner.query(`
+     DROP TRIGGER IF EXISTS tr_projects_extent ON projects;
+     ALTER TABLE projects
+     DROP COLUMN bbox;
+     `);
+  }
+}
--- a/api/src/migrations/api/1621439031072-AddBboxToProjects.ts
+++ b/api/src/migrations/api/1621439031072-AddBboxToProjects.ts
@@ -1,7 +1,6 @@
 import { MigrationInterface, QueryRunner } from 'typeorm';
 
-export class AddBboxToProjects1621439031072
-  implements MigrationInterface {
+export class AddBboxToProjects1621439031072 implements MigrationInterface {
   public async up(queryRunner: QueryRunner): Promise<void> {
     await queryRunner.query(`
     -----------------------------------------
@@ -9,7 +8,7 @@ export class AddBboxToProjects1621439031072
     -----------------------------------------
     ALTER TABLE projects ADD COLUMN bbox jsonb;
 
-    UPDATE projects SET bbox = jsonb_build_array(ST_XMax(extent), ST_XMin(extent), ST_YMax(extent), ST_YMin(extent));
+    UPDATE projects SET bbox = jsonb_build_array(ST_XMax(extent), ST_XMin(extent), ST_YMax(extent), ST_YMin(extent)) where extent is not null;
 
     -----------------------------------------
     -- tr_GetBbox()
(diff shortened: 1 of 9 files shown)


api/src/modules/countries/country.geo.entity.ts
```diff
@@ -37,6 +37,14 @@ export class Country {
   @ApiProperty()
   @Column('geometry', { name: 'the_geom' })
   theGeom: any;
+
+   /**
+   * Bbox.
+   */
+    @ApiProperty()
+    @Column('character varying', { name: 'bbox' })
+    bbox!: number[];
```
kgajowy, line 46: Most likely `Entity()` should be replaced with [ViewEntity](https://orkhan.gitbook.io/typeorm/docs/view-entities)

api/src/modules/projects/project.api.entity.ts
```diff
@@ -115,12 +115,19 @@ export class Project extends TimeUserEntityMetadata {
...
   @ApiPropertyOptional()
   @Column('geometry')
   extent?: Record<string, unknown> | null;
 
+  /**
+   * Bbox of the custom extent
+   */
+  @ApiProperty()
+  @Column('character varying', { name: 'bbox' })
+  bbox!: number[];
```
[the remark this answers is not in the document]
aagm: nice catch, is a jsonb column; what i am a bit surprise is taht why it didn't throw an error?

geoprocessing/src/migrations/geoprocessing/1621431520169-AddBboxToAdminAreasAndProjects.ts
```diff
@@ -0,0 +1,78 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+export class AddBboxToAdminAreasAndProjects1621431520169
+  implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+    -----------------------------------------
+    -- Generates the new column we need
+    -----------------------------------------
+    ALTER TABLE admin_regions ADD COLUMN bbox jsonb;
```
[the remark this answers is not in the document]
aagm: That one is entending from the country View, so only defining it in the country should be more than  enough

api/src/modules/countries/country.geo.entity.ts
kgajowy: No, it shouldn't do any troubles (as well as marking as @ViewColumn) - but writing explicitly will give developers some clue that it is a view and most likely one shouldn't insert into this table directly.

api/src/modules/projects/project.api.entity.ts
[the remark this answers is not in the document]
kgajowy: @aagm there is no direct connection between database column definition (from code), actual db definition and code typings in terms of type safety.
While `character varying` will be just ignored (unless you will want to generate migration from code), the `number / jsonb` MAY work assuming that its a `number[]` array inside - so thats just the matter of us not putting any inserts from SQL that arent `number[]` and not putting an invalid `bbox` from code (like: `[0]` or `[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0....]`)

geoprocessing/src/migrations/geoprocessing/1621431520169-AddBboxToAdminAreasAndProjects.ts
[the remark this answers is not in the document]
kgajowy: @aagm you don't have to expose it to frontend (i.e. don't put Api decorators) but I believe we should keep the code definitions really close (the same) as they are in the database - it will help with both generating migrations automatically AND keep the usage clean (right now, if you would like to use bbox from within the code - you can't; and in few months one will need to figure out that the column is there by reviewing every single migration where this table is present)

api/src/migrations/api/1621439031072-AddBboxToProjects.ts
```diff
@@ -0,0 +1,48 @@
...
+export class AddBboxToProjects1621439031072
+  implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+    -----------------------------------------
+    -- Generates the new column we need
+    -----------------------------------------
+    ALTER TABLE projects ADD COLUMN bbox jsonb;
+
+    UPDATE projects SET bbox = jsonb_build_array(ST_XMax(extent), ST_XMin(extent), ST_YMax(extent), ST_YMin(extent));
```
hotzevzl, line 12: as we don't enforce `extent` to be `not null`, are we ok with having a bbox such as `[null, null, null, null]`?

hotzevzl: I see there's logic about this in the trigger function - do we want to apply it here too?

api/src/migrations/api/1621439031072-AddBboxToProjects.ts
```diff
@@ -0,0 +1,48 @@
...
+    RETURNS trigger AS $BODY$
+    BEGIN
+      IF NEW.extent IS NOT NULL THEN
+        NEW.bbox := jsonb_build_array(ST_XMax(NEW.extent), ST_XMin(NEW.extent), ST_YMax(NEW.extent), ST_YMin(NEW.extent));
+      ELSE
+        NEW.bbox := NULL;
+      END IF;
+      RETURN NEW;
+    END;
+    $BODY$ LANGUAGE plpgsql;
```
hotzevzl, line 28: what a shame that a generated column will not work for this 😢

api/src/migrations/api/1621439031072-AddBboxToProjects.ts
```diff
@@ -0,0 +1,48 @@
...
+    -----------------------------------------
+    -- Creates the trigger
+    -----------------------------------------
+    DROP TRIGGER IF EXISTS tr_projects_extent ON projects;
+
+    CREATE TRIGGER tr_projects_extent
+    BEFORE INSERT or UPDATE ON projects
+    FOR EACH ROW EXECUTE
+    PROCEDURE tr_GetBbox();
```
hotzevzl, line 37: happy to go with your preference here - my own is to avoid type-prefixes on db entities (vw, tr, etc.)

api/src/modules/projects/project.api.entity.ts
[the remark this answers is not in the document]
hotzevzl: time to switch from TypeORM to Prisma? (not really... for this project at least 😅)

hotzevzl (approved):

api/src/migrations/api/1621439031072-AddBboxToProjects.ts
aagm: nice catch

api/src/migrations/api/1621439031072-AddBboxToProjects.ts
aagm: My only exception are triggers, it helps me  locate easily those.

alexeh (approved):

api/src/modules/countries/country.geo.entity.ts
aagm: Done

api/src/modules/projects/project.api.entity.ts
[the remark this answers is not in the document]
aagm: done

PR #203, opened 2021-05-21
MARXAN-416 - Project's specific WDPA from shapefile

kgajowy:
# Users should be able to upload shapefiles of project-specific protected areas

This PR handles updating database in `wdpa` table with new geometries.
Old geometries related to given `projectId` are removed in transaction.
Supports FeatureCollection/Multipolyon/Polygon.
Includes two integration (e2e) tests of the whole worker processor (one case enforces breaking db's constraint check, leading to rollback)


geoprocessing/src/modules/protected-areas/worker/protected-area-processor.ts
```diff
@@ -0,0 +1,59 @@
...
+
+      await queryRunner.commitTransaction();
+    } catch (error) {
+      await queryRunner.rollbackTransaction();
+      throw error;
+    } finally {
+      await queryRunner.release();
+    }
+  }
+}
```
Dyostiq, line 59: Have you tried to run it like this?

```suggestion
import { Injectable } from '@nestjs/common';
import { EntityManager, InjectEntityManager } from 'typeorm';
import { Job } from 'bullmq';

import { WorkerProcessor } from '../../worker';
import { ProtectedAreasJobInput } from './worker-input';
import { ShapefileService } from '../../shapefiles/shapefiles.service';
import { ProtectedArea } from '../protected-areas.geo.entity';
import { GeoJSON } from 'geojson';
import { GeometryExtractor } from './geometry-extractor';

@Injectable()
export class ProtectedAreaProcessor
  implements WorkerProcessor<ProtectedAreasJobInput, void> {
  constructor(
    private readonly shapefileService: ShapefileService,
    private readonly geometryExtractor: GeometryExtractor,
    @InjectEntityManager()
    private readonly entityManager: EntityManager,
  ) {}

  async process(job: Job<ProtectedAreasJobInput, void>): Promise<void> {
    const geo: GeoJSON = (await this.shapefileService.getGeoJson(job.data.file))
      .data;

    const geometries = this.geometryExtractor.extract(geo);

    if (geometries.length === 0) {
      throw new Error(
        'No supported geometries found. Ensure that file contains FeatureCollection / Multipolygon / Polygon.',
      );
    }
    
    await this.entityManager.transaction(async (transactionEntityManager) => {
      await transactionEntityManager.delete(ProtectedArea, {
        projectId: job.data.projectId,
      });
      await transactionEntityManager.insert(
        ProtectedArea,
        geometries.map((geometry) => ({
          projectId: job.data.projectId,
          fullName: job.data.file.filename,
          theGeom: () =>
            `st_multi(ST_GeomFromGeoJSON('${JSON.stringify(geometry)}'))`,
        })),
      );
    });
}
```

geoprocessing/src/modules/protected-areas/worker/protected-area-worker.service.ts
```diff
@@ -0,0 +1,23 @@
...
+    private readonly wrapper: WorkerBuilder,
+    private readonly processor: ProtectedAreaProcessor,
+  ) {
+    this.#worker = wrapper.build(queueName, processor);
+    this.#worker.on('completed', ({ returnvalue }) => {
+      console.log(`---- trigger ApiEvent - ok `, returnvalue);
+    });
+    this.#worker.on('failed', ({ failedReason }) => {
+      console.log(`---- trigger ApiEvent - failure `, failedReason);
+    });
```
Dyostiq, line 21: Hmm, is there a reason to use `console.log` instead of `Logger`, or is it a leftover?

Dyostiq (approved):

geoprocessing/src/modules/protected-areas/worker/protected-area-worker.service.ts
kgajowy: @Dyostiq its a placeholder for ApiEvents for next PR

geoprocessing/src/modules/protected-areas/protected-areas.module.ts
```diff
@@ -5,11 +5,16 @@ import { ProtectedAreasController } from './protected-areas.controller';
...
-  imports: [TypeOrmModule.forFeature([ProtectedArea]), TileModule],
+  imports: [
+    TypeOrmModule.forFeature([ProtectedArea]),
+    TileModule,
+    ProtectedAreaWorkerModule,
+  ],
   providers: [ProtectedAreasService],
   controllers: [ProtectedAreasController],
-  exports: [ProtectedAreasService],
+  exports: [ProtectedAreasService, TypeOrmModule],
```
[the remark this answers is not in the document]
kgajowy: @Dyostiq correct me if I am wrong, but otherwise, while using `ProtectedAreasService` , in target module, TypeOrm raised that entity is missing (i.e. would require to manually load .forFeature)

geoprocessing/src/modules/protected-areas/worker/__mocks__/geometry-extractor.ts
```diff
@@ -0,0 +1,11 @@
...
+import { GeoJSON, MultiPolygon, Polygon } from 'geojson';
+
+@Injectable()
+export class FakeGeometryExtractor {
+  extractMock: jest.Mock<(MultiPolygon | Polygon)[]> = jest.fn();
+
+  extract(geo: GeoJSON): (MultiPolygon | Polygon)[] {
+    return this.extractMock(geo);
+  }
+}
```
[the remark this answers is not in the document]
kgajowy: not doable due to private member 🤔

geoprocessing/src/modules/protected-areas/worker/__mocks__/connection.ts
```diff
@@ -0,0 +1,24 @@
...
+    connect: async () => this.connectMock(),
+    startTransaction: async () => this.startTransactionMock(),
+    commitTransaction: async () => this.commitMock(),
+    rollbackTransaction: async () => this.rollbackMock(),
+    release: async () => this.releaseMock(),
+    manager: {
+      delete: this.deleteMock,
+      insert: this.insertMock,
+    },
+  });
```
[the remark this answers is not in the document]
kgajowy: fixed in https://github.com/Vizzuality/marxan-cloud/pull/203/commits/7ddb4582b113d4ae592c145e23d27cd0be000661

geoprocessing/src/modules/protected-areas/worker/geometry-extractor.ts
```diff
@@ -0,0 +1,24 @@
...
+  extract(geo: GeoJSON): (MultiPolygon | Polygon)[] {
+    let geometries: (MultiPolygon | Polygon)[] = [];
+
+    if (geo.type === 'FeatureCollection') {
+      geometries = geo.features
+        .map((feature) => feature.geometry)
+        .filter(this.#isGeometrySupported);
+    }
+
+    if (geo.type === 'MultiPolygon' || geo.type === 'Polygon') {
```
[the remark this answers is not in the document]
kgajowy: required a bit more - fixed in https://github.com/Vizzuality/marxan-cloud/pull/203/commits/a7caf01f0a8c476be3e5c259eb2005feebbb6f58

geoprocessing/src/modules/protected-areas/worker/protected-area-worker.service.ts
Dyostiq: Ok, I didn't know :)

geoprocessing/src/modules/protected-areas/worker/__mocks__/geometry-extractor.ts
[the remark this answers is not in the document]
Dyostiq: then something like
```ts
type MockedInterface<T> = jest.Mocked<
  {
    [index in keyof T]: T[index];
  }
>;

const mocked = <T, K extends keyof T>(): MockedInterface<T>[K] => {
  return jest.fn() as any;
};

@Injectable()
export class FakeGeometryExtractor
  implements MockedInterface<GeometryExtractor> {
  extract = mocked<GeometryExtractor, 'extract'>();
}
```
would do the trick, but it bloats it a bit, so up to you

geoprocessing/src/modules/protected-areas/protected-areas.module.ts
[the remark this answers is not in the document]
Dyostiq: No, it shouldn't, because the provider is in this module, not in modules that imports this module and use the service.

geoprocessing/src/modules/protected-areas/worker/__mocks__/geometry-extractor.ts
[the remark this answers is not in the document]
kgajowy: Thanks! I believe current form is fair enough for test dummies, so I would vote for leaving it as is.

geoprocessing/test/integration/protected-areas/steps/invalid-shapefile.ts
```diff
@@ -0,0 +1,27 @@
+import { readFileSync } from 'fs';
+import { Readable } from 'stream';
+
+export const shapePath = __dirname + `/new-shape-name.zip`;
```
hotzevzl, line 4: I am not sure we can easily make this setup work for tests running under Docker - I suggest to first copy the files to `/tmp`.

The issue is that we bind-mount `src` and `test` inside the container, and this ends up mounting them as a user other than the one that owns the `node` process for the NestJS service.

I have seen suggestions to *`mkdir` and `chown`* directories that are then used for bind mounts, but in practice we already do this via the `COPY --chown ...` commands in the `Dockerfile`s, and in any case none of this works at least with the Docker engine version I am using locally nor the one we have in the GH Actions runner.

The `config` directory gets copied over with correct ownership, and so should do `src` and `test`, but once we bind mount volumes, the owner gets overridden with the owner of the source directory of the volume in the host filesystem.

geoprocessing/src/modules/protected-areas/worker/__mocks__/shapefile-service.ts
```diff
@@ -0,0 +1,10 @@
+import { Injectable } from '@nestjs/common';
```
hotzevzl, line 1: worth adding this mock under the directory of the existing `ShapefilesModule` where we have the actual `ShapefilesService`?

geoprocessing/test/integration/protected-areas/steps/invalid-shapefile.ts
kgajowy:
```
[Nest] 133   - 05/25/2021, 1:43:59 PM   [ExceptionHandler] Error: new-shape-name could not be extracted: Error: Error: EACCES: permission denied, mkdir '/opt/marxan-geoprocessing/test/integration/protected-areas/steps/new-shape-name' +130ms
``` 
indeed, I didn't catch it 🤔 

I don't feel fine to modify the tests just due to different environment (would be those test reliable?)
If we are going to use shared volume, those will change anyway "a bit", so I struggle to figure out how to make move forwards without going back and forth 🤔 

What do you propose @hotzevzl ?

kgajowy: (just to mention: running those specs locally via docker-compose works nice 🤔 )

geoprocessing/src/modules/protected-areas/worker/__mocks__/connection.ts
```diff
@@ -0,0 +1,24 @@
+import { Injectable } from '@nestjs/common';
```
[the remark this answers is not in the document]
kgajowy: I am happy with that once we hit monorepo!

geoprocessing/src/modules/protected-areas/worker/protected-area-processor.ts
```diff
@@ -0,0 +1,59 @@
...
+        'No supported geometries found. Ensure that file contains FeatureCollection / Multipolygon / Polygon.',
+      );
+    }
+
+    const queryRunner = this.connection.createQueryRunner();
+
+    await queryRunner.connect();
+    await queryRunner.startTransaction();
+    try {
+      await queryRunner.manager.delete(ProtectedArea, {
```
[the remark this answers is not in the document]
kgajowy: It will be validated on controller/application level I believe (not a worker itself)

geoprocessing/src/modules/protected-areas/worker/protected-area-processor.ts
```diff
@@ -0,0 +1,59 @@
...
+        projectId: job.data.projectId,
+      });
+      await queryRunner.manager.insert(
+        ProtectedArea,
+        geometries.map((geometry) => ({
+          projectId: job.data.projectId,
+          fullName: job.data.file.filename,
+          theGeom: () =>
+            `st_multi(ST_GeomFromGeoJSON('${JSON.stringify(geometry)}'))`,
```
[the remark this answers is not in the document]
kgajowy: @hotzevzl what exactly do you have on mind? We cannot pass pure `geometry` as it won't be accepted by SQL (`[object Object]`)

geoprocessing/src/modules/protected-areas/worker/__mocks__/shapefile-service.ts
kgajowy: it it would be widely used, I would rather think to put it under `utils/__mocks__`. 
However, every module could need different behavior/mock. Not sure if we want to put all mocks around into bucket getting larger and larger 🤔

geoprocessing/test/integration/protected-areas/steps/invalid-shapefile.ts
```diff
@@ -0,0 +1,27 @@
+import { readFileSync } from 'fs';
```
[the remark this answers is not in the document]
kgajowy: fixed in https://github.com/Vizzuality/marxan-cloud/pull/203/commits/868bde88c8f340356783588b5cc69d9c8b5f8d81

geoprocessing/src/modules/protected-areas/worker/__mocks__/connection.ts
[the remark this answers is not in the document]
kgajowy: moved to utils in https://github.com/Vizzuality/marxan-cloud/pull/203/commits/a4d224abc76db2fac79f38c0279ae848dd4eed67

geoprocessing/src/modules/protected-areas/worker/__mocks__/shapefile-service.ts
kgajowy: moved to utils in https://github.com/Vizzuality/marxan-cloud/pull/203/commits/a4d224abc76db2fac79f38c0279ae848dd4eed67

geoprocessing/test/integration/protected-areas/steps/invalid-shapefile.ts
kgajowy: hm, it actually fails not on reading a file but while trying to unpack it

kgajowy: image
as I mentioned a few times, we should probably unpack it to %TMP folder instead of the very same location

kgajowy: I am out of ideas for now. @hotzevzl should we mark those tests as `skipped` and move forwards until we manage to figure out shared volume?

hotzevzl: @kgajowy (cc @alexeh) yes, I think we should unpack within a temporary folder, letting the unpacker code (or something close to it) take care of this, so that the callers don't need to know about details of the filesystem (at least some of them... we need to draw a line somewhere for MVP).

For now let's `.skip`, but we need to add `tmp` handling to the unpacker or so in order to enable these tests - shared volume won't by itself help with this - we'd still need to unpack the contents of zip files outside of the `source`/`test` subtrees.

BTW in practice these tests *could* work as they are *iff* the owner of the `test` folder *on the host machine* is set to 5000, but we cannot rely on this.

PR #208, opened 2021-05-24
refactor(api): transform project to monorepo

Dyostiq:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


Dyostiq:
I'll omit the test commit during the rebase

hotzevzl (approved):
@Dyostiq thank you! almost everything that I could check so far seems ok:

- build from clean-slate
- migrations
- seed-dbs
- unit tests
- e2e tests

Given that such a far-reaching reorganization can be hard to keep up to date for long as we keep merging other PRs, I would be keen to merge this sooner rather than later if nobody has strong concerns.

Given that "stuff seems to work", and especially that tests keep passing, I think that it should be easier to deal with any snags as they emerge, if at all, rather than trying to unearth any snags before merging.

kgajowy (approved):

hotzevzl (approved):
@Dyostiq thanks, this is awesome 🚀 

All works ok from my point of view, I'd be keen to merge after confirming with the team tomorrow morning first thing that everyone is ok with rebasing their open PRs and adjusting paths 😱 

We may want to revisit the Dockerfiles in a second moment to avoid copying into the container images source we may not need, but we'd need to do some work there in any case for production images, so I would not worry about this now.

Dyostiq:
@hotzevzl 
regarding copying not-needed files to the container – I believe we should copy there only an artifact of built code, and tsc/webpack compiles the required dependency graph, but it requires a Dockerfile rewrite (and maybe some fixes around workers). Now, with ts files in the container and with shared code possibility, we can't be sure what code is really needed.

hotzevzl:
> @hotzevzl
> regarding copying not-needed files to the container – I believe we should copy there only an artifact of built code, and tsc/webpack compiles the required dependency graph, but it requires a Dockerfile rewrite (and maybe some fixes around workers). Now, with ts files in the container and with shared code possibility, we can't be sure what code is really needed.

yep, that's the plan as we move towards proper staging and eventually prod deployments - I'll likely do a multi-stage Docker image build

PR #209, opened 2021-05-25
feature MARXAN-399 validate uploaded shapefile

alexeh:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -11,23 +12,40 @@ export class ShapefileService {
...
+        '',
+      )}/*.shp -clean rewind -info -o ${outputKey}`,
     );
 
     return JSON.parse(_geoJson[outputKey].toString('utf-8'));
   }
 
+  isValidGeoJson(geoJson: any): Error | void {
+    if (
+      geoJson.type !== 'FeatureCollection' ||
```
kgajowy, line 26: have you tried `pure-geojson-validation`?

alexeh: I didn't, I'll have a look thanks @kgajowy !

api/src/pipes/zipfile-validation.pipe.ts
```diff
@@ -0,0 +1,9 @@
+import { ArgumentMetadata, PipeTransform } from '@nestjs/common';
+
+export class ZipfileValidationPipe implements PipeTransform {
+  transform(value: Express.Multer.File, metadata: ArgumentMetadata): any {
+    console.log('value', value);
+    console.log('metadata', metadata);
```
Dyostiq, line 6: I think these `console.log` are not necessary. If they are, please use `Logger#debug`

geoprocessing/src/modules/files/files.service.ts
```diff
@@ -2,32 +2,41 @@ import { Injectable } from '@nestjs/common';
...
     return new Promise((resolve, reject) => {
-      createReadStream(fileInfo.path)
+      createReadStream(origin)
         .pipe(
           Extract({
             path: path.join(
-              fileInfo.destination,
-              path.basename(fileInfo.filename.replace('.zip', '')),
+              destination,
+              path.basename(fileName.replace('.zip', '')),
```
Dyostiq, line 20: Hmm, actually
```suggestion
              path.basename(fileName, '.zip'),
```
will do the same

geoprocessing/src/modules/files/files.service.ts
```diff
@@ -2,32 +2,41 @@ import { Injectable } from '@nestjs/common';
...
   }
 
+  async areFilesInFolder(
+    path: string,
+    fileExtensions: Array<string>,
+  ): Promise<boolean> {
+    const filesInPath = await readdir(path);
+    const extensions = filesInPath.map((file) => file.split('.').pop());
+    return fileExtensions.every((ext: string) => extensions.includes(ext));
+  }
```
Dyostiq, line 38: the method lacks cohesion to the class

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -1,33 +1,68 @@
...
   private async shapeFileToGeoJson(fileInfo: Express.Multer.File) {
+    if (
+      !(await this.fileService.areFilesInFolder(
+        fileInfo.path.replace('.zip', ''),
+        MIN_REQUIRED_FILES,
+      ))
+    ) {
+      throw new Error('Some required file mising');
+    }
+    console.log('works');
```
Dyostiq, line 22: looks like a leftover

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -1,33 +1,68 @@
...
     const outputKey = `shapefile-${new Date().getTime()}.geojson`;
 
     const _geoJson = await mapshaper.applyCommands(
-      `-i ${fileInfo.path.replace('.zip', '')}/*.shp -info -o ${outputKey}`,
+      `-i no-topology snap ${fileInfo.path.replace(
+        '.zip',
+        '',
+      )}/*.shp -proj EPSG:4326 -clean rewind -info -o ${outputKey}`,
```
Dyostiq, line 29: how would map-shaper handle paths with spaces in them? Maybe we should wrap them with `"`

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -1,33 +1,68 @@
...
+          shapeFile.filename,
+          shapeFile.destination,
+        ),
+      );
+      const geoJson = await this.shapeFileToGeoJson(shapeFile);
+      this.isValidGeoJson(geoJson);
+      return { data: geoJson };
     } catch (err) {
       this.logger.error(err);
+      throw new BadRequestException('Invalid Shapefile');
```
Dyostiq, line 61: leaks a presentation a little bit, the user of the method doesn't have to be a controller
but maybe it's ok, as it accepts `Express.Multer.File`

Dyostiq (approved):

api/src/pipes/zipfile-validation.pipe.ts
alexeh: @Dyostiq I'm not done with the API side validation part so this is WIP
Actually I think I'll need your help here :)

geoprocessing/src/modules/files/files.service.ts
alexeh: @Dyostiq could you explain it a little bit please?

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
alexeh: @kgajowy I had a look and I realised  [you use it in the API,](you use it in the) but then you 'vanilla check' same things as mine, so I'm not sure which is the point to use it if `mapshapper` does not complain at generating the GeoJSON

geoprocessing/src/modules/files/files.service.ts
alexeh: Thanks @Dyostiq 3891b195524fcb5a64e83d62f0ca8b010c42dbba

geoprocessing/src/modules/files/files.service.ts
alexeh: Thoughts and opinions always appreciated, thanks @hotzevzl a0793bab77839c74ad79ca3231d99a45f4078167

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -1,33 +1,68 @@
...
   constructor(private readonly fileService: FileService) {}
 
   private async shapeFileToGeoJson(fileInfo: Express.Multer.File) {
+    if (
+      !(await this.fileService.areFilesInFolder(
+        fileInfo.path.replace('.zip', ''),
+        MIN_REQUIRED_FILES,
+      ))
+    ) {
+      throw new Error('Some required file mising');
```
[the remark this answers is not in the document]
alexeh: For sure, much better! a0793bab77839c74ad79ca3231d99a45f4078167

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
alexeh: Nice catch! I've actually had some issues there, I decided to delete whitespaces in MulterOptions
Changes here 61dd5b268ff1f26718f885c3b84a22e9d8354bd0

kgajowy (changes requested):
Please ensure to rebase to current develop. 

PS as FileService is changing contract in non-backward compatible way, it may be required to  verify other features.

PS2 similar to https://github.com/Vizzuality/marxan-cloud/pull/203 it may have some issues in docker env.

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -1,33 +1,68 @@
...
+      `-i no-topology snap ${fileInfo.path.replace(
+        '.zip',
+        '',
+      )}/*.shp -proj EPSG:4326 -clean rewind -info -o ${outputKey}`,
     );
 
     return JSON.parse(_geoJson[outputKey].toString('utf-8'));
   }
 
+  isValidGeoJson(geoJson: FeatureCollection): Error | void {
```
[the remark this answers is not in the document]
alexeh: Thanks @hotzevzl . It seemed nasty to explicitly throw an error inside a try/catch block for me at first.
591e7097cd28b32cb3cbe8effd61644ad56dd28b

PS: Now wondering if we should add any error that may occur here to `BadRequestException` message...

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -1,33 +1,68 @@
...
-      this.logger.log(await this.fileService.unzipFile(shapeFile));
+      this.logger.log(
+        await this.fileService.unzipFile(
+          shapeFile.path,
+          shapeFile.filename,
+          shapeFile.destination,
+        ),
+      );
+      const geoJson = await this.shapeFileToGeoJson(shapeFile);
+      this.isValidGeoJson(geoJson);
```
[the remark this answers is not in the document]
alexeh: https://github.com/Vizzuality/marxan-cloud/pull/209#discussion_r639518364

591e7097cd28b32cb3cbe8effd61644ad56dd28b

geoprocessing/src/modules/shapefiles/shapefiles.service.ts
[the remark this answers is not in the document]
hotzevzl: up to you - I'd say where it is easy and meaningful to send useful error details we should do it, especially if the situation may be due to user error (which could be common, for example an oversight) rather than application error (such as an bug in the way a specific software/library assembles a shapefile), as the user may not know what to do in any case in the latter case.

api/apps/api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -146,20 +148,24 @@ export class ScenariosController {
...
-      response,
-    );
-    return proxyServiceResponse;
+    const { data } = await this.httpService
+      .post(
+        `${this.geoprocessingUrl}${apiGlobalPrefixes.v1}/planning-units/planning-unit-shapefile`,
+        file,
+        {
+          headers: { 'Content-Type': 'application/json' },
+          validateStatus: (status) => status <= 499,
```
hotzevzl, line 163: Are we really fine with the 4xx range? While flicking through the code I do find it surprising that we resolve the promise for 4xx responses from downstream - maybe a quick tiny comment to explain why could help here.

api/apps/geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -1,33 +1,80 @@
...
 
 @Injectable()
 export class ShapefileService {
   private readonly logger: Logger = new Logger(ShapefileService.name);
+  private readonly minRequiredFiles = ['prj', 'dbf', 'shx', 'shp'];
   constructor(private readonly fileService: FileService) {}
 
   private async shapeFileToGeoJson(fileInfo: Express.Multer.File) {
+    if (
+      !(await this.areRequireShapefileFilesInFolder(
```
hotzevzl, line 16:
```suggestion
      !(await this.areRequiredShapefileFilesInFolder(
```

api/apps/geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -1,33 +1,80 @@
...
+      `-i no-topology snap ${fileInfo.path.replace(
+        '.zip',
+        '',
+      )}/*.shp -proj EPSG:4326 -clean rewind -info -o ${outputKey}`,
     );
 
     return JSON.parse(_geoJson[outputKey].toString('utf-8'));
   }
 
+  async areRequireShapefileFilesInFolder(path: string): Promise<boolean> {
```
hotzevzl, line 36:
```suggestion
  async areRequiredShapefileFilesInFolder(path: string): Promise<boolean> {
```

hotzevzl (approved):
@alexeh thanks! Looks good to me, just have a look at the inline comment about accepted responses from upstream and two tiny naming suggestions, and the PR can be merged from my point of view 🚀

api/apps/api/src/modules/scenarios/scenarios.controller.ts
alexeh: @hotzevzl I added this because **axios** (wrapped by HttpService) rejects the promise if status > 399, so if we return a response above that in geoprocessing (because shapefile is not validated), axios wraps that message and the API consumer does not get a 'Invalid Shapefile' response

We could add a trycatch block here but this seemed the best option to me.

hotzevzl: ok, I understand, indeed it makes sense as we're only proxying. thanks!

hotzevzl:
@alexeh thanks for the updates responding to reviews - do you need to do any more work on this or to address Kamil's earlier comments? Otherwise I'd say we could merge. If you need to do further checks about mapshaper handling specific cases, I suggest to do this in a separate PR.

I think we're missing tests for this feature - at least e2e: can you add some in a separate PR, maybe using the test files you've already been using?

alexeh:
@hotzevzl my pleasure. 
Some of this changes (flags added to mapshapper to fix automagically some geometry issues that the SHP might have) caused to break some tests previously implemented. 
There is a `new-shape-name.zip` that has overlapping geometries that was expected to throw a PostGIS error at `'ST_IsValid(the_geom)'` check. Now that overlapping geometries gets fixed by mapshapper and thats why 2 test in the CI for this branch are failing.

I asked Alicia if she can kindly provide one that would actually pass all the validation pipe and still throw an error, so once I get that one I will fix the issue and merge.

Regarding to the test thing. You are totally right, shame on me. I'll add some tests for this feature in a different PR, I need to properly understand the (at least for me) complex underground that has been created there.

Thanks!

api/apps/geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -1,33 +1,80 @@
...
 
+  async areRequiredShapefileFilesInFolder(path: string): Promise<boolean> {
+    const filesInPath = await readdir(path);
+    const extensions = filesInPath.map((file) => file.split('.').pop());
+    return this.minRequiredFiles.every((ext: string) =>
+      extensions.includes(ext),
+    );
+  }
+
+  isValidGeoJson(geoJson: FeatureCollection): boolean {
```
kgajowy, line 44: 1. if you are checking GeoJson, put relevant type instead concrete one. 
2. It does not actually check if GeoJson is valid - it checks if given Geo is a Feature Collection and contains only relevant types (i.e. `isGeoJsonTypeSupported`)

api/apps/geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -1,33 +1,80 @@
...
+    if (
+      geoJson.type !== 'FeatureCollection' ||
+      geoJson.features.every(
+        (geom: any) =>
+          geom.geometry.type !== 'Polygon' &&
+          geom.geometry.type !== 'MultiPolygon',
+      )
+    )
+      return false;
+    return true;
```
kgajowy, line 54:
```
 if (condition) { ... return false } return true
```
->
```
  return !(condition)
```

api/apps/geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -1,33 +1,80 @@
...
+      this.logger.log(
+        await this.fileService.unzipFile(
+          shapeFile.path,
+          shapeFile.filename,
+          shapeFile.destination,
+        ),
+      );
+      const geoJson = await this.shapeFileToGeoJson(shapeFile);
+      if (!this.isValidGeoJson(geoJson)) {
+        throw new Error();
```
kgajowy, line 68: Missing message? Consumer won't know what failed.

api/apps/geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -1,33 +1,80 @@
...
+      }
+      return { data: geoJson };
     } catch (err) {
       this.logger.error(err);
+      throw new BadRequestException('Invalid Shapefile');
```
kgajowy, line 73: The module shouldn't be aware of the REST context.

alexeh: @kgajowy Got it, but if a `Error` is thrown, NestJS returns a status > 4xx causing HttpService on the API to wrap it up so the consumers does not get a 'Invalid Shapefile' msg

api/apps/geoprocessing/src/modules/shapefiles/shapefiles.service.ts
alexeh: @kgajowy We agreed on returning just a Invalid Shapefile message and provide the user with a list of reasons that would make a SHP fail.

But yes, I think I should have added a more precise msg for this error

api/apps/geoprocessing/src/modules/shapefiles/shapefiles.service.ts
```diff
@@ -1,33 +1,80 @@
...
+        '',
+      )}/*.shp -proj EPSG:4326 -clean rewind -info -o ${outputKey}`,
     );
 
     return JSON.parse(_geoJson[outputKey].toString('utf-8'));
   }
 
+  async areRequiredShapefileFilesInFolder(path: string): Promise<boolean> {
+    const filesInPath = await readdir(path);
+    const extensions = filesInPath.map((file) => file.split('.').pop());
```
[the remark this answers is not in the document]
alexeh: @kgajowy  does not work good with files with underscores in it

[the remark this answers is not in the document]
kgajowy:
```
> p.extname('some_fil asd jdaioshd8(#$& _.  .... e.xt')
'.xt'
> 
```
what exactly does not work?

api/apps/geoprocessing/src/modules/shapefiles/shapefiles.service.ts
kgajowy: @alexeh the consumer should catch the error then and map it to relevant error in given transport layer

api/apps/geoprocessing/src/modules/shapefiles/shapefiles.service.ts
alexeh: Changes ad322a92ea39795a68cd1db0d2cdeff3d1dec48d

api/apps/geoprocessing/src/modules/shapefiles/shapefiles.service.ts
alexeh: Changes: ad322a92ea39795a68cd1db0d2cdeff3d1dec48d

api/apps/geoprocessing/src/modules/shapefiles/shapefiles.service.ts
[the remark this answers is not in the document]
alexeh: Yeh it does, sorry

Changes: 5e790328f1c5c31a9dc5a8f67744cf49942741a0

kgajowy (commented):
https://github.com/Vizzuality/marxan-cloud/pull/209#discussion_r643746499 still pending

api/apps/geoprocessing/src/modules/shapefiles/shapefiles.service.ts
kgajowy: @alexeh why did you resolve this without any comment?

alexeh: @kgajowy Didn't notice this, mistake

kgajowy (approved):

PR #210, opened 2021-05-25
configure volume for sharing of uploaded files between api and geoprocessing service (development environments)

hotzevzl:
See https://vizzuality.atlassian.net/browse/MARXAN-440 (and https://vizzuality.atlassian.net/browse/MARXAN-439 for context)


hotzevzl:
@aagm having discussed this PR with Kamil, and given the relatively small blast radius, I am going to merge it now.

PR #211, opened 2021-05-25
chore(geoprocessing): move cost surface processing from api to geo

kgajowy:
This PR creates shell for cost surface update entirely in Geoprocessing service (within job processor).

Moving db adapter - https://github.com/Vizzuality/marxan-cloud/pull/213
Removing leftovers from API & adding queue - ...


hotzevzl (commented):
I got mildly lost through this PR - nothing jumps out as obviously needing attention, but I may need to see this again after merging the related PR to map things out in my mind, and better if not late in the evening 😬

hotzevzl (approved):
@kgajowy thanks, all looks good and nicely structured to me 🚀 

only a tiny naming note - can you do a final check of all the cost surface vs surface cost naming instances in class/function/etc names, please? I think when you use surface cost you are referring to the cost itself rather than the "cost surface", but I would argue that's more of a PU cost - maybe we can drop the "surface" part when referring to the cost as that should be inferrable from the context in any case.

Regardless, the module itself should be `CostSurfaceModule` IMHO. Anyway, these are only details that should hopefully make reading the code easier at first sight.

PR #212, opened 2021-05-25
feat(geoprocessing): protected-areas: geo extractor: emit api events

kgajowy:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


api/src/migrations/api/1621963684141-ProjectSpecificWdpaEvents.ts
```diff
@@ -0,0 +1,18 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class CostSurfaceEvents1621963684141 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+      INSERT INTO api_event_kinds (id) values
+      ('project.wdpa.submitted/v1/alpha'),
+      ('project.wdpa.finished/v1/alpha'),
+      ('project.wdpa.failed/v1/alpha');
```
hotzevzl, line 9: I suggest to use *protected areas* rather than (the more specific) *wdpa* term where relevant - I know our core db table for protected areas is called `wdpa` but in practice it's a table for protected area info, some (most in some cases?) of which happen to be from the WDPA source. We've discussed this with Alicia at some point in the past, though we have kept the `wdpa` table name and some other such references around.

```suggestion
      ('project.protectedAreas.submitted/v1/alpha'),
      ('project.protectedAreas.finished/v1/alpha'),
      ('project.protectedAreas.failed/v1/alpha');
```

api/src/migrations/api/1621963684141-ProjectSpecificWdpaEvents.ts
```diff
@@ -0,0 +1,18 @@
...
+    `);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `DELETE FROM api_event_kinds WHERE id like 'project.wdpa.%';`,
```
hotzevzl, line 15: as above if accepting the previous suggestion
```suggestion
      `DELETE FROM api_event_kinds WHERE id like 'project.protectedAreas.%';`,
```

hotzevzl: and so on for the rest of the PR 😱 - best to cross check with @aagm that she is indeed happy with this suggestion 😃

api/src/modules/api-events/api-event.api.entity.ts
```diff
@@ -32,6 +32,9 @@ export enum API_EVENT_KINDS {
   scenario__costSurface__shapeConversionFailed__v1_alpha1 = 'scenario.costSurface.shapeConversionFailed/v1alpha1',
   scenario__costSurface__costUpdateFailed__v1_alpha1 = 'scenario.costSurface.costUpdateFailed/v1alpha1',
   scenario__costSurface__finished__v1_alpha1 = 'scenario.costSurface.finished/v1alpha1',
+  project__wpda__submitted__v1__alpha = 'project.wdpa.submitted/v1/alpha',
+  project__wpda__submitted__v1__finished = 'project.wdpa.finished/v1/alpha',
+  project__wpda__submitted__v1__failed = 'project.wdpa.failed/v1/alpha',
```
hotzevzl, line 37: please double check the labels of these enum members - did you mean `project__wdpa__finished__v1__alpha` and so on? (or `project__protectedAreas__finished__v1__alpha` if replacing wdpa)

geoprocessing/src/modules/api-events/events.enum.ts
```diff
@@ -10,4 +10,12 @@ export enum API_EVENT_KINDS {
...
   user__passwordResetSucceeded__v1alpha1 = 'user.passwordResetSucceeded/v1alpha1',
   user__passwordResetFailed__v1alpha1 = 'user.passwordResetFailed/v1alpha1',
+  scenario__costSurface__submitted__v1_alpha1 = 'scenario.costSurface.submitted/v1alpha1',
+  scenario__costSurface__shapeConverted__v1_alpha1 = 'scenario.costSurface.shapeConverted/v1alpha1',
+  scenario__costSurface__shapeConversionFailed__v1_alpha1 = 'scenario.costSurface.shapeConversionFailed/v1alpha1',
+  scenario__costSurface__costUpdateFailed__v1_alpha1 = 'scenario.costSurface.costUpdateFailed/v1alpha1',
+  scenario__costSurface__finished__v1_alpha1 = 'scenario.costSurface.finished/v1alpha1',
+  project__wpda__submitted__v1__alpha = 'project.wdpa.submitted/v1/alpha',
+  project__wpda__submitted__v1__finished = 'project.wdpa.finished/v1/alpha',
+  project__wpda__submitted__v1__failed = 'project.wdpa.failed/v1/alpha',
```
hotzevzl, line 20: same as in api service

hotzevzl (commented):
thanks @kgajowy.

Besides the minor naming suggestions, I wonder if we should use the `data` prop of these events to also include in the series of events (submitted/finished or failed) for each request a unique id that links them together, a sort of correlation id. Actually, maybe such correlation id should be a top-level column/prop for API events 🤔...

Probably we can get away without this at this stage - the main issue may be that if users submit two shapefiles for protected areas within a short timeframe and the first one is still processing while the second starts being processed, we may never know of a failure of the second one if the first one takes longer to complete processing and completes successfully (or variations thereof of similar timing issues).

If we feel we should handle these possible race conditions explicitly, we'd still need to handle these correlation ids in a meaningful way, of course.

kgajowy:
> thanks @kgajowy.
> 
> Besides the minor naming suggestions, I wonder if we should use the `data` prop of these events to also include in the series of events (submitted/finished or failed) for each request a unique id that links them together, a sort of correlation id. Actually, maybe such correlation id should be a top-level column/prop for API events 🤔...
> 
> Probably we can get away without this at this stage - the main issue may be that if users submit two shapefiles for protected areas within a short timeframe and the first one is still processing while the second starts being processed, we may never know of a failure of the second one if the first one takes longer to complete processing and completes successfully (or variations thereof of similar timing issues).
> 
> If we feel we should handle these possible race conditions explicitly, we'd still need to handle these correlation ids in a meaningful way, of course.

It sounds like a request tracing. If we would like to, consumer (or we at controller level) can assign some `transaction-id` which is attached to every track-able step (for example, logger in each step) so that we could track full request path.

We can add timestamp / uuid to job as well.

api/src/modules/api-events/api-event.api.entity.ts
kgajowy: 🤦 shame on me! thanks for catching this!

kgajowy:
@hotzevzl fixed the above in https://github.com/Vizzuality/marxan-cloud/pull/212/commits/24066f0be1d4fa0b0ed4953686111edf2e4ad0e0

hotzevzl (approved):

hotzevzl:
> > thanks @kgajowy.
> > Besides the minor naming suggestions, I wonder if we should use the `data` prop of these events to also include in the series of events (submitted/finished or failed) for each request a unique id that links them together, a sort of correlation id. Actually, maybe such correlation id should be a top-level column/prop for API events thinking...
> > Probably we can get away without this at this stage - the main issue may be that if users submit two shapefiles for protected areas within a short timeframe and the first one is still processing while the second starts being processed, we may never know of a failure of the second one if the first one takes longer to complete processing and completes successfully (or variations thereof of similar timing issues).
> > If we feel we should handle these possible race conditions explicitly, we'd still need to handle these correlation ids in a meaningful way, of course.
> 
> It sounds like a request tracing. If we would like to, consumer (or we at controller level) can assign some `transaction-id` which is attached to every track-able step (for example, logger in each step) so that we could track full request path.
> 
> We can add timestamp / uuid to job as well.

I think I overshot by using "correlation id" which indeed is used in request tracing - that's a useful thing to have in general, but here I mean only a way to correlate the ApiEvents related to the same file upload for a specific project. Only something to think about for the reasons/caveats I outlined in my comment, but not for this PR.

Dyostiq:
@kgajowy it is not `chore` :D

PR #215, opened 2021-05-28
feat(api): add service for cost surface template with queue management

Dyostiq:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


kgajowy (approved):

hotzevzl:
@Dyostiq thanks. Interfaces and tests look good.

I am not recognizing much of Kamil's recent work on queues here - I may be misunderstanding the intent, but I see a partial overlap and not much reuse: could we stick to the queue interfaces and workflow that Kamil has prepared (and used in some of the modules he's been working on)? If there are features missing or some interface mismatch with what is needed here, I'd suggest to try to update the existing Queues implementation,

Dyostiq:
@hotzevzl

Hmm, actually the code here is using Kamil's QueueModule. The `BullmqQueue` implements specific needs of `QueuedCostTemplateService` and does not provide a universal abstraction. Kamil's module also does not do this. It keeps the queue management – creating the queue, unregistering it, etc., and I'm injecting the queue using the module's token. It's not very visible now because it's not brought together in the ScenariosModule, yet.

hotzevzl (approved):

hotzevzl:
> @hotzevzl
> 
> Hmm, actually the code here is using Kamil's QueueModule. The `BullmqQueue` implements specific needs of `QueuedCostTemplateService` and does not provide a universal abstraction. Kamil's module also does not do this. It keeps the queue management – creating the queue, unregistering it, etc., and I'm injecting the queue using the module's token. It's not very visible now because it's not brought together in the ScenariosModule, yet.

yes, we need to converge on uniform structure and implementation details (for what is not really unique, of course) for these queue-based modules -- it seems we're still finding our way here, and that's fine at this stage, as long as we do some lightweight refactor once we've stabilised how we use these queues, to avoid having details differing across them, or every time a dev needs to make any changes they'd have to first orient themselves on what exactly to touch in each implementation, which is obviously not efficient.

Dyostiq:
@hotzevzl differences between using queues (and using any other tool) are ok, because a dev may have specific needs (for example, here I use the specific nature of jobId). It's why I esteem the inversion – details are hidden behind an interface that expresses the needs of our class.  I try to avoid generalizing things too early to avoid creating a wrong abstraction. Then, it requires workarounds that bloat the code (and the code is not that expressive). As the rule of three says, with three very similar usages (adapters) we may generalize things.

hotzevzl:
@Dyostiq I think we can merge this, right? feel free to do so if you don't need to apply any further changes

aagm (approved):

PR #216, opened 2021-05-28
refactor: prefix apps

Dyostiq:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


kgajowy (approved):
many love <3

kgajowy:
> thanks @Dyostiq
> 
> I am still keen to keep app-root-relative import paths (so mostly `@marxan-api/modules/etc` or similarly for `@marxan-geoprocessing`) rather than relative paths, with the exception of in-module imports (e.g. `import { WhateverDTO } from './dto/whatever.dto.ts`.
> 
> My preference for this is simple - with app-root-relative paths I can see at a glance which module/part of the tree we are importing something from, without even looking at the path of the current file. And if we move a file around, app-root-relative paths stay the same, whereas file-relative paths may change and show up as (to me) a tiny bit of extra noise in the diff.
> 
> @kgajowy I think you had opinions on this, feel free to raise them 😃

I totally agree. The reason I went on a shortcut with relative to other modules was just jest issues ;)

Dyostiq:
@hotzevzl I also prefer "absolute" paths, especially with index.ts as a `public` interface of the package (to imitate something like `public` vs `internal` in java)

PR #218, opened 2021-05-28
feat(api): projects: deny uploading shapefile for non-existing project

kgajowy:
Diff is messed up. Recommended reading: commit via commit
* https://github.com/Vizzuality/marxan-cloud/pull/218/commits/88845bb6a71acd7530074b88d3ead4977b680541 - moves current project.service to `project-crud` & creates project-root (later renamed) which will contain more suitable logic (like ACL/resource verification)
* https://github.com/Vizzuality/marxan-cloud/pull/218/commits/96494e9b9cc421280b16788eddbb78ac067e86b1 - check projectId

### DIFF
--- a/api/apps/api/src/modules/projects/dto/geo-feature.mapper.ts
+++ b/api/apps/api/src/modules/projects/dto/geo-feature.mapper.ts
@@ -1,0 +1,16 @@
+import { Injectable } from '@nestjs/common';
+import { GeoFeaturesService } from '../../geo-features/geo-features.service';
+import { PaginationMeta } from '../../../utils/app-base.service';
+import { GeoFeature } from '../../geo-features/geo-feature.api.entity';
+
+@Injectable()
+export class GeoFeatureMapper {
+  constructor(private readonly geoFeaturesService: GeoFeaturesService) {}
+
+  async serialize(
+    entities: Partial<GeoFeature> | (Partial<GeoFeature> | undefined)[],
+    paginationMeta?: PaginationMeta,
+  ): Promise<any> {
+    return this.geoFeaturesService.serialize(entities, paginationMeta);
+  }
+}
--- a/api/apps/api/src/modules/projects/projects-crud.ts
+++ b/api/apps/api/src/modules/projects/projects-crud.ts
@@ -5,8 +5,6 @@ import { Repository, SelectQueryBuilder } from 'typeorm';
 import { Project } from './project.api.entity';
 import { CreateProjectDTO } from './dto/create.project.dto';
 import { UpdateProjectDTO } from './dto/update.project.dto';
-
-import * as faker from 'faker';
 import { UsersService } from '@marxan-api/modules/users/users.service';
 import { ScenariosService } from '@marxan-api/modules/scenarios/scenarios.service';
 import { PlanningUnitsService } from '@marxan-api/modules/planning-units/planning-units.service';
@@ -34,7 +32,7 @@ type ProjectFilterKeys = keyof Pick<
 type ProjectFilters = Record<ProjectFilterKeys, string[]>;
 
 @Injectable()
-export class ProjectsService extends AppBaseService<
+export class ProjectsCrud extends AppBaseService<
   Project,
   CreateProjectDTO,
   UpdateProjectDTO,
@@ -100,32 +98,6 @@ export class ProjectsService extends AppBaseService<
         ],
       },
     };
-  }
-
-  async importLegacyProject(_file: Express.Multer.File): Promise<Project> {
-    return new Project();
-  }
-
-  async fakeFindOne(_id: string): Promise<Project> {
-    const project = {
-      ...new Project(),
-      id: faker.random.uuid(),
-      name: faker.lorem.words(5),
-      description: faker.lorem.sentence(),
-      users: await Promise.all(
-        Array.from({ length: 10 }).map(
-          async (_userId) =>
-            await this.usersService.fakeFindOne(faker.random.uuid()),
-        ),
-      ),
-      attributes: await Promise.all(
-        Array.from({ length: 5 }).map(
-          async (_scenarioId) =>
-            await this.scenariosService.fakeFindOne(faker.random.uuid()),
-        ),
-      ),
-    };
-    return project;
   }
 
   /**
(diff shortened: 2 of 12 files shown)


api/apps/api/src/modules/projects/dto/geo-feature.mapper.ts
```diff
@@ -0,0 +1,16 @@
+import { Injectable } from '@nestjs/common';
+import { GeoFeaturesService } from '../../geo-features/geo-features.service';
+import { PaginationMeta } from '../../../utils/app-base.service';
+import { GeoFeature } from '../../geo-features/geo-feature.api.entity';
+
+@Injectable()
+export class GeoFeatureMapper {
```
hotzevzl, line 7: what do you intend the role of these `Mapper` classes to be? right now they only provide serialization. do you suggest to possibly use these classes for other functionality going forward? I am also slightly confused by the naming as the (current) CRUD-ish service classes do implement a kind of mapper pattern. But besides naming choices, my question is mainly about the architectural intent of these `Mapper` classes.

api/apps/api/src/modules/projects/projects-crud.ts
```diff
@@ -34,7 +32,7 @@ type ProjectFilterKeys = keyof Pick<
 type ProjectFilters = Record<ProjectFilterKeys, string[]>;
 
 @Injectable()
-export class ProjectsService extends AppBaseService<
+export class ProjectsCrud extends AppBaseService<
```
hotzevzl, line 35: I suggest to stick to a qualified noun for these classes (I wouldn't consider CRUD to be a noun but more of an adjective, if anything) - `ProjectsCrudService`? `ProjectsBaseService`?

api/apps/api/src/modules/projects/projects-root.service.ts
```diff
@@ -0,0 +1,67 @@
...
+import { GeoFeaturesService } from '../geo-features/geo-features.service';
+
+import { ProjectsCrud } from './projects-crud';
+import { ProtectedAreasFacade } from './protected-areas/protected-areas.facade';
+import { Project } from './project.api.entity';
+import { CreateProjectDTO } from './dto/create.project.dto';
+import { UpdateProjectDTO } from './dto/update.project.dto';
+
+@Injectable()
+export class ProjectsRootService {
```
hotzevzl: again naming/intent - if this should be come the root/demigod service for Projects, I'd maybe just call it `ProjectsService` 🤔

hotzevzl: though in practice it seems to be a mix of mapper + facade, if I understand the intent?

hotzevzl (commented):
thanks @kgajowy - I agree with clearing up responsibilities in our existing crud-ish services (as we go, I wouldn't do a stop-the-world source code garbage collection run to refactor them all at once 😅) - I am mostly not sure about the suggested naming.

api/apps/api/src/modules/projects/projects-root.service.ts
hotzevzl: ok, I was looking at an earlier commit - agree with keeping the name as `ProjectsService` ofc 😄

kgajowy: 👍 started sweating I missed this 🕺

kgajowy: @hotzevzl the only reason it was named `-root` in first commit is to have a nice diff, otherwise (as in general diff) it was  messed up with new implementation - and thus hard to read.

api/apps/api/src/modules/projects/dto/geo-feature.mapper.ts
kgajowy: @hotzevzl `mapper` is designed to prepare response in the view layer (REST) and is solely aimed for clear responsibilities and separate from transport(access) layer - base service it strongly coupled for JSON Api format.

api/apps/api/src/modules/projects/projects-root.service.ts
hotzevzl: yep, sorry about the noise 😅

api/apps/api/src/modules/projects/dto/geo-feature.mapper.ts
hotzevzl: ok, I agree with the separation. I am still not sure about the name, but I can live with it. would `*Serializer` not convey the intent better? or `*View`? `*ResourceView`? 🤔 
just minor of course, but while we're touching this, I'd try to settle on a solid naming convention - though it may be just me, and all of this is somewhat subjective in any case.

kgajowy: @hotzevzl I am totally fine with any of those, no strong attachment to a particular one. `map` seems to be most adequate and common (mapping one to another) but I may be just "used to" it ;), while `serializing` is rather related to converting into byte stream.

so I will let you decide and adapt 💪

api/apps/api/src/modules/projects/projects-crud.ts
kgajowy: Almost everything can be/is a service thus I feel it may be redundant but sure, we may stick to this convention 👍

api/apps/api/src/modules/projects/dto/geo-feature.mapper.ts
kgajowy: it could also be `marshalling` ; )

few more words on `serializer` - indeed to relates to mapping an object into a format that can be stored/transmitted.

However `Nestjs` uses `serialization` as well https://docs.nestjs.com/techniques/serialization

So again - I am fine with any of those, let's make a choice and stick to it (may be some style/contribution guide?)

api/apps/api/src/modules/projects/projects-crud.ts
kgajowy: https://github.com/Vizzuality/marxan-cloud/pull/218/commits/f48dc4e0f057e25b2366b059a4fbcf55ee6e103a

api/apps/api/src/modules/projects/dto/geo-feature.mapper.ts
kgajowy: https://github.com/Vizzuality/marxan-cloud/pull/218/commits/0f09c1946f706fb33eea96931ef326ccd8d5c019

hotzevzl (approved):

PR #219, opened 2021-05-28
Cost Surface - shapefile converting & extracting PU cost/id

kgajowy:
Includes moving `isDefined` into lib (`util`)

(yes, I messed up branch name with switching surface/cost order 😅 )


kgajowy:
TODO: `e2e` specs don't see libs, config needs to be updated

api/apps/geoprocessing/src/modules/surface-cost/adapters/extract-pu-cost.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+describe(`when given GeoJson isn't a FeatureCollection`, () => {
+  it(`throws exception`, () => {
+    expect(() => sut.extract(fixtures.simpleGeometry())).toThrow(
+      /is supported/,
+    );
+  });
+});
+
+describe(`when given GeoJson has pu costs`, () => {
+  it(`throws exception`, () => {
```
hotzevzl, line 35: the test description doesn't sound right

api/apps/geoprocessing/src/modules/surface-cost/adapters/extract-pu-cost.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+    expect(() => sut.extract(fixtures.geoFeaturesWithoutCost())).toThrow(
+      /missing cost/,
+    );
+  });
+});
+
+describe(`when given GeoJson isn't a FeatureCollection`, () => {
+  it(`throws exception`, () => {
+    expect(() => sut.extract(fixtures.simpleGeometry())).toThrow(
+      /is supported/,
```
hotzevzl, line 29: super minor - we'll see the test passing, but asserting that an exception will be raised containing the text `is supported` seems slightly misleading (why should an error be thrown if *something* "is supported"?): in this specific case, maybe matching case-insensitively the full text of the error (`/Only FeatureCollection is supported/i`) should make things clearer for whoever skims through these tests in the future.

api/apps/geoprocessing/src/modules/surface-cost/adapters/extract-pu-cost.ts
```diff
@@ -0,0 +1,46 @@
+import { isDefined } from '@marxan/utils';
+import { FeatureCollection, GeoJSON, Geometry } from 'geojson';
+import { PlanningUnitCost } from '../ports/planning-unit-cost';
+import { PuExtractorPort } from '../ports/pu-extractor/pu-extractor.port';
+
+type Properties = {
+  cost: number;
+  planningUnitId: string;
+};
+type MaybeProperties = Partial<Properties> | undefined | null;
```
hotzevzl, line 10: this is quite useful in general - I suggest to create a `MaybeSome<T>` and then `type MaybeProperties = MaybeSome<Properties>` or something like that.

api/apps/geoprocessing/src/modules/surface-cost/adapters/extract-pu-cost.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export class ExtractPuCost implements PuExtractorPort {
```
hotzevzl, line 12: I am not too keen on class names based on verbs - why don't we name this `PuCostExtractor`?

hotzevzl: I think this would also match some of your previous code

hotzevzl (approved):

api/apps/geoprocessing/src/modules/surface-cost/adapters/extract-pu-cost.ts
kgajowy: https://github.com/Vizzuality/marxan-cloud/pull/219/commits/573622f92a11fdd126982184dc16a4ab3e03ecb3

api/apps/geoprocessing/src/modules/surface-cost/adapters/extract-pu-cost.ts
kgajowy: https://github.com/Vizzuality/marxan-cloud/pull/219/commits/e9ae13e05dc6fddc7d10fb612c7b378de4b495af

api/apps/geoprocessing/src/modules/surface-cost/adapters/extract-pu-cost.spec.ts
kgajowy: fixed in https://github.com/Vizzuality/marxan-cloud/pull/219/commits/573622f92a11fdd126982184dc16a4ab3e03ecb3

api/apps/geoprocessing/src/modules/surface-cost/adapters/extract-pu-cost.spec.ts
kgajowy: fixed in https://github.com/Vizzuality/marxan-cloud/pull/219/commits/573622f92a11fdd126982184dc16a4ab3e03ecb3

api/apps/geoprocessing/src/modules/surface-cost/adapters/extract-pu-cost.ts
Dyostiq: https://github.com/piotrwitek/utility-types#nullish

PR #222, opened 2021-05-31
Project new

mbarrenechea:
### Overview

This PR adds:
- [x] Country, region, and subregion bbox move the map
- [x] Form scroll to prevent weird behavior in the map


mbarrenechea:
I will continue in smaller PRs

PR #224, opened 2021-05-31
feat(api): scenario planning units: integrity rule

kgajowy:
This rule will be used by both API and Geoprocessing when updating PU lock status.

Recommended review order:
1. https://github.com/Vizzuality/marxan-cloud/pull/221 
2. https://github.com/Vizzuality/marxan-cloud/pull/224 (this)
3. https://github.com/Vizzuality/marxan-cloud/pull/225


api/libs/scenarios-planning-unit/src/domain/lock-planning-units.ts
```diff
@@ -0,0 +1,11 @@
+import { differenceWith } from 'lodash';
+
+export const lockPlanningUnits = (
```
hotzevzl, line 3: I suggest to clarify the function name - this function is not actually locking units nor even returning a difference of the lockable units, but checking if the units to lock are a non-strict subset of the available ones.

api/libs/scenarios-planning-unit/src/domain/lock-planning-units.ts
```diff
@@ -0,0 +1,11 @@
...
+  planningUnitsIds: string[],
+  availablePlanningUnitsIds: string[],
+): { errors: string[] } => {
+  const diff = differenceWith(planningUnitsIds, availablePlanningUnitsIds);
```
hotzevzl, line 7: `_.difference()` should be enough here as we don't pass in a comparator function

api/libs/scenarios-planning-unit/src/domain/lock-planning-units.ts
```diff
@@ -0,0 +1,11 @@
...
+  return {
+    errors: diff.map((missingId) => `Missing ${missingId}`),
```
hotzevzl, line 9: IIRC we tried to improve these error messages recently, while trying to not make them too verbose as they actually only differ by the PU id... Maybe just `Planning Unit with id ${missingId} is not part of this project`?

hotzevzl: or tweak this into a `canPlanningUnitsBeLocked()` and return true if diff is empty, false otherwise?

api/libs/scenarios-planning-unit/src/domain/lock-planning-units.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+  [[], fixtures.availablePlanningUnits, 0],
+  [fixtures.someOfTheAvailable, [], fixtures.someOfTheAvailable.length],
+])(
+  `when changing %p out of %p, it returns %p errors`,
+  (unitsToChange, availableUnits, errorsCount) => {
+    expect(
+      lockPlanningUnits(unitsToChange, availableUnits).errors.length,
+    ).toEqual(errorsCount);
+  },
+);
```
hotzevzl, line 22: while the test makes sense, the ambiguity of the `lockPlanningUnits()` function (see comment below), or my misunderstanding if it, seems to make this test not so obvious - aka I am not sure what exactly we want to test here.

- `lockPlanningUnits()` doesn't seem to do what the name suggests (see below)
- I'd use different ids (even just an array of four UUIDv4s, a subset of two elements of this array, and a strict subset + another generated UUID) to avoid any possible confusion with the lock status values, which use similar numbers to the ones in the fixtures

Dyostiq (approved):

api/libs/scenarios-planning-unit/src/domain/lock-planning-units.ts
kgajowy: I would stay with returning message (or maybe ids in future) and let consumer decide.

api/libs/scenarios-planning-unit/src/domain/lock-planning-units.ts
kgajowy: @hotzevzl fixed in https://github.com/Vizzuality/marxan-cloud/pull/224/commits/8b9c4a7bbed5d99bc818bedd331199dc29415f8e

api/libs/scenarios-planning-unit/src/domain/lock-planning-units.ts
kgajowy: @hotzevzl fixed in https://github.com/Vizzuality/marxan-cloud/pull/224/commits/8b9c4a7bbed5d99bc818bedd331199dc29415f8e

api/libs/scenarios-planning-unit/src/domain/lock-planning-units.ts
kgajowy: @hotzevzl fixed in https://github.com/Vizzuality/marxan-cloud/pull/224/commits/8b9c4a7bbed5d99bc818bedd331199dc29415f8e (message)

api/libs/scenarios-planning-unit/src/domain/lock-planning-units.spec.ts
kgajowy: @hotzevzl fixed in https://github.com/Vizzuality/marxan-cloud/pull/224/commits/8b9c4a7bbed5d99bc818bedd331199dc29415f8e - more verbose names, added named types as well

hotzevzl (approved):

PR #225, opened 2021-05-31
feat(geoprocessing): scenario planning units: use integrity rule

kgajowy:
* uses integrity rule from https://github.com/Vizzuality/marxan-cloud/pull/224
* adds the last piece of repo for adapter - and adjusts adapter to only find required things for integrity

Recommended review order:
1. https://github.com/Vizzuality/marxan-cloud/pull/221 
2. https://github.com/Vizzuality/marxan-cloud/pull/224 
3. https://github.com/Vizzuality/marxan-cloud/pull/225 (this)


api/apps/geoprocessing/src/modules/surface-cost/adapters/typeorm-available-planning-units.ts
```diff
@@ -0,0 +1,28 @@
+import { Repository } from 'typeorm';
+import { Injectable } from '@nestjs/common';
+import { InjectRepository } from '@nestjs/typeorm';
+
+import { ScenariosPlanningUnitGeoEntity } from '@marxan/scenarios-planning-unit';
+
+import { GetAvailablePlanningUnits } from '../ports/available-planning-units/get-available-planning-units';
+
+@Injectable()
+export class TypeormAvailablePlanningUnits
```
hotzevzl, line 10: hm, I get this is the point of a decoupled architecture, but realistically we're not going to use anything else than TypeORM in this context - and if we did, we would likely replace the use of TypeORM with something else in a whole adaptor, not just in part of it.

To me, this class is a plain getter of the db state of things, happening to do so via TypeORM as an implementation detail - I would avoid proliferation of naming conventions and prefixes, and rename it to something like `AvailablePlanningUnits`.

Moreover, I would invite to avoid breaking things down so finely unless there's a compelling reason to - this class' `get()` method could pretty much be `ScenariosService.getPlanningUnits()` - it could still be composed via an interface while implementing this and other interfaces in a class such as `ScenariosService`.

It may be my FP neurons, but class or interface names that use verbs (such as `GetAvailablePlanningUnits`) are a bit of a smell that we are breaking things down too aggressively (for the kind and size of domain at hand, at least).

api/apps/geoprocessing/src/modules/surface-cost/application/surface-cost-processor.spec.ts
```diff
@@ -82,43 +85,43 @@ describe(`when shapefile was converted to geojson`, () => {
...
       await expect(sut.process(getJob(scenarioId))).rejects.toThrow(/Missing/);
-      expect(puExtractor.mock).toHaveBeenCalledWith(getGeoJson());
+      expect(puExtractor.mock).toHaveBeenCalledWith(
+        getGeoJson(availablePlanningUnitIds),
+      );
     });
   });
 
   describe(`when cost was resolved`, () => {
+    const cost = getCost(availablePlanningUnitIds);
```
hotzevzl, line 95: minor, especially since this is a test, but I think this function and its use may be clearer if we called it `getCostSurface()` or something like this

api/apps/geoprocessing/src/modules/surface-cost/application/surface-cost-processor.ts
```diff
@@ -6,25 +6,30 @@ import { CostSurfaceJobInput } from '../cost-surface-job-input';
...
+    const planningUnitsIds = (
+      await this.availablePlanningUnits.get(job.data.scenarioId)
+    ).ids;
+
+    const { errors } = lockPlanningUnits(
       surfaceCosts.map((cost) => cost.planningUnitId),
+      planningUnitsIds,
     );
     if (errors.length > 0) {
       throw new Error(errors.join('.'));
```
hotzevzl, line 35: given the way we use these errors, and referencing my earlier comment about this, I think we could even return the ids of "unknown" scenario-planning-units and generate here the error message to tell the user which ones are not part of the scenario - we leak abstractions a bit, but at the end this is still a `CostSurfaceProcessor` (or `SurfaceCostProcessor`? 😅), and I think we can assume we know about the concept of scenario-planning-units in this class.

api/apps/geoprocessing/src/modules/surface-cost/application/surface-cost-processor.ts
```diff
@@ -6,25 +6,30 @@ import { CostSurfaceJobInput } from '../cost-surface-job-input';
...
   ) {}
 
   async process(job: Job<CostSurfaceJobInput, true>): Promise<true> {
     const geoJson = await this.shapefileConverter.convert(job.data.shapefile);
     const surfaceCosts = this.puExtractor.extract(geoJson);
-    const { errors } = await this.puValidator.validate(
-      job.data.scenarioId,
+    const planningUnitsIds = (
+      await this.availablePlanningUnits.get(job.data.scenarioId)
+    ).ids;
```
hotzevzl, line 28: linking to my earlier comment about breaking things too finely - ok, it's easier to test this, but after all in much of this we're after a list of scenario-planning-units, calculating a difference, and persisting the associated costs if there are no errors - aren't we introducing too many moving pieces here for something that we don't expect to need to change?

I would suggest to also call this `const scenarioPlanningUnitIds` to make it clear which set of PUs this const is expected to hold.

api/apps/geoprocessing/src/modules/surface-cost/adapters/typeorm-available-planning-units.ts
Dyostiq: Prefixing `TypeORM` is one of the strategies to avoid naming implementations the same way as their interfaces.

One of the reasons behind breaking things down so finely is that to make the code more unit testable (and have a lower cognitive load due to a smaller context) – especially when we have the interface (of course, a class can implement a bunch of them). For me, during refactor is easier to join things than split.

Using verbs in names are CQRS-ish, one class - one purpose.

kgajowy: @hotzevzl  how about `DbAvailablePlanningUnits`/`AvailablePlanningUnitsRepository` then? it hides implementation details (a little) but still exposes what is the aim of the adapter. It can be it could be resolved from something different.

```
 it could still be composed via an interface while implementing this and other interfaces in a class such as ScenariosService
``` 
it could be, even we got such case somewhere. The *service tho is "hard" to test and does not exist in geoprocessing context.

Verbs - as @Dyostiq stated was the intention. I agree tho to name it just `AvailablePlanningUnitsRepository`

api/apps/geoprocessing/src/modules/surface-cost/application/surface-cost-processor.spec.ts
kgajowy: well we hit naming things again ; ) its already in `cost surface` context, so would adding it again deliver more value to reader?

api/apps/geoprocessing/src/modules/surface-cost/application/surface-cost-processor.ts
kgajowy: The message already does contain those IDs - and the data (of failure) is not (for now on) processed in any manner 🤔

api/apps/geoprocessing/src/modules/surface-cost/adapters/typeorm-available-planning-units.ts
hotzevzl: @kgajowy `AvailablePlanningUnitsRepository` sounds good, reuses concepts already in use and does not bind to orm/db.

And yes, I agree that almost-isolated functions are easier to mock and test, as well as an opportunity to use as precise as possible/needed typing.

I am fine with the approach - in practice, this is a way to namespace things and separate providers, but I would like to all converge on a way to do this consistently and to make this easy to read, or we end up having a bit of a patchwork of similar but slightly different rationales for breaking things down, using them and so on, which to me is a bit of a highway to the dreaded ball of mud :)

Right now, what I think I need to do even just to review a PR is:
- check the list of providers in a module
- check their interfaces (as we go on this becomes less of a cognitive load as I'd have seen most of them in earlier PRs)
- check their implementations
- check how all the pieces fit together and are used in the overall workflow of the feature

And frankly, I think this is a bit of a readability issue (with related cognitive load) mostly because we don't have a default structure for this, yet at least, and each module ends up being slightly off and ad-hoc, for problems that may not be all that unique.

I would be much more comfortable if we could align on a workflow (with all the relevant differences) equivalent to the general structure of BaseService - in its dumbness and limitations (coupling and so on), it does at least provide a very simple and understandable workflow for crud components where the base class describes all the default workflows and related lifecycle "hooks".

I am not even sure what to suggest as way forward as I am not yet seeing enough patterns emerging from recent work, from which I could propose a general default structure for feedback and alignment (happy to accept such a proposal though, as you may have something in mind already at least implicitly while you've been working on these components).

However, after having gone on a huge semi-philosophical tangent: for the issue at hand in this PR comment: let's do as suggested (`AvailablePlanningUnitsRepository`), and if we can stick to `GetSomething` for the abstract class that only implements a `get()` method, and `SomethingRepository` for the implementation, I'd say we would already have some understandable convention. I'm still not a fan of naming classes or interfaces with a verb, but if thinking about them as namespaces I think I can live with this convention :)

api/apps/geoprocessing/src/modules/surface-cost/application/surface-cost-processor.spec.ts
hotzevzl: no problem in leaving as is as it's just a test, but well, we are returning a cost surface, not a cost: if going by context, should a reader assume we return the *cost of a cost surface*? but that's not even a thing in Marxan parlance, maybe `getSurface()` or `getCostByPlanningUnit()` would be more closely descriptive. bug again, not an issue in practice in this specific context.

api/apps/geoprocessing/src/modules/surface-cost/application/surface-cost-processor.ts
hotzevzl: yep, we discussed this (in a similar context? can't remember the details) - just saying that if we concatenate errors here, we may as well also only return the naked ids from `canPlanningUnitsBeLocked()` and do the templating here instead of returning full error messages, but I am happy with your latest updates in `canPlanningUnitsBeLocked()`.

api/apps/geoprocessing/src/modules/surface-cost/adapters/typeorm-available-planning-units.ts
kgajowy: @hotzevzl 
`Right now, what I think I need to do even just to review a PR is:`
not entirely. Application (without adapters implementation of `useClass`) is one main piece of given functionality and reviewing (even with test) is purely enough to know (also for newcomers) how it works. Implementation details of storage are just... details, kind of irrelevant from the product perspective. If one is interested, may jump into implementation details. It makes it a little more intuitive when getting familiar with whole feature - like reading API doc first what is available and what it does.

> check how all the pieces fit together and are used in the overall workflow of the feature

integration test is a perfect purpose to verify all the linkings and overall workflow to get familiar with "usage".

> However, after having gone on a huge semi-philosophical tangent
its always a pleasure to discuss things in such manner with skilled people :)

api/apps/geoprocessing/src/modules/surface-cost/application/surface-cost-processor.spec.ts
kgajowy: fixed in https://github.com/Vizzuality/marxan-cloud/pull/225/commits/a6068c200c2bb6245a9b13cdd2fdb10f3134a821

api/apps/geoprocessing/src/modules/surface-cost/adapters/typeorm-available-planning-units.ts
kgajowy: fixed in https://github.com/Vizzuality/marxan-cloud/pull/225/commits/a6068c200c2bb6245a9b13cdd2fdb10f3134a821

api/apps/geoprocessing/src/modules/surface-cost/application/surface-cost-processor.ts
kgajowy: https://github.com/Vizzuality/marxan-cloud/pull/225/commits/59f75722ee095cd6518244024ecadf51e85f618f

PR #228, opened 2021-05-31
feat: implement artifact db cache for cost surfaces

Dyostiq:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/libs/scenario-cost-surface/src/cost-surface-file-cache.api.entity.ts
+++ b/api/libs/scenario-cost-surface/src/cost-surface-file-cache.api.entity.ts
@@ -1,0 +1,52 @@
+import {
+  Column,
+  CreateDateColumn,
+  Entity,
+  Index,
+  PrimaryColumn,
+  UpdateDateColumn,
+} from 'typeorm';
+
+export enum ArtifactType {
+  CostTemplate = 'CostTemplate',
+}
+
+@Entity('cost_surface_file_cache')
+@Index(['scenarioId', 'artifactType'])
+export class CostSurfaceFileCache {
+  @PrimaryColumn({
+    generated: 'uuid',
+    type: 'uuid',
+  })
+  id!: string;
+
+  @Column()
+  scenarioId!: string;
+
+  @Column({ type: 'int4', nullable: true })
+  artifact?: number;
+
+  @Column({
+    type: 'enum',
+    enum: ArtifactType,
+  })
+  artifactType!: ArtifactType;
+
+  @Column()
+  contentType!: string;
+
+  @CreateDateColumn({ type: 'timestamp without time zone' })
+  createdAt!: Date;
+
+  @UpdateDateColumn({ type: 'timestamp without time zone' })
+  lastModifiedAt!: Date;
+}
+
+export const createCostSurfaceFileCacheFields = [
+  'scenarioId',
+  'artifactType',
+  'contentType',
+] as const;
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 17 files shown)


api/libs/scenario-cost-surface/src/cost-surface-file-cache.api.entity.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  @UpdateDateColumn({ type: 'timestamp without time zone' })
+  lastModifiedAt!: Date;
+}
+
+export const createCostSurfaceFileCacheFields = [
+  'scenarioId',
+  'artifactType',
+  'contentType',
+] as const;
```
kgajowy, line 49: is it an entity responsibility to decide which files are under cache? 🤔

Dyostiq: the const is defining only which fields are needed to create an instance

api/libs/scenario-cost-surface/src/cost-surface-file-cache.api.entity.ts
```diff
@@ -0,0 +1,49 @@
...
+@Index(['scenarioId', 'artifactType'])
+export class CostSurfaceFileCache {
+  @PrimaryColumn({
+    generated: 'uuid',
+    type: 'uuid',
+  })
+  id!: string;
+
+  @Column()
+  scenarioId!: string;
```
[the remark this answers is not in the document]
Dyostiq: AFAIR the generator does not work, so yes, it is written by hand.
The scenario id is a key here, artifacts are generated per scenario. I can drop the reference, but then the validation and cascading are dropped. Honestly, I don't have a strong opinion on that.

Moving the scenario is rather not feasible, as it has other references and the entire graph would land in the lib.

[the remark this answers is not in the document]
kgajowy: May want to just drop a note then why the difference is there. I am fine with the pros/cons.

api/libs/scenario-cost-surface/src/cost-surface-file-cache.api.entity.ts
kgajowy: just for a sake of security of typing, shouldn't they relate to actual fields?

api/libs/scenario-cost-surface/src/cost-surface-file-cache.api.entity.ts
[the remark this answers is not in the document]
hotzevzl: we've not been using entities as source of truth so far, so it's ok to have the migration be the source of truth here

api/libs/scenario-cost-surface/src/cost-surface-file-cache.api.entity.ts
hotzevzl: it should be ok as https://github.com/Vizzuality/marxan-cloud/pull/228/files/14076746b22f7c4e320e8e6473503343e85b8ba3#diff-0298940b2ab35a3d32a4536133f253445f3c2ce55b768011266c254e8fd80596R42 would bark if we used this array including in it elements which don't match props of `CostSurfaceFileCache`

Dyostiq: I've added a `type check` for that

kgajowy (approved):

hotzevzl (approved):

hotzevzl:
@Dyostiq feel free to merge this after merging its base branch if you are happy with it and don't need to apply any further changes

PR #232, opened 2021-06-02
refactor: move admin_regions entity to shared lib

kgajowy:
Recreated almost every column definition with the exception of `AdminLevel` enum - TypeORM keeps generating changes due to:
```
"level" adm_level NOT NUL
```

```
"level" "adm_level" NOT NUL
```

which follows to re-creating primary keys & indexes; other than that, entity fully matches database definition (but will still generate few sqls for migration :( ) - I gave up at this point.


api/libs/admin-regions/src/admin-areas.geo.entity.ts
```diff
@@ -0,0 +1,125 @@
...
+  PrimaryGeneratedColumn,
+  UpdateDateColumn,
+} from 'typeorm';
+import { BBox, Geometry } from 'geojson';
+import { defaultSrid } from '@marxan/utils/geo';
+
+export enum AdminLevel {
+  Country = 'country',
+  Adm1 = 'adm_1',
+  Amd2 = 'adm_',
```
hotzevzl, line 18:
```suggestion
  Amd2 = 'adm_2',
```

hotzevzl (approved):
except for a typo (see change suggestions), looks good.

As we've been creating migrations manually so far, I would not be much concerned with trying to fully get entities to be source of truth, especially as the schema changes I expect us to have to deal with in this phase should be fairly small, so I (at least 😅) could live with keeping writing migrations manually.

hotzevzl (approved):

PR #234, opened 2021-06-02
feat(api): projects: verify admin regions and assign bbox

kgajowy:
- [x] convert pu- jobs to use QueueModule to be able to intercept job submission
- [x] remove redundant/invalid tests/params from json-api specs
- [x] refactor project specs for brevity 
- [x] refactor protected-areas-jobs-triggering-for-admin-areas specs for brevity (previously copied from projects)
- [ ] fix/refactor `protected-areas` specs


api/apps/api/src/modules/projects/projects-crud.service.ts
```diff
@@ -199,4 +204,57 @@ export class ProjectsCrudService extends AppBaseService<
       return this.planningUnitsService.create(createModel);
     }
   }
+
+  async validateBeforeCreate(
```
hotzevzl, line 208: if we want to avoid mutating in place, we can use `setDataCreate()` and `setDataUpdate()`

kgajowy:
@hotzevzl as per discussion, skipping tests interfering with project creation side effects (they could be refactored tho in future)

api/apps/api/src/modules/projects/projects-crud.service.ts
```diff
@@ -199,4 +213,51 @@ export class ProjectsCrudService extends AppBaseService<
...
+    } catch (error) {
+      // currently no easy way to propagate specific error to controller, thus throwing Http error already
+      throw new BadRequestException(error);
+    }
+
+    if (geometry) {
+      return geometry;
+    } else {
+      throw new InternalServerErrorException(
+        `Missing bbox for given admin region.`,
```
hotzevzl, line 259: not a big deal, but since we already have the planning area, it could be useful to add country/l1/l2 (where they are defined) to the error message - that we have no bbox should normally not happen, but when looking at log traces in staging/prod, this info can help get to the right bit of (incorrect) data easily.

api/apps/api/test/project-protected-areas/protected-areas.e2e-spec.ts
```diff
@@ -344,64 +255,6 @@ describe('ProtectedAreasModule (e2e)', () => {
...
         });
 
-        /**
-         * @todo Enable this test once we start supporting user uploads of
-         * protected area geometries. Until then, with the data available in our
-         * seed for tests, the WDPA areas we select pretending that they are
-         * 'project-specific' end up coinciding with the WDPA areas already
-         * selected via their IUCN categories, so the test would fail.
-         */
-        test.skip('As a user, when I update a scenario to which custom protected areas are already associated, these should be preserved if I only add WDPA protected areas to it via their IUCN category', async () => {
```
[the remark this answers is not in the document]
kgajowy: @hotzevzl this scenario is already being (somehow) processed in `api/apps/api/test/project-protected-areas/project-protected-areas-upload-shapefile.e2e-spec.ts`.  TBH I don't recognize this requirement from JIRA when I was implementing the upload feature - may be we should create a standalone task to ensure this feature is included?

Nevertheless, keeping unused test only bloats this file without real value and makes the maintenance hard :( (not to mention it tests serialization and pagination as well)

api/apps/api/test/projects.e2e-spec.ts
```diff
@@ -145,30 +88,6 @@ describe('ProjectsModule (e2e)', () => {
       expect(jsonAPIResponse.data[0].type).toBe('projects');
     });
 
-    test.skip('A user should be able to get a list of projects and related scenarios', async () => {
```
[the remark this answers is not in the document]
kgajowy: restored in https://github.com/Vizzuality/marxan-cloud/pull/234/commits/d4f780c24130947d3b43a81ecb3b2e15d10cf425

api/apps/api/src/modules/projects/projects-crud.service.ts
kgajowy: @hotzevzl correct me if I am wrong but if it is Http Error, all the payload (assuming redact of some sensitive data) should be there already - they are usually captured by monitoring services (like Sentry). What benefit I am missing?

hotzevzl: you are right - I was thinking about our current `AllExceptionsFilter`, where we do log some error data, but not the whole request or response (or relevant parts of them at least) - indeed, adding at least request payload to what we log (maybe behind a feature flag) could help to provide better context to our raw logs (even without using a monitoring service).

Let's leave this as is then - by logging request payloads we could get context for all errors rather than having to care about adding some context to individual exception being thrown.

api/apps/api/test/project-protected-areas/protected-areas.e2e-spec.ts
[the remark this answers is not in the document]
hotzevzl: you are right, this is not explicitly specified (for reference, the design that sets this scenario is this: https://www.figma.com/file/hq0BZNB9fzyFSbEUgQIHdK/Marxan-Visual_V02?node-id=2991%3A1853)... it "should work" by virtue of not touching any existing associations of project-specific protected areas when we link WDPA protected areas, so in a way this test should really be resurrected (and made it work).

I have created https://vizzuality.atlassian.net/browse/MARXAN-449 for resurrecting the test - feel free to remove it if you wish.

hotzevzl (approved):

hotzevzl:
@kgajowy I'd merge this as is if you are happy with it, and add refactoring of specs missing from your todo list for this PR in a separate small PR- feel free to merge

PR #240, opened 2021-06-07
feat(api): projects: include most-specific admin area name/id

kgajowy:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/projects/projects-crud.service.ts
+++ b/api/apps/api/src/modules/projects/projects-crud.service.ts
@@ -24,6 +24,7 @@ import { AppConfig } from '@marxan-api/utils/config.utils';
 import { AdminArea } from '@marxan/admin-regions';
 import { BboxResolver } from './bbox/bbox-resolver';
 import { Polygon } from 'geojson';
+import { FetchSpecification } from 'nestjs-base-service';
 
 const projectFilterKeyNames = [
   'name',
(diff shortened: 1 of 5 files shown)


api/apps/api/src/modules/projects/projects-crud.service.ts
```diff
@@ -78,6 +79,8 @@ export class ProjectsCrudService extends AppBaseService<
         'scenarios',
         'createdAt',
         'lastModifiedAt',
+        'planningAreaId',
+        'planningAreaName',
```
hotzevzl, line 83: This will not work, because `_processBaseFilter()` will try to add a SQL `AND WHERE` but these are not db columns

hotzevzl: but it shouldn't be a problem to drop these two elements from this array as we'd not need to filter projects by area id or name

kgajowy: @hotzevzl 
image

without those, fields are not available in response. What do you recommend?

kgajowy: (feel free to play with it `project/project-get.e2e-spec.ts:` )

hotzevzl: sorry, total oversight on my side - I read this in the wrong context because of the way it was presented in the PR diff. I was convinced I was looking at the array of prop names against which to filter. Ignore 😬

hotzevzl (approved):

PR #241, opened 2021-06-07
MARXAN - 448 Cost surface - scenarioId guard

kgajowy:
Similar to `projects`:
* move current service to `crud-service` and wrap existing functions,
* the ones that act on a given `scenarioId` received existing `guard` which verifies existence of given Scenario,
* move serializers to standalone classes
* leaves space for ACL stuff
* "fixed" spec for changing planning unit's inclusivity (by id / geo) - it was running without existing scenario id, added check prevents that

### DIFF
--- a/api/apps/api/src/modules/scenarios/scenario.service.ts
+++ b/api/apps/api/src/modules/scenarios/scenario.service.ts
@@ -1,0 +1,21 @@
+import { HttpService, Injectable } from '@nestjs/common';
+import { AppConfig } from '@marxan-api/utils/config.utils';
+import { ScenariosCrudService } from '@marxan-api/modules/scenarios/scenarios-crud.service';
+import { ScenarioFeaturesService } from '@marxan-api/modules/scenarios-features';
+import { AdjustPlanningUnits } from '@marxan-api/modules/analysis/entry-points/adjust-planning-units';
+import { CostSurfaceFacade } from '@marxan-api/modules/scenarios/cost-surface/cost-surface.facade';
+
+@Injectable()
+export class ScenarioService {
+  private readonly geoprocessingUrl: string = AppConfig.get(
+    'geoprocessing.url',
+  ) as string;
+
+  constructor(
+    public readonly service: ScenariosCrudService,
+    private readonly scenarioFeatures: ScenarioFeaturesService,
+    private readonly updatePlanningUnits: AdjustPlanningUnits,
+    private readonly costSurface: CostSurfaceFacade,
+    private readonly httpService: HttpService,
+  ) {}
+}
--- a/api/apps/api/src/modules/scenarios/scenario.service.ts
+++ b/api/apps/api/src/modules/scenarios/scenario.service.ts
@@ -1,9 +1,19 @@
 import { HttpService, Injectable } from '@nestjs/common';
+import { FetchSpecification } from 'nestjs-base-service';
+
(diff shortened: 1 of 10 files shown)


api/apps/api/src/modules/scenarios/scenario.service.ts
```diff
@@ -0,0 +1,117 @@
...
+import { CostSurfaceFacade } from './cost-surface/cost-surface.facade';
+import { ScenariosCrudService } from './scenarios-crud.service';
+import { JobStatus } from './scenario.api.entity';
+
+import { CreateScenarioDTO } from './dto/create.scenario.dto';
+import { UpdateScenarioDTO } from './dto/update.scenario.dto';
+import { UpdateScenarioPlanningUnitLockStatusDto } from './dto/update-scenario-planning-unit-lock-status.dto';
+
+@Injectable()
+export class ScenarioService {
```
hotzevzl, line 19: I suggest to stick to `<Plural>Service` (as for the recently-refactored `ProjectsService`), and `<Plural>CrudService` (this is already ok in this PR for `ScenariosCrudService`). Filename to be adjusted accordingly here.

api/apps/api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -92,8 +83,8 @@ export class ScenariosController {
   async findAll(
     @ProcessFetchSpecification() fetchSpecification: FetchSpecification,
   ): Promise<ScenarioResult> {
-    const results = await this.service.findAllPaginated(fetchSpecification, {});
-    return this.service.serialize(results.data, results.metadata);
+    const results = await this.service.findAll(fetchSpecification);
+    return this.scenarioSerializer.serialize(results.data, results.metadata);
```
hotzevzl, line 87: we should keep pagination here

hotzevzl (approved):
@kgajowy nice cleanup, thank you!

see suggestion about sticking to existing naming convention for services (I think yours was simply a typo here) and about a pagination that was removed - once you've taken care of these, feel free to merge, thanks.

api/apps/api/src/modules/scenarios/scenario.service.ts
kgajowy: @hotzevzl I actually forgot to mention this but I have done this "miss" on purpose - to have a clean diff (otherwise git would mess with diff merging new/old content).

kgajowy: (as you already reviewed, will change it in this PR)

api/apps/api/src/modules/scenarios/scenarios.controller.ts
kgajowy:
```
async findAll(fetchSpecification: FetchSpecification) {
    return this.crudService.findAllPaginated(fetchSpecification);
  }
```
we do 🤔 

Will change tho.

kgajowy: https://github.com/Vizzuality/marxan-cloud/pull/241/commits/8e4e9a68d948000c3dda981102d5afb8a224884d

api/apps/api/src/modules/scenarios/scenario.service.ts
kgajowy: https://github.com/Vizzuality/marxan-cloud/pull/241/commits/8e4e9a68d948000c3dda981102d5afb8a224884d

PR #244, opened 2021-06-08
feat(api): scenario's job status - API

kgajowy:
image

image

### DIFF
--- a/api/apps/api/src/modules/projects/job-status/job-status.service.ts
+++ b/api/apps/api/src/modules/projects/job-status/job-status.service.ts
@@ -1,0 +1,33 @@
+import { Injectable } from '@nestjs/common';
+import { ApiEventsService } from '@marxan-api/modules/api-events/api-events.service';
+import { ProjectsService } from '@marxan-api/modules/projects/projects.service';
+import { ScenariosService } from '@marxan-api/modules/scenarios/scenarios.service';
+
+@Injectable()
+export class JobStatusService {
+  constructor(
+    private readonly apiEvents: ApiEventsService,
+    private readonly scenariosService: ScenariosService,
+    private readonly projectsService: ProjectsService,
+  ) {}
+
+  async getJobStatusFor(projectId: string): Promise<void> {
+    await this.projectsService.findOne(projectId);
+
+    // get status of project job(s) ?
+
+    // get all scenarios for given project
+
+    // for each scenario, find its jobs and relevant statuses
+
+    /**
+     *
+     * Draft of SQL
+     *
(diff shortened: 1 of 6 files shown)


api/apps/api/src/modules/projects/projects.controller.ts
```diff
@@ -4,6 +4,7 @@ import {
   Delete,
   Get,
   NotFoundException,
+  NotImplementedException,
```
Dyostiq, line 7: not used

api/apps/api/src/modules/projects/job-status/job-status.service.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    /**
+     *
+     * Draft of SQL
+     *
+     * SELECT kind, topic, MAX(timestamp) from api_events where
+     --topic in ('03fb678e-689b-473c-af80-6915685a53a8', 'ce2069ee-2925-4c73-a328-882447e6c84d') and
+     ( kind like 'project.protectedAreas%' or kind like 'user.account%' )
+     group by kind, topic
+     */
```
Dyostiq, line 31: a couple of ideas:
1. have a model of the status and rehydrate its state from events,
2. have a read model of the status and react to events to update the read model and save,
3. have a view that calculates the current status
4. use a plain sql as suggested (what is very similar to a view, though)

Dyostiq (approved):

api/apps/api/src/modules/projects/job-status/job-status.service.ts
kgajowy: Great suggestions @Dyostiq !

@hotzevzl your take? I would love to have some more real ES there but knowing the deadline it may be tricky :(

kgajowy:
@hotzevzl needed this to move forward, feel free to add comments anytime (will apply them in standalone PR)

api/apps/api/src/modules/projects/job-status/job-status.service.ts
Dyostiq: The another idea is to go with TypeORM ViewEntity and when needed transform it to a regular entity – a read model reacting on events. Then, the switch should be seamless.

hotzevzl: yep, digging into event sourcing at this stage may not be that good of an idea.

we already have a basic view (two actually) that approximate this - they just sort events by timestamp desc as we adapted them from a different implementation where we needed what they do. I would go for a view and suitable indexes. what do you think?

PR #254, opened 2021-06-09
test(api): project-jobs-status

kgajowy:
Failing test is fine - it should be "green" once valid implementation is in place.


api/apps/api/src/modules/scenarios/scenarios-crud.service.ts
```diff
@@ -80,6 +80,7 @@ export class ScenariosCrudService extends AppBaseService<
   get serializerConfig(): JSONAPISerializerConfig<Scenario> {
     return {
       attributes: [
+        'id',
```
kgajowy, line 83: Was missing in response, even tho it was marked as "required" in dto/entity.

hotzevzl (approved):

PR #264, opened 2021-06-14
feat(api): allow to get scenarios solutions results

kgajowy:
This PR mainly touches base implementation of an API layer and leaves space for actual implementation; includes:
* entity 
* dtos
* crud-service based on AppBaseService

It may happen that some `runId` or other fields aren't correct (according to discussion on slack) but I would opt in for leaving it for future PR once it is well settled.

image

image

image

### DIFF
--- a/api/apps/api/src/modules/scenarios/dto/scenario-solution-result.dto.ts
+++ b/api/apps/api/src/modules/scenarios/dto/scenario-solution-result.dto.ts
@@ -1,0 +1,46 @@
+import {
+  ApiExtraModels,
+  ApiProperty,
+  getSchemaPath,
+  refs,
+} from '@nestjs/swagger';
+import { ScenariosOutputResultsGeoEntity } from '@marxan/scenarios-planning-unit';
+import { oneOf } from 'purify-ts';
+
+class ScenarioSolutionsDataDto {
+  @ApiProperty()
+  type: 'solutions' = 'solutions';
+
+  @ApiProperty()
+  id!: string;
+
+  @ApiProperty({
+    isArray: true,
+    type: () => ScenariosOutputResultsGeoEntity,
+  })
+  attributes!: ScenariosOutputResultsGeoEntity[];
+}
+
+class ScenarioSolutionDataDto {
+  @ApiProperty()
+  type: 'solution' = 'solution';
+
+  @ApiProperty()
+  id!: string;
+
+  @ApiProperty({
+    type: () => ScenariosOutputResultsGeoEntity,
+  })
+  attributes!: ScenariosOutputResultsGeoEntity;
+}
+
+@ApiExtraModels(ScenarioSolutionsDataDto, ScenarioSolutionDataDto)
+export class ScenarioSolutionResultDto {
+  @ApiProperty({
+    oneOf: [
+      { $ref: getSchemaPath(ScenarioSolutionsDataDto) },
+      { $ref: getSchemaPath(ScenarioSolutionDataDto) },
+    ],
+  })
+  data!: ScenarioSolutionsDataDto | ScenarioSolutionDataDto;
+}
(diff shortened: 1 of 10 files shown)


api/apps/api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -170,4 +177,53 @@ export class ScenariosController {
...
+  @ApiImplicitQuery({
+    name: 'best',
+    required: false,
+    type: Boolean,
+  })
+  @ApiImplicitQuery({
+    name: 'most-different',
+    required: false,
+    type: Boolean,
+  })
```
Dyostiq, line 193: `ApiImplicitQuery` is deprecated/not in API anymore, you should use `ApiQuery`

api/apps/api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -170,4 +177,53 @@ export class ScenariosController {
...
+  @JSONAPIQueryParams()
+  @Get(`:id/marxan/run/:runId/solutions`)
+  async getScenarioRunSolutions(
+    @Param('id', ParseUUIDPipe) id: string,
+    @Param('runId', ParseUUIDPipe) runId: string,
+    @ProcessFetchSpecification() fetchSpecification: FetchSpecification,
+    @Query('best', ParseBoolPipe) selectOnlyBest?: boolean,
+    @Query('most-different', ParseBoolPipe) selectMostDifferent?: boolean,
```
Dyostiq, line 201: they seem to be mutually exclusive
maybe it's not a problem, though
also a dto could be used here

Dyostiq (approved):

api/apps/api/src/modules/scenarios/scenarios.controller.ts
kgajowy: yes, they are exclusive (and returning different DTO). Any idea how to show it in code?

Dyostiq: honestly I'd create another method and route for it, trying to show it in the code with one method is tricky (especially when you'd like to keep it in the openapi spec) and can be not worth the effort

api/apps/api/src/modules/scenarios/dto/scenario-solution-result.dto.ts
```diff
@@ -0,0 +1,46 @@
+import {
+  ApiExtraModels,
+  ApiProperty,
+  getSchemaPath,
+  refs,
+} from '@nestjs/swagger';
+import { ScenariosOutputResultsGeoEntity } from '@marxan/scenarios-planning-unit';
+import { oneOf } from 'purify-ts';
```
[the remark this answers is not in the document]
kgajowy: thanks, fixed in https://github.com/Vizzuality/marxan-cloud/pull/264/commits/15a54c11958c4ef5c440ad70efe9fd694ab2de11

api/apps/api/src/modules/scenarios/scenarios.controller.ts
kgajowy: Thanks for catching it, fixed in https://github.com/Vizzuality/marxan-cloud/pull/264/commits/78de3b1034837a70b3563efe89283787a28f90ad

hotzevzl (approved):

api/apps/api/src/modules/scenarios/scenarios.controller.ts
hotzevzl: @kgajowy should we remove in a new PR the `best` and `most-different` queries and relative code paths in this controller, or did you want to leave these in to let API clients query for best and most different solutions in either way? (I am fine either way, though I'd prefer to have only one way to query for best and most efficient, via their own endpoint).

PR #270, opened 2021-06-15
fix a few links in the architecture chart

hotzevzl:


hotzevzl:
trivial plantuml change: I am going to merge directly 😬

PR #273, opened 2021-06-15
feat(scenario): validate marxan input

kgajowy:
Marxan Input parameters are separated on purpose from the dto itself:
* (future) difference in dto shape
* allow to validate regardless of the transport layer (and validation pipe flag in our case as well)

image

### DIFF
--- a/api/apps/api/test/project-scenarios.e2e-spec.ts
+++ b/api/apps/api/test/project-scenarios.e2e-spec.ts
@@ -136,5 +136,28 @@ describe('ScenariosModule (e2e)', () => {
 
       expect(resources).toBeUndefined();
     });
+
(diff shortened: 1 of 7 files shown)


api/apps/api/test/project-scenarios.e2e-spec.ts
```diff
@@ -136,5 +136,28 @@ describe('ScenariosModule (e2e)', () => {
 
       expect(resources).toBeUndefined();
     });
+
+    it('should not allow to create scenario with invalid marxan properties', async () => {
```
hotzevzl, line 140: Albeit less immediately useful, I would also ensure that if we supply valid `input.dat` metadata we can successfully create or update a scenario (may be useful to catch some gross regressions).

hotzevzl (approved):

PR #274, opened 2021-06-15
[MARXAN-415] Chore/user journeys

tamaramegan:
### Overview
I have mapped as many of the paths as I could come up with ( I am sure there are many more), but it would be handy for you to check that the language and the structure meets the needs of product testing.
I have separated the files into blocks according to the section where the actions are happening (Projects, Scenario, Features, Analysis, Solutions, etc..) but let me know if another structure is preferred (for example, uniting all the download actions into one file).
I guess the main thing is to review the type of actions (if they are too broad or too detailed) so I can get an idea of how to refine this for the next round.

### Feature relevant tickets
JIRA task: MARXAN-415
https://vizzuality.atlassian.net/browse/MARXAN-415?atlOrigin=eyJpIjoiY2UzYjYxZTYyMThhNDIwOGExNGQ1NDA3Njc4OTg3YzEiLCJwIjoiaiJ9

Thank you!


e2e-product-testing/cypress/integration/AdjustPlanningUnits.feature
```diff
@@ -0,0 +1,63 @@
+Feature: 'Exclude' planning units
```
alexeh, line 1: pro-tip for next time: When all the scenarios of a Feature shares the same 'pre-condition' step, you can use:

`Feature: xxxxx`
   Background: Given the user is in XXXX
      Scenario: AAA
      Sceario: BBB

kgajowy (commented):
@tamaramegan 

few key notes: 
* we want to start with most critical for business scenarios
* we don't need to focus on all edge cases, including "error" ones - happy path is perfectly fine (of course those existing could be used later on to test directly on front/backend in separation)
* keep attention to language "user has saved" "user has tapped save"
* scenario shouldn't describe too much, especially if it includes some "processing" (` When the features are processed (Continue)`)
* steps should be precise to measure (` Then the user gets a notifciation on screenF` what notification?)
* ` Then the user is sent to the Project dashboard to view the status of the Scenario` we should rather avoid declarations of purpose (scenario/feature should describe them)
* as Alex mentioned, using `Background` would be great 
* we need to aware that some steps may not be technically possible to check from user perspective (`Then an email is sent to the email provided` - it is out of our web application at all)
*  some steps aren't clear (`When the user adds a contributor with a Marxan profile`)
* we could make use of `Examples:` notation for few cases (`Feature: Order solutions table` - but again, does this give value to business critical process?)
* each feature/scenario should describe a business value/feature (`Feature: Advanced settings` how this applies?)
*

alexeh:
@tamaramegan Congrats on your first fight with Gherkin!

There is room (as always) for some improvements from my PoV. I do agree with some of the points Kamil has pointed out about making steps (mostly the expected result or 'Then' ) and the use of more advanced features like `Examples` or `Scenario Outlines`

But I think we could keep the definition of possible edge-cases instead of just the happy path (We can decide then to automate the happy-path but I think having said edge-cases documented has a great value as part of Marxan's test-set battery)

Also we can update and improve some steps when it's time to automate this tests. Up to you if you want to follow Kamil's improvement suggestions

Thanks!

alexeh (approved):

hotzevzl:
@tamaramegan there's a lot of feedback on this PR, some of which can be guidance to further improve future iterations/new additions to the gherkin scripts, but please consider if you can/wish to incorporate some immediately actionable feedback.

From my point of view, small adjustments to clarify intent/what to check would be useful.

For example following Kamil's comments about:
* "precise to measure" descriptions
* attention to language describing user actions
* maybe narrowing down some scenarios

hotzevzl:
@tamaramegan at the same time, please don't feel you need to change much - this was meant to be a first iteration, and what you have added is great. I'm also ok to merge this with minimal changes or as is and leave improvements to later PRs.

Devs can also tweak these test scripts a bit when they feel there's no clear action that can be tested: I wanted to involve you in the drafting of the scenarios and tests because you have a thorough understanding of the user actions on the platform, but I don't want to be pushing detailed dev work on you.

PR #275, opened 2021-06-16
[MARXAN-462] input parameter file

Dyostiq:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/scenarios/input-parameter-file.provider.spec.ts
+++ b/api/apps/api/src/modules/scenarios/input-parameter-file.provider.spec.ts
@@ -1,0 +1,165 @@
+import { PromiseType } from 'utility-types';
+import { Test } from '@nestjs/testing';
+import { InputParameterFileProvider } from './input-parameter-file.provider';
+import { JobStatus, Scenario, ScenarioType } from './scenario.api.entity';
+import { ScenariosCrudService } from './scenarios-crud.service';
+
+jest.mock('config', () => ({
+  get: () => 'value',
+  has: () => true,
+}));
+
+let fixtures: PromiseType<ReturnType<typeof getFixtures>>;
+let sut: InputParameterFileProvider;
+
+beforeEach(async () => {
+  fixtures = await getFixtures();
+  sut = fixtures.getInputParameterFileProvider();
+});
+
+describe(`when a full scenario available`, () => {
+  let parameterFile: string;
+  beforeEach(async () => {
+    // given
+    const scenarioId = await fixtures.hasInDb({
+      ...fixtures.scenario(),
+      ...fixtures.withBoundaryLengthModifier(),
+      ...fixtures.withNumberOfRuns(),
+      ...fixtures.withInputParameters(),
+    });
+
+    // when
+    parameterFile = await sut.getInputParameterFile(scenarioId);
+  });
+
+  // then
+  it(`should return a string with valid format`, () => {
+    expect(parameterFile).toEqual(`BLM 0
+NUMREPS 100
+MARXANRUNKEY1 value1
+MARXANRUNKEY2 value2
+MARXANRUNKEY3 3`);
+  });
+});
+
+describe(`when a full scenario with duplicated BLM & NUMREPS available`, () => {
+  let parameterFile: string;
+  beforeEach(async () => {
+    // given
+    const scenarioId = await fixtures.hasInDb({
+      ...fixtures.scenario(),
+      ...fixtures.withBoundaryLengthModifier(),
+      ...fixtures.withNumberOfRuns(),
+      ...fixtures.withInputParameters(),
+    });
+
+    // when
+    parameterFile = await sut.getInputParameterFile(scenarioId);
+  });
+
+  // then
+  it(`should return a file without duplications`, () => {
+    expect(parameterFile).toEqual(`BLM 0
+NUMREPS 100
+MARXANRUNKEY1 value1
+MARXANRUNKEY2 value2
+MARXANRUNKEY3 3`);
+  });
+});
+describe(`when a scenario without parameters`, () => {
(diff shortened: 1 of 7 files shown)


api/apps/api/src/modules/scenarios/input-parameter-file.provider.spec.ts
```diff
@@ -0,0 +1,92 @@
+import { PromiseType } from 'utility-types';
+import { Test } from '@nestjs/testing';
+import { InputParameterFileProvider } from './input-parameter-file.provider';
+import { JobStatus, Scenario, ScenarioType } from './scenario.api.entity';
+import { ScenariosCrudService } from './scenarios-crud.service';
+
+jest.mock('config', () => ({
+  get: () => 'value',
+  has: () => true,
+}));
```
Dyostiq, line 10: we need to decouple the code from the config

hotzevzl (approved):

api/apps/api/src/modules/scenarios/input-parameter-file.provider.ts
```diff
@@ -0,0 +1,42 @@
...
+import { ScenariosCrudService } from './scenarios-crud.service';
+
+class InputParameterFile {
+  constructor(
+    private readonly boundaryLengthModifier?: number,
+    private readonly numberOfRuns?: number,
+    private readonly marxanInputParameterFile?: Record<string, string | number>,
+  ) {}
+
+  toString(): string {
```
kgajowy, line 12: if it shall serve the real `input.dat` file, it probably needs headers:

```
Input file for Annealing program.

This file generated by ArcMarxan Toolbox version 1.0.3
created by Apropos Information Systems Inc.

General Parameters
VERSION 0.1
BLM 1.00000000000000E+0000
PROP 5.00000000000000E-0001
RANDSEED -1
NUMREPS 100

Annealing Parameters
NUMITNS 1000000
STARTTEMP -1.00000000000000E+0000
COOLFAC -1.00000000000000E+0000
NUMTEMP 10000

Cost Threshold
COSTTHRESH 0.00000000000000E+0000
THRESHPEN1 0.00000000000000E+0000
THRESHPEN2 0.00000000000000E+0000

Input Files
INPUTDIR input
SPECNAME spec.dat
PUNAME pu.dat
PUVSPRNAME puvsp.dat
BOUNDNAME bound.dat
MATRIXSPORDERNAME puvsp_sporder.dat

Save Files
SCENNAME output
SAVERUN 3
SAVEBEST 3
SAVESUMMARY 3
SAVESCEN 3
SAVETARGMET 3
SAVESUMSOLN 3
SAVELOG 3
SAVESNAPSTEPS 0
SAVESNAPCHANGES 0
SAVESNAPFREQUENCY 0
OUTPUTDIR output

Program control.
RUNMODE 1
MISSLEVEL 9.50000000000000E-0001
ITIMPTYPE 1
HEURTYPE -1
CLUMPTYPE 0
VERBOSITY 2
SAVESOLUTIONSMATRIX 3
```

not sure tho, @aagm ?

api/apps/api/src/modules/scenarios/input-parameter-file.provider.ts
```diff
@@ -0,0 +1,42 @@
...
+    let builder = '';
+    if (isDefined(this.boundaryLengthModifier)) {
+      builder += `BLM ${this.boundaryLengthModifier}\n`;
+    }
+    if (isDefined(this.numberOfRuns)) {
+      builder += `NUMREPS ${this.numberOfRuns}\n`;
+    }
+    if (isDefined(this.marxanInputParameterFile)) {
+      builder += Object.entries(this.marxanInputParameterFile)
```
kgajowy, line 21: AFAIR the NUMREPS and BLM are already within the `marxanInputParameterFile` as well, so it would produce this parameter twice. Not sure tho if this would break the marxan run, @aagm ?

api/apps/api/src/modules/scenarios/input-parameter-file.provider.ts
hotzevzl: there's no need for anything else than the known `KEY value` pairs - the marxan cli iterates over the lines of `input.dat` and extracts the known `KEY value` pairs - everything else is ignored.

api/apps/api/src/modules/scenarios/input-parameter-file.provider.ts
hotzevzl: well spotted... IIRC from the Marxan source code, duplicated `KEY value` pairs will result in the first one to be found to be used.

If we don't want to enforce a schema for `marxanInputParameterFile` (and hence outright reject keys such as `BLM` and `NUMREPS`), then we should filter out entries with both these keys here.

Dyostiq: @hotzevzl what value should BLM have when `boundaryLengthModifier` is null? Should it be taken from `marxanInputParameterFile` or omitted entirely?

hotzevzl: @Dyostiq it is not required according to the Marxan manual and if not supplied it will default to 1.

We should not even allow `BLM` to be in `marxanInputParameterFile`: the frontend will not send it, but if a generic API clients tries to do so as part of a scenario create/update, we should reject the request.

Relatedly, we could either add known keys with default values if they are not in `marxanInputParameterFile` but I would avoid to do so - but cross checking this in Slack in a moment...

Dyostiq: fixed in 2c76cec1

kgajowy (approved):

hotzevzl (approved):

PR #278, opened 2021-06-16
feat(api): add io settings to input parameter file

Dyostiq:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/config/custom-environment-variables.json
+++ b/api/apps/api/config/custom-environment-variables.json
@@ -32,5 +32,19 @@
     "sharedFileStorage": {
       "localPath": "API_SHARED_FILE_STORAGE_LOCAL_PATH"
     }
+  },
+  "marxan": {
+    "inputFiles": {
+      "inputDat": {
+        "ioSettings": {
+          "INPUTDIR": "MARXAN_IO_SETTING_INPUTDIR",
+          "PUNAME": "MARXAN_IO_SETTING_PUNAME",
+          "SPECNAME": "MARXAN_IO_SETTING_SPECNAME",
+          "PUVSPRNAME": "MARXAN_IO_SETTING_PUVSPRNAME",
+          "BOUNDNAME": "MARXAN_IO_SETTING_BOUNDNAME",
+          "OUTPUTDIR": "MARXAN_IO_SETTING_OUTPUTDIR"
+        }
+      }
+    }
   }
 }
--- a/api/apps/api/config/default.json
+++ b/api/apps/api/config/default.json
@@ -31,6 +31,20 @@
     "sharedFileStorage": {
       "localPath": "/tmp/storage"
     }
+  },
+  "marxan": {
+    "inputFiles": {
+      "inputDat": {
+        "ioSettings": {
+          "INPUTDIR": "input",
+          "PUNAME": "pu.dat",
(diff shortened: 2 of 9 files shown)


api/apps/api/config/custom-environment-variables.json
```diff
@@ -32,5 +32,19 @@
     "sharedFileStorage": {
       "localPath": "API_SHARED_FILE_STORAGE_LOCAL_PATH"
     }
+  },
+  "marxan": {
+    "inputFiles": {
+      "inputDat": {
+        "ioSettings": {
+          "NPUTDIR": "MARXAN_IO_SETTING_NPUTDIR",
```
hotzevzl, line 40:
```suggestion
          "INPUTDIR": "MARXAN_IO_SETTING_INPUTDIR",
```

Dyostiq: thanks, fixed :)

api/apps/api/config/custom-environment-variables.json
```diff
@@ -32,5 +32,19 @@
...
+    "inputFiles": {
+      "inputDat": {
+        "ioSettings": {
+          "INPUTDIR": "MARXAN_IO_SETTING_INPUTDIR",
+          "PUNAME": "MARXAN_IO_SETTING_PUNAME",
+          "SPECNAME": "MARXAN_IO_SETTING_SPECNAME",
+          "PUVSPRNAME": "MARXAN_IO_SETTING_PUVSPRNAME",
+          "BOUNDNAME": "MARXAN_IO_SETTING_BOUNDNAME",
+          "OUTPUTDIR": "MARXAN_IO_SETTING_OUTPUTDIR"
+        }
```
kgajowy, line 46: INPUTDIR & OUTPUTDIR

not sure if those makes sense. Due to Marxan's CLI nature, we need to create directories it runs as temporary ones per run (and in fact, clean them up afterwards)

kgajowy: other files names also may not have any point, as we will be creating them manually as well (and no way to get them afterwards - unless we want to keep them ofc)

Dyostiq: I've taken the list from the task description
@hotzevzl could you address these doubts?

hotzevzl: @kgajowy my plan was to create per-run root directories on fs and then use the same names for folders and files within each per-run root - so all these settings should be relative to each root. I would not use the marxan settings to point the CLI to the full path of input and output folders.
As for creating these config settings, they could even be hardcoded constants in the application source, but we will also in fact allow users to download the input files (https://vizzuality.atlassian.net/browse/MARXAN-308), so in case instance admins want to expose these files with non-standard names to users, then they could do so by fiddling with config rather than with the application source.

kgajowy: @hotzevzl
ah, I missed they are relative. All fine!
Good luck, thats what I did in https://github.com/Vizzuality/marxan-cloud/pull/277 as well (per-run directories).

hotzevzl (approved):

kgajowy (approved):

PR #281, opened 2021-06-18
Fix/api/marxan 479 fixed size calc

aagm:
fix the calculation for min / max adding a conversion factor


hotzevzl (commented):
@aagm I am not sure about this approach - we are calculating the area via generated columns, and then basically overwriting these calculated numbers here, in ~70LOC vs... 2 if we did this in SQL? 🤔

Besides the LOC count which is not a real issue (most of it is repeated anyway), I find it also confusing to be recalculating some props in the extendResults hooks - since we can control what comes from these properties to start with, I would do that: extendResults in my opinion makes more sense for either reshaping or calculating data we don't have control over, or to add *new* properties, and so on.

also, what is the concept we are dealing with? in the db we calculate the area in m², here we calculate the length of the side of the square with that area, in km, but we use the same prop/name.

tl;dr - why not calculate both the m² area and the square side length (there should be a better name for this, but my neurons need lunch 😱), and adding these two new props to the entity and json:api serialization config?

hotzevzl:
> tl;dr - why not calculate both the m² area and the square side length (there should be a better name for this, but my neurons need lunch scream), and adding these two new props to the entity and json:api serialization config?

⬆️ I meant calculating everything as generated columns

aagm:
I was probably overengineering there and didn't want to touch the DB :facepalm: but as you said it should be the best approach

PR #282, opened 2021-06-18
feat(protected-areas): fundamentals for gc for outdated geometries

kgajowy:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


api/apps/geoprocessing/src/modules/planning-area/gc/config/config.provider.ts
```diff
@@ -0,0 +1,29 @@
+import cron from 'cron-validate';
+import { GcSettings } from './gc-settings';
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const ms = require('ms');
```
Dyostiq, line 5: maybe we should turn on the esModuleInterop flag

Dyostiq (approved):

hotzevzl (commented):
Having thought =more about the problem we're trying to address here, my sense is that we're venturing a bit in YAGNI territory here - from my point of view, firing a single `delete from planning_areas where <age condition>` each time a new geometry is uploaded is more than enough for the problem at hand.

We don't even need to wait for this operation to complete - as long as the query is executing, and it should take in the order of 100s of ms in most cases, if not less, with a proper index - we are ok.

The only downside I see is if a user uploads say hundreds of large geometries within a few minutes, then nobody else uploads anything for > 2days (or whatever is the GC schedule set on an instance) - we end up holding all that stale data for longer. But apart from wasted storage, as long as SELECTs for actual geometries linked to projects do happen with an index lookup, this wouldn't have more negative effects.

Is there any other downside I may not be considering?

hotzevzl:
another issue I see with the proposed cron-based implementation is that containers cannot be fully stateless anymore. This is not so much of an issue for short GC schedules, but even if scheduling GC events (or any event, for that matter) at larger intervals (for some definition of "larger": I would say from a few hours and up), we should assume that containers may be randomly restarted and so "miss" their schedules.

Given this and my earlier note, I would definitely recommend to go for a "trigger GC on each upload" approach.

Dyostiq:
@hotzevzl, I've implemented it with GC-on-save approach

api/package.json
```diff
@@ -91,13 +91,14 @@
     "@types/bcrypt": "^3.0.0",
     "@types/config": "^0.0.38",
     "@types/cors": "^2.8.9",
+    "@types/cron": "1.7.2",
```
kgajowy, line 94: no longer needed I believe (as well as nestjs scheduler?)

api/apps/geoprocessing/test/integration/planning-areas/custom-planning-area.repository.e2e-spec.ts
```diff
@@ -0,0 +1,164 @@
...
+    expect(bbox).toStrictEqual([
+      -30.146484374999996,
+      -40.341796875,
+      38.20365531807149,
+      29.22889003019423,
+    ]);
+  });
+});
+
+it(`should confirm deny existence of not existing entity `, async () => {
```
hotzevzl, line 51: just minor - i think the current test description has some extra leftover words, maybe it can be amended like this?

```suggestion
it(`should deny existence of unknown entity `, async () => {
```

api/apps/geoprocessing/src/modules/planning-area/planning-area.service.ts
```diff
@@ -1,23 +1,46 @@
...
       id: result[0].id,
       data,
     };
   }
+
+  private scheduleGarbageCollection() {
+    setTimeout(async () => {
+      this.logger.debug(`garbage collection scheduled`);
+      const removedCount = await this.collectGarbage();
+      this.logger.debug(
```
hotzevzl, line 33: very minor, but if we want to keep some logging output, I suggest to make this `.log()`, to pass in the max age value and to output this too (`...removed <count> entities older than <age>`).

We already have some time diff in the default logs to know how long the GC takes.

I would be tempted to suggest to add a `v4()` to the two log lines in case we have >1 GC ops triggered at the kind of exact same time (two users saving planning areas at the same time) but I'd say it's very unlikey and the utility we could get from knowing for sure which of the two (or more 😁) GC ops started and ended when it's very low.
Feel free to leave this as is, in any case - since we already have some output, I just suggested to make it more informative with age data, in case we need to check back anything in logs

hotzevzl (approved):
@Dyostiq looks good, thanks!
I would suggest to add a db index on `geodb.planning_areas (created_at)` and maybe on `geodb.planning_areas (project_id) where project_id is null` to optimize lookups of GC-eligible entities.

api/apps/geoprocessing/test/integration/planning-areas/custom-planning-area.repository.e2e-spec.ts
Dyostiq: yes, sure, fixed :) `confirm deny` 😅

Dyostiq:
@hotzevzl I'll add indices in another PR, ok? :)

PR #283, opened 2021-06-18
With BBox present there is no need to simplify.

aagm:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/geoprocessing/src/modules/planning-units/planning-units.service.ts
+++ b/api/apps/geoprocessing/src/modules/planning-units/planning-units.service.ts
@@ -110,14 +110,17 @@ export class PlanningUnitsService {
     const gridSize = this.calculateGridSize(planningUnitAreakm2);
     // 156412 references to m per pixel at z level 0 at the equator in EPSG:3857
     const ratioPixelExtent = gridSize / (156412 / 2 ** z);
-    let query = `( SELECT row_number() over() as id, (${gridShape}(${gridSize}, \
+    /**
+     * We are checking that the pixel ration is < 8 px and bbox is not present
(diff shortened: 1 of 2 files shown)


api/apps/geoprocessing/src/modules/planning-units/planning-units.service.ts
```diff
@@ -111,7 +114,7 @@ export class PlanningUnitsService {
                     ST_Transform(ST_TileEnvelope(${z}, ${x}, ${y}), 3857))).geom as the_geom)`;
     // (so we are checking that the pixel ration is < 8 px)
     // If so the shape we are getting is down the optimal to visualize it
-    if (ratioPixelExtent < 8) {
+    if (ratioPixelExtent < 8 && !filters?.bbox) {
```
hotzevzl, line 117: Looks good. I would add a tiny explanation for those like me who don't have much of a clue about what's going on here as to why we don't need to use the second query if the pixel ratio is  within the limit or if we have a bbox 😬 

also (but this is veeery much my preference for ternary operator), this could be written (more concisely in my opinion) as

```
const query = (ratioPixelExtent < 8 && !filters?.bbox) ?
  `( SELECT row_number() over() as id, st_centroid((${gridShape}(${gridSize}, \
        ST_Transform(ST_TileEnvelope(${z}, ${x}, ${y}), 3857))).geom ) as the_geom )` :
  `( SELECT row_number() over() as id, (${gridShape}(${gridSize}, \
                    ST_Transform(ST_TileEnvelope(${z}, ${x}, ${y}), 3857))).geom as the_geom)`;
```

hotzevzl (approved):

PR #285, opened 2021-06-18
add redis cache in mvt service

elpamart:
### Overview

Add redicscache to mvt service. We need to check the overwriting of task inside redis.

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/geoprocessing/src/modules/tile/tile.service.ts
+++ b/api/apps/geoprocessing/src/modules/tile/tile.service.ts
@@ -10,8 +10,7 @@ import { getConnection } from 'typeorm';
 import * as zlib from 'zlib';
 import { Transform } from 'class-transformer';
 import { IsInt, Max, Min } from 'class-validator';
-
-const Redis = require('ioredis');
+import * as Redis from 'ioredis';
 
 /**
  * @description The specification of the tile request
@@ -56,9 +55,6 @@ export interface TileCacheOptions {
     ttl?: number;
   };
 }
-
-export type TTtl = (zoomLevel: number) => number;
-
 
 /**
  * @description The required input values for the tile renderer
@@ -204,8 +200,8 @@ export class TileService {
    */
   async getTile(tileInput: TileInput<string>): Promise<Buffer> {
     try {
-      const {table, z, x, y , customQuery} = tileInput;
-      const filters = !!customQuery ? customQuery : [];
+      const {table, z, x, y , customQuery } = tileInput;
+      const filters = !!customQuery ? customQuery : '';
       const cacheOptions: TileCacheOptions = {
         enabled: true,
         redisOptions: {
--- a/api/apps/geoprocessing/src/modules/tile/tile.service.ts
+++ b/api/apps/geoprocessing/src/modules/tile/tile.service.ts
@@ -1,5 +1,5 @@
 // to-do: work on cache later
-import { RedisOptions } from 'ioredis'
+import { RedisOptions } from 'ioredis';
 import {
   BadRequestException,
   Injectable,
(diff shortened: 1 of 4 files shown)


Dyostiq (commented):
Hmm, maybe instead of implementing the cache by hand, we could utilize a built-in nest's support on the HTTP level?

api/apps/geoprocessing/src/modules/tile/tile.service.ts
```diff
@@ -38,6 +41,25 @@ export class TileRequest {
...
+   */
+  redisOptions?: RedisOptions & {
+    /**
+     * @description The time to live in seconds. Default is 86400 (1 day)
+     */
+    ttl?: number;
+  };
+}
+
+export type TTtl = (zoomLevel: number) => number;
```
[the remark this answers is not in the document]
elpamart: true! it was used before but I forgot to get rid of it.

api/apps/geoprocessing/src/modules/tile/tile.service.ts
```diff
@@ -181,18 +203,51 @@ export class TileService {
...
-    let data: Buffer = Buffer.from('');
     try {
-      const queryResult: Record<
-        'mvt',
-        Buffer
-      >[] = await this.fetchTileFromDatabase(tileInput);
-      // zip data
-      data = await this.zip(queryResult[0].mvt);
+      const {table, z, x, y , customQuery} = tileInput;
+      const filters = !!customQuery ? customQuery : [];
```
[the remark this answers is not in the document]
elpamart: true, I have defaulted the custom query to string.

api/apps/geoprocessing/src/modules/tile/tile.service.ts
```diff
@@ -10,6 +11,8 @@ import * as zlib from 'zlib';
 import { Transform } from 'class-transformer';
 import { IsInt, Max, Min } from 'class-validator';
 
+const Redis = require('ioredis');
```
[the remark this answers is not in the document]
elpamart: changed! thanks!

api/apps/geoprocessing/src/modules/tile/tile.service.ts
```diff
@@ -181,18 +199,51 @@ export class TileService {
...
+      const cacheOptions: TileCacheOptions = {
+        enabled: true,
+        redisOptions: {
+          ttl: 86400, // 24 hours
+          host: process.env.REDIS_HOST,
+        },
+      };
+
+      const  redisCache = new Redis(cacheOptions.redisOptions);
+      const cacheKey = `${table}-${z}-${x}-${y}-${filters}`
```
hotzevzl, line 214: This initial block of code is all that we are wrapping in the outermost `try` block - I think we can remove this `try/catch` outright and if anything happens here, let the global exception filter deal with it.

I would even go a bit further and avoid any nested `try/catch` in this function altogether: I think we can get away with something like this (totally untested!)

```typescript
async getTile(tileInput: TileInput<string>): Promise<Buffer> {
    const { table, z, x, y, customQuery } = tileInput;
    const filters = !!customQuery ? customQuery : [];
    const cacheOptions: TileCacheOptions = {
      enabled: true,
      redisOptions: {
        ttl: 86400, // 24 hours
        host: process.env.REDIS_HOST,
      },
    };

    const redisCache = new Redis(cacheOptions.redisOptions);
    const cacheKey = `${table}-${z}-${x}-${y}-${filters}`;

    const tileFromCache = await redisCache
      .getBuffer(cacheKey)
      .catch((error) => {
        // In case the cache get fail, we continue to generate the tile
        this.logger.error(`Cache failed: ${error.message}`);
      });

    if (tileFromCache) return tileFromCache;

    return await this.fetchTileFromDatabase(tileInput)
      .then(async (result) => {
        return await this.zip(result[0].mvt);
      })
      .then(async (result) => {
        redisCache.set(cacheKey, result, 'EX', cacheOptions.redisOptions?.ttl);

        return result;
      })
      .catch((error) => {
        this.logger.error(`Database error: ${error.message}`);
        throw new Error(error.message);
      });
  }
```

hotzevzl (commented):
@elpamart thanks! looks, good, just see comments on avoiding nested try/catch blocks (feel free to keep try/catch, or to adapt my try/catch-less version, but in any case I'd try to avoid unneeded ones and above all to nest them, as it's easier to follow a more linear flow without nested blocks.

api/apps/geoprocessing/src/modules/tile/tile.service.ts
hotzevzl: also, don't forget to run `yarn format` before a final push

hotzevzl: and `yarn add -D @types/ioredis` would be useful for development too

elpamart: everything done! thanks for your help!

hotzevzl: @elpamart does this still work? 😅 
I see you've taken up most of my suggestion for this function, but be aware that I had only checked that it made sense from a typing point of view and *very roughly* logic-wise, but I didn't check that it would do what it's supposed to do 😱 

Feel free to merge if you've checked that things work the way they did with your initial implementation, but I just wanted to clarify that I didn't check that my suggested implementation *actually worked* (nor have I done so now on the latest commit in this PR) 😉

hotzevzl (approved):

hotzevzl:
also, one of the tests in the e2e suite is failing, but I think this is unrelated with the changes in this PR - if you could have a look that would be helpful but I think we'll have to address that separately (assuming it's not a flaky failure).

aagm:
Let me close this for nw

PR #288, opened 2021-06-22
[MARXAN-482, MARXAN-483] geo features: POST and PUT interface

hotzevzl:
see https://vizzuality.atlassian.net/browse/MARXAN-182


api/apps/api/src/modules/geo-features/geo-features.service.ts
```diff
@@ -138,19 +141,48 @@ export class GeoFeaturesService extends AppBaseService<
...
+        .where(
+          `st_intersects(
+        st_makeenvelope(:xmin, :ymin, :xmax, :ymax, 4326),
+        "geoFeatureGeometries".the_geom
+      )`,
+          {
+            xmin: project.bbox[0],
+            ymin: project.bbox[1],
+            xmax: project.bbox[2],
+            ymax: project.bbox[3],
```
hotzevzl, line 173: @aagm I think I got this wrong and this should be instead as below - could you sanity-check this for me, please?

```suggestion
            xmin: project.bbox[1],
            ymin: project.bbox[3],
            xmax: project.bbox[0],
            ymax: project.bbox[2],
```

api/apps/api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -130,6 +137,28 @@ export class ScenariosController {
     );
   }
 
+  @ApiOperation({ description: 'Create feature set for scenario' })
+  @ApiCreatedResponse({ type: ScenarioResult })
+  @Post(':id/features/specification')
+  async createFeatureSetFor(
+    @Body(new ValidationPipe()) dto: CreateGeoFeatureSetDTO,
```
hotzevzl, line 144: @Dyostiq I've seen you giving advanced suggestions for Swagger documentation - this is a WIP, but your feedback would be appreciated: here and in the following endpoint we use DTOs that themselves use union types with class-transformer discriminators for nested validation, which from what I understand and see don't play well with the kind of introspectioni the Swagger plugin could do... short of not using union types and adding some duplication, would you have suggestions for the Swagger documentation to reflect the full shape of DTOs?

api/apps/api/src/modules/geo-features/geo-features.service.ts
```diff
@@ -138,19 +141,48 @@ export class GeoFeaturesService extends AppBaseService<
...
+        .then((result) => result.map((i) => i.id))
+        .catch((error) => {
+          throw new Error(error);
+        });
+
       queryFilteredByPublicOrProjectSpecificFeatures = query.andWhere(
-        `${this.alias}.projectId = :projectId OR ${this.alias}.projectId IS NULL`,
-        { projectId },
+        `${this.alias}.projectId = :projectId OR ${this.alias}.projectId IS NULL
+        AND ${this.alias}.id IN (:...geoFeaturesWithinProjectBbox)`,
```
hotzevzl, line 184: @aagm I am not sure about this narrowing clause - see above where `geoFeaturesWithinProjectBbox` comes from and how it's computed (not sure about that query either). The intent is to limit our selection to features that somehow intersect the project planning area's bbox (when testing with a random Rwanda project I got ~100K rows out of 1.7M) which in turn should help with the sp-far-quite-expensive view that extracts properties (`geodb.feature_properties`).

Feedback welcome on any of this ❤️

api/apps/api/src/modules/scenarios/scenarios.controller.ts
Dyostiq: Hmm, for unions/polymorphic type you may use `oneOf` (along with `getSchemaPath`) and `discriminator` property of `ApiProperty`

Dyostiq: (and `ApiExtraModels` in order to have the model referenced by `getSchemaPath` included in the open api doc)

api/apps/api/src/modules/geo-features/geo-features.service.ts
hotzevzl: @aagm nevermind, this will never work with real data:

```
status: 500
title: "bind message has 5009 parameter formats but 0 parameters"
```

I have removed the `AND` part of the clause for the moment. I am thinking about how to deal with this spatial "join" across the two dbs without blowing things up at any scale - maybe we could set the bbox of each feature in the apidb at import time, and move the `st_intersects()` there (in a view?), but I am not too keen on this either...

PR #290, opened 2021-06-24
Feature/api/marxan 480 scenario pu vl

aagm:
## Retrieve back vector tiles from the planning units under a scenario
Retrives:
Protected Area, lockin status and cost.

Piece missing: features and Outputs

- [ ] Update CHANGELOG file


api/apps/api/test/proxy.vector-tiles.e2e-spec.ts
```diff
@@ -100,7 +105,11 @@ describe('ProxyVectorTilesModule (e2e)', () => {
      * https://www.figma.com/file/hq0BZNB9fzyFSbEUgQIHdK/Marxan-Visual_V02?node-id=2991%3A2492
      */
     describe('Admin-areas layers', () => {
-      test.todo('we should test that the response is a valid mvt');
+
+      test.todo('We should test that the response is a valid mvt');
+      test.todo('We should add test for user upload planning units');
+      test.todo('We should add test for irregular planning units');
```
hotzevzl, line 111: nothing to do here, but you could even write the actual test statement (the "we should test..." part is already implicit in the `.todo()` part.

api/apps/api/test/proxy.vector-tiles.e2e-spec.ts
```diff
@@ -109,78 +118,98 @@ describe('ProxyVectorTilesModule (e2e)', () => {
           .expect(HttpStatus.OK);
       });
       describe('Filter by guid', () => {
-        test.skip('guid country level', async () => {
+        test('guid country level', async () => {
```
hotzevzl, line 121: very minor, but I suggest to try to stick to test statements that describe what should (or should not) happen. BDD/Gherkin style is my personal favourite because of clarity and of being widely known, but in any case, something less cryptic than the current statement would be welcome 😉 

same for the l1 and l2 tests below. the "guid level > 2 should not be allowed" one is clear enough, instead.

and the `.description()` call just above this comment could be improved too (e.g. even just "When filtering by guid")...

api/apps/api/test/proxy.vector-tiles.e2e-spec.ts
```diff
@@ -109,78 +118,98 @@ describe('ProxyVectorTilesModule (e2e)', () => {
...
         const response = await request(app.getHttpServer())
-          .get('/api/v1/administrative-areas/1/preview/tiles/100/60/30.mvt')
+          .get('/api/v1/administrative-areas/0/preview/tiles/21/30/25.mvt')
           .set('Authorization', `Bearer ${jwtToken}`)
           .expect(HttpStatus.BAD_REQUEST);
       });
-
-      test('Should throw a 400 error if filtering by level other than 0, 1 or 2', async () => {
+    });
+    describe('WDPA preview vector tiles.', () => {
```
hotzevzl, line 159: not for this PR, but I think it's worth splitting up these tests according to the kind of tiles we test, similarly to what we did for the actual controllers.

api/apps/geoprocessing/src/modules/scenarios/scenarios.service.ts
```diff
@@ -0,0 +1,78 @@
...
+  private readonly logger: Logger = new Logger(ScenariosService.name);
+  constructor(
+    @InjectRepository(ScenariosPuPaDataGeo)
+    private readonly ScenariosPlanningUnitGeoEntityRepository: Repository<ScenariosPuPaDataGeo>,
+    @Inject(TileService)
+    private readonly tileService: TileService,
+  ) {}
+
+  /**
+   * @param filters bounding box of the area where the grids would be generated
```
hotzevzl, line 33: maybe add a `@todo` note if this needs to be implemented later on (or outright delete the method if we won't need it)

api/apps/geoprocessing/test/index.html
```diff
@@ -42,58 +42,67 @@
...
-            },
-            'postgis-tileswdpa': {
-              type: 'vector',
-              minzoom: 1,
-              maxzoom: 12,
-              tiles: [
-                'http://localhost:3040/api/v1/protected-areas/preview/tiles/{z}/{x}/{y}.mvt',
-              ],
-            },
+            // 'postgis-tiles0': {
```
hotzevzl, line 45: there's a lot of commented-out source in this file - can we remove it?
if it's not used only temporarily, maybe these sources could be moved temporarily to a `skippedSources` prop or something like that.

but it's a test file, so feel free to leave it as is, of course 😅

api/libs/scenarios-planning-unit/src/scenarios-pu-pa-data.geo.entity.ts
```diff
@@ -0,0 +1,43 @@
...
+  protectedArea!: number;
+
+  @OneToOne(() => PlanningUnitsGeom, PlanningUnitsGeom => PlanningUnitsGeom.id, {
+    onDelete: 'CASCADE',
+  })
+  @JoinColumn({
+    name: 'pu_geom_id',
+    referencedColumnName: 'id',
+  })
+  planning?: PlanningUnitsGeom | null;
```
hotzevzl, line 32: I suggest to make this property name more descriptive - this is referencing a full planning unit geometry, right? if so, what about `planningUnitGeom`?

hotzevzl: I would also suggest to add the cascades in the actual table definitions via a migration (I think they're not there)

api/libs/scenarios-planning-unit/src/scenarios-pu-pa-data.geo.entity.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  @OneToOne(() => ScenariosPuCostDataGeo, ScenariosPuCostDataGeo => ScenariosPuCostDataGeo.id, {
+    onDelete: 'CASCADE',
+  })
+  @JoinColumn({
+    name: 'pu_geom_id',
+    referencedColumnName: 'id',
+  })
+  cost?: PlanningUnitsGeom | null;
```
hotzevzl, line 41: same as above - this is not really a `cost`, but something like `costData` 😬 , and same as above for the cascade (I think).

Not sure if a view could maybe also help with all these joins, by giving us a denormalized but immediately-usable view entity (not for this PR unless you think this is possible, beneficial *and* easy to get right 😅 )

hotzevzl (approved):
@aagm thanks! looks good overall, just see if any of the comments make sense/need action from your point of view.

api/apps/api/test/proxy.vector-tiles.e2e-spec.ts
aagm: Well indeed the testing is very similar, check that the tiles are being generated as expected. If this files grows a lot i will do it ofc

api/apps/geoprocessing/src/modules/scenarios/scenarios.service.ts
aagm: i did deleted as under scenarios we should no longer filter by bbox as the data should be only refer to the Planning area location so, i will delete it for good

api/apps/geoprocessing/test/index.html
aagm: yes is our sandbox for map testing, so ideally each layer and style represents one of our designed tiles endpoints, and in this specific I was avoiding noise.

api/libs/scenarios-planning-unit/src/scenarios-pu-pa-data.geo.entity.ts
aagm: i was avoiding a view at this stage as the idea is to plugin this joins as needed and requested from the FE

PR #292, opened 2021-06-25
feat(api): assign projects to planning areas

Dyostiq:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


api/libs/planning-area-repository/src/custom-planning-area.repository.ts
```diff
@@ -43,4 +43,15 @@ INSERT INTO "planning_areas"("the_geom")
       | undefined = await this.planningAreas.findOne(id, { select: ['id'] });
     return isDefined(selectedId?.id);
   }
+
+  async assignProject(id: string, projectId: string): Promise<void> {
+    await this.planningAreas.update(
```
kgajowy, line 48: Is this correct that given planning area can only be assigned to single project id? what if planning area is a "generic one"?

Dyostiq: hmm, good question :D also, should we unassign it if project's planning area has changed?

@hotzevzl I think you may know the answer :D

hotzevzl: yes, an *uploaded* planning area (as in those stored in the `planning_areas` db table) can only be associated to a single project (at most). if users upload the same identical file for a different project, we create a new row and link the new one to the second project, even if they contain the exact same geometry

In terms of unassigning, if a user uploads a new shapefile, then yes, the latest will need to be linked to the project, after unlinking the previous one (if any) - did I understand your question correctly?

this is a "last to finish wins" logic with all its potential caveats, but we're fine with this at this stage.

Dyostiq: Yes, thank you.

So, after being unlinked, the planning area also should be garbage collected during the next upload.

kgajowy: Keeping in mind we have some constraints (AFAIR) that there can be only one geometry of given kind (thus - planning areas are unique), if it happens someone uses the very same shape, which wouldn't be uncommon), with this approach, we probably introduce potential serious issue 🤔 
@hotzevzl @Dyostiq

kgajowy: ah sorry, thought its `planning_units_geom` 😅

kgajowy (approved):

PR #298, opened 2021-06-30
refactor(api): move from purify-ts to fp-ts, fix swagger docs

Dyostiq:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


kgajowy (approved):

api/apps/api/src/modules/projects/dto/planning-area-created.dto.ts
```diff
@@ -0,0 +1,14 @@
...
+  @ApiProperty({
+    description: 'An ID of the created planning area',
+  })
+  id!: string;
+
+  @ApiProperty({
+    description: 'GeoJSON of the created planning area',
+  })
+  data!: GeoJSON;
+}
```
hotzevzl, line 14: very minor and subtle note - we already have a `ShapefileGeoJSONResponseDTO` which may partly overlap with this at least in intention, though that's a generic GeoJSON response DTO. Can we keep the two linked (e.g. by extending the generic one)?

Also, is this DTO unique to "created"-type responses? or is it a generic planning area result DTO, even if only used in "created"-type responses? If so, I would stick to something like `PlanningAreaResponseDTO` or something like this.

api/apps/api/src/modules/projects/projects.controller.ts
```diff
@@ -206,27 +209,29 @@ export class ProjectsController {
     return;
   }
 
-  @ApiConsumesShapefile(true)
+  @ApiOkResponse({ type: PlanningAreaCreatedDto })
   @ApiOperation({
     description: 'Upload shapefile with project planning-area',
   })
+  @ApiConsumesShapefile(false)
```
hotzevzl, line 216: I had missed this earlier on - minor and can be for another PR if you think it's worth tweaking it: I suggest to use a config object for this decorator, such as `{ withGeoJsonResponse?: boolean }` defaulting to true, and maybe also adding the option to pass `type` as parameter (to avoid having a two ways of decorating this with `@ApiOkResponse()`.
But also because `@ApiConsumesShapefile(false)` is misleading until one looks at the source code of the decorator.
Again, this doesn't have much to do with the current PR but I'm just noting this because it seems a sensible and small tidy-up that can be done separately if you think it's worthy.

api/package.json
```diff
@@ -57,6 +57,7 @@
     "config": "^3.3.3",
     "express": "^4.17.1",
     "faker": "^5.1.0",
+    "fp-ts": "^2.10.5",
```
hotzevzl, line 60: do we still need `purify-ts` in `package.json` and `yarn.lock`?

hotzevzl (approved):

api/package.json
Dyostiq: hmm, it was added by mistake while adding the geojson validator, I'll remove it.

Dyostiq: https://github.com/Vizzuality/marxan-cloud/pull/298/commits/84b4e9453b6c140d631cb3c782904d27ebf2bb9c

api/apps/api/src/modules/projects/dto/planning-area-created.dto.ts
Dyostiq: ed1f455678144f55728ae940863249a58ca51cfa

api/apps/api/src/modules/projects/projects.controller.ts
Dyostiq: ed1f455678144f55728ae940863249a58ca51cfa

PR #299, opened 2021-06-30
[MARXAN-446] planning area assigning

Dyostiq:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/libs/planning-area-repository/src/custom-planning-area.repository.ts
+++ b/api/libs/planning-area-repository/src/custom-planning-area.repository.ts
@@ -53,6 +53,14 @@ INSERT INTO "planning_areas"("the_geom")
   async assignProject(id: string, projectId: string): Promise<void> {
     await this.planningAreas.update(
       {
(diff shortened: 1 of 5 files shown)


kgajowy (approved):

api/libs/planning-area-repository/src/custom-planning-area.repository.ts
```diff
@@ -51,6 +51,14 @@ INSERT INTO "planning_areas"("the_geom")
   }
 
   async assignProject(id: string, projectId: string): Promise<void> {
+    await this.planningAreas.update(
```
hotzevzl, line 54: in practice we could even delete this planning area straight away rather than orphaning it and letting it be GC'ed later on (but we'd need to do this in a transaction to make sure the delete can be rolled back if the successive update fails

hotzevzl (approved):

PR #300, opened 2021-06-30
feat(scenarios): allow to get input files via x-api-key

kgajowy:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/scenarios/marxan-run/marxan-run.controller.ts
+++ b/api/apps/api/src/modules/scenarios/marxan-run/marxan-run.controller.ts
@@ -1,0 +1,48 @@
+import {
+  Controller,
+  Get,
+  Header,
+  Param,
+  ParseUUIDPipe,
+  Res,
+  UseGuards,
+} from '@nestjs/common';
+import { ApiOkResponse, ApiOperation, ApiProduces } from '@nestjs/swagger';
+import { Response } from 'express';
+import { apiGlobalPrefixes } from '@marxan-api/api.config';
+import { XApiGuard } from '@marxan-api/guards/x-api.guard';
+import { MarxanRunService } from './marxan-run.service';
+
+@UseGuards(XApiGuard)
+@Controller(`${apiGlobalPrefixes.v1}/marxan-run/scenarios`)
+export class MarxanRunController {
+  constructor(private readonly service: MarxanRunService) {}
+
+  @Header('Content-Type', 'text/csv')
+  @ApiOkResponse({
+    schema: {
+      type: 'string',
+    },
+  })
+  @ApiOperation({
+    description: `Uploaded cost surface data`,
+  })
+  @Get(`:id/marxan/dat/pu.dat`)
+  async getScenarioCostSurface(
+    @Param('id', ParseUUIDPipe) id: string,
+    @Res() res: Response,
+  ): Promise<void> {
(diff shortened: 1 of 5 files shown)


api/apps/api/test/scenarios/input-file.e2e-spec.ts
```diff
@@ -203,6 +230,16 @@ async function getFixtures() {
         .set('Authorization', `Bearer ${jwtToken}`)
         .send(input);
     },
+    async ThenScenarioHasMetadataForMarxan(id: string) {
```
kgajowy, line 233: Will probably move later on to https://github.com/Vizzuality/marxan-cloud/pull/306

api/apps/api/test/scenarios/input-file.e2e-spec.ts
```diff
@@ -203,6 +230,16 @@ async function getFixtures() {
...
+      return await request(app.getHttpServer())
+        .get(`/api/v1/marxan-run/scenarios/${id}/marxan/dat/input.dat`)
+        .set(
+          'X-Api-Key',
+          process.env.API_AUTH_X_API_KEY ?? 'sure it is valid in envs?',
```
hotzevzl, line 238: can we get this via `config`?

hotzevzl: so that we only couple setting to env var in `custom-environment-variables.json`

hotzevzl (approved):

api/apps/api/test/scenarios/input-file.e2e-spec.ts
kgajowy: not sure if this is desirable from e2e point of view, no hard opinion tho.

@Dyostiq your take?

hotzevzl: I have no hard opinion either - I would say that having a way to refer to settings consistently throughout may be desirable, but if the coupling was updated and we forgot to update the env var here the test would fail so we'd notice (but then extra tedious work 😅)

Dyostiq: if it's e2e, by using env we test a bit more, like if the mapping is correct etc.
I try to have e2e tests as much separate from the production code and tools as possible – then, they're still valid after refactor.

api/apps/api/src/modules/scenarios/marxan-run/marxan-run.controller.ts
```diff
@@ -0,0 +1,48 @@
...
+  Res,
+  UseGuards,
+} from '@nestjs/common';
+import { ApiOkResponse, ApiOperation, ApiProduces } from '@nestjs/swagger';
+import { Response } from 'express';
+import { apiGlobalPrefixes } from '@marxan-api/api.config';
+import { XApiGuard } from '@marxan-api/guards/x-api.guard';
+import { MarxanRunService } from './marxan-run.service';
+
+@UseGuards(XApiGuard)
```
Dyostiq, line 16: btw. (a note outside the scope of the PR)
x- prefixes are deprecated https://datatracker.ietf.org/doc/html/rfc6648
also for api-key we may use standard `Authorization` header

PR #303, opened 2021-06-30
refactor geofeatures properties query

hotzevzl:
see https://vizzuality.atlassian.net/browse/MARXAN-490


hotzevzl:
@aagm except for a test which seems fickle (failed when I last pushed, then succeeded when I run tests again in CI manually), I think all is as we discussed this morning... geo indexes and other feedback from yourself to go into a separate PR

PR #316, opened 2021-07-07
feat(marxan-run): zip input & output directories

kgajowy:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


api/apps/geoprocessing/src/marxan-sandboxed-runner/adapters/marxan-directory.service.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import { FileReader } from './file-reader';
+
+@Injectable()
+export class MarxanDirectory {
+  constructor(private readonly reader: FileReader) {}
+
+  get(
+    type: 'INPUTDIR' | 'OUTPUTDIR',
+    within: Workspace,
```
Dyostiq, line 14: I think `Workspace` is not entirely required here, maybe consider passing just the `workingDirectory`?

api/apps/geoprocessing/src/marxan-sandboxed-runner/adapters/solutions-output/solutions-output.service.ts
```diff
@@ -30,10 +33,19 @@ export class SolutionsOutput implements Cancellable {
...
+    stdOutput: string[],
+    stdErr?: string[],
+  ): Promise<void> {
     if (!existsSync(workspace.workingDirectory + `/output/output_sum.csv`)) {
       throw new Error(`Output is missing from the marxan run.`);
     }
+    await this.metadataRepository.save(scenarioId, workspace, {
+      stdOutput,
+      stdErr,
+    });
```
Dyostiq, line 48: according to my previous comment, the workspace path could be passed here instead of entire workspace

api/libs/marxan-output/src/marxan-execution-metadata.geo.entity.ts
```diff
@@ -0,0 +1,35 @@
...
+  @Column({
+    nullable: false,
+  })
+  scenarioId!: string;
+
+  @Column({ name: `std_out`, nullable: true, type: 'character varying' })
+  stdOutput?: string | null;
+
+  @Column({ name: `std_err`, nullable: true, type: 'character varying' })
+  stdError?: string | null;
```
Dyostiq, line 20: maybe we could keep the same names in columns, so `std_ouput` and `std_error`

api/apps/geoprocessing/src/marxan-sandboxed-runner/adapters/marxan-directory.service.ts
```diff
@@ -0,0 +1,34 @@
...
+  ): {
+    name: string;
+    fullPath: string;
+  } {
+    const config = this.reader.read(
+      joinPath(within.workingDirectory, 'input.dat'),
+    );
+    const matcher = new RegExp(`(${type} )(?<dir>.*)$`, 'gm');
```
Dyostiq, line 22: shouldn't the regex start with `^`?

Dyostiq (approved):

api/apps/geoprocessing/src/marxan-sandboxed-runner/adapters/marxan-directory.service.ts
kgajowy: @Dyostiq it could but doesn't matter there (multiline & known format)

api/apps/geoprocessing/src/marxan-sandboxed-runner/adapters/marxan-directory.service.ts
kgajowy: It could but the intention was to provide this via Workspace to ensure it is what it is supposed to be instead raw string.

Not strongly convinced to either option tho. Do you see any strong benefits by passing only path?
@Dyostiq

api/apps/geoprocessing/src/marxan-sandboxed-runner/adapters/marxan-directory.service.ts
Dyostiq: matters, as the file may contain comments, so:
```
a comment like this suggesting using INPUTDIR something
```
matches the regex

api/apps/geoprocessing/src/marxan-sandboxed-runner/adapters/marxan-directory.service.ts
Dyostiq: From my point of view, it violates the principle of least knowledge a bit and makes the code harder to test as you need to pass the entire workspace instead of only the part. Then, the big object propagates through the stack just only to pass the path. If you think that passing a raw string is not as expressive as you'd like it to be, maybe wrapping it in an object is a solution – but still, you can pass only really needed data.

api/apps/geoprocessing/src/marxan-sandboxed-runner/adapters/marxan-directory.service.ts
kgajowy: Gotcha, you are right!

kgajowy: @Dyostiq https://github.com/Vizzuality/marxan-cloud/pull/316/commits/e6738911d900188778e736492f65887266b9682b

api/apps/geoprocessing/src/marxan-sandboxed-runner/adapters/marxan-directory.service.ts
kgajowy: https://github.com/Vizzuality/marxan-cloud/pull/316/commits/9e4b34bef2825e79f15971ea147f78223049ec4c

PR #323, opened 2021-07-09
feat(api): pass assets to marxan run job

Dyostiq:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/config/default.json
+++ b/api/apps/api/config/default.json
@@ -13,6 +13,9 @@
      },
    "concurrency": 50
    },
+  "api": {
+    "url": "http://api:3000"
+  },
   "jobOptions":{
     "removeOnComplete": 100,
     "removeOnFail": 1000,
--- a/api/apps/api/src/modules/scenarios/marxan-run/assets.service.ts
+++ b/api/apps/api/src/modules/scenarios/marxan-run/assets.service.ts
@@ -1,0 +1,45 @@
+import { Assets } from '@marxan/scenario-run-queue';
+import { FactoryProvider, Inject, Injectable } from '@nestjs/common';
+import { AppConfig } from '@marxan-api/utils/config.utils';
+import { IoSettings, ioSettingsToken } from './io-settings';
+
+export const apiUrlToken = Symbol('api url token');
+export const apiUrlProvider: FactoryProvider<string> = {
+  provide: apiUrlToken,
+  useFactory: () => AppConfig.get<string>('api.url'),
+};
+
+@Injectable()
+export class AssetsService {
+  constructor(
+    @Inject(ioSettingsToken)
+    private readonly settings: IoSettings,
+    @Inject(apiUrlToken)
+    private readonly apiUrlToken: string,
+  ) {}
+
+  async forScenario(id: string): Promise<Assets> {
+    return [
+      {
+        url: `${this.apiUrlToken}/api/v1/marxan-run/scenarios/${id}/marxan/dat/input.dat`,
+        relativeDestination: 'input.dat',
+      },
+      {
+        url: `${this.apiUrlToken}/api/v1/marxan-run/scenarios/${id}/marxan/dat/pu.dat`,
+        relativeDestination: `${this.settings.INPUTDIR}/${this.settings.PUNAME}`,
+      },
+      {
+        url: `${this.apiUrlToken}/api/v1/marxan-run/scenarios/${id}/marxan/dat/bound.dat`,
+        relativeDestination: `${this.settings.INPUTDIR}/${this.settings.BOUNDNAME}`,
+      },
+      {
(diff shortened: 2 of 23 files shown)


api/apps/api/config/default.json
```diff
@@ -13,6 +13,9 @@
      },
    "concurrency": 50
    },
+  "api": {
+    "url": "http://api:3000"
```
kgajowy, line 17: @hotzevzl please keep in mind to set those on deployments.

api/apps/api/src/modules/scenarios/marxan-run/input-parameter-file.provider.ts
```diff
@@ -78,13 +70,15 @@ class InputParameterFile {
...
+    @InjectRepository(Scenario)
+    private readonly scenarioRepository: Repository<Scenario>,
     @Inject(ioSettingsToken)
     private readonly ioSettings: IoSettings,
   ) {}
 
   async getInputParameterFile(scenarioId: string): Promise<string> {
-    const scenario = await this.scenariosService.getById(scenarioId);
+    const scenario = await this.scenarioRepository.findOne(scenarioId);
+    assertDefined(scenario);
```
kgajowy, line 81: shouldn't (isn't it already?) it be done on service layer?
`api/apps/api/src/modules/scenarios/scenarios.service.ts`

api/apps/api/src/modules/scenarios/marxan-run/assets.service.ts
```diff
@@ -0,0 +1,45 @@
+import { Assets } from '@marxan/scenario-run-queue';
+import { FactoryProvider, Inject, Injectable } from '@nestjs/common';
+import { AppConfig } from '@marxan-api/utils/config.utils';
+import { IoSettings, ioSettingsToken } from './io-settings';
+
+export const apiUrlToken = Symbol('api url token');
+export const apiUrlProvider: FactoryProvider<string> = {
+  provide: apiUrlToken,
+  useFactory: () => AppConfig.get<string>('api.url'),
+};
```
kgajowy, line 10: don't we usually keep those in standalone file? (like for api/apps/api/src/modules/scenarios/marxan-run/io-settings.ts below)

api/apps/api/src/modules/scenarios/marxan-run/run.service.ts
```diff
@@ -44,14 +45,18 @@ export class RunService {
...
     private readonly scenarios: Repository<Scenario>,
+    private readonly assets: AssetsService,
   ) {
     queueEvents.on(`completed`, ({ jobId }) => this.handleFinished(jobId));
     queueEvents.on(`failed`, ({ jobId }) => this.handleFailed(jobId));
   }
 
   async run(scenarioId: string): Promise<void> {
+    const assets = await this.assets.forScenario(scenarioId);
+    assertDefined(assets);
```
kgajowy, line 56: As above, we shall keep it in top service which already does that (btw that assert wouldn't ever trigger, as AsserService does the check as well)

api/libs/scenario-run-queue/src/index.ts
```diff
@@ -0,0 +1,12 @@
+export const queueName = 'scenario-run-queue';
+export type Assets = {
+  url: string;
+  relativeDestination: string;
+}[];
```
kgajowy, line 5: Would be great to add a subtask to use this in geoprocessing ❤️ 🙇

kgajowy (approved):

api/apps/api/src/modules/scenarios/marxan-run/assets.service.ts
Dyostiq: io-settings previously also was in one file with its user. I tend to make it a separate files when it has more than one user.

api/apps/api/src/modules/scenarios/marxan-run/input-parameter-file.provider.ts
Dyostiq: it is but I needed to narrow the type here, as `findOne` returns `| undefined`.

On the other hand, it's not in a transaction so it's possible that the assertion will actually throw :)

api/apps/api/src/modules/scenarios/marxan-run/run.service.ts
Dyostiq: it's a leftover, thanks :)

api/apps/api/config/default.json
hotzevzl: yep, as long as they are in `README.md` and in `env.default` we're good - I think you added this one some time ago.

hotzevzl: for values needed in CI, we also need to add them to `.env-test-e2e` (or GH secrets if applicable) and to pass them through via `docker-compose-test-e2e.ci.yml`

hotzevzl (approved):

PR #335, opened 2021-07-14
tag happy-path scenario specs

hotzevzl:
Given that wiring up specs to actual testing behaviour will take a while, I have tried to focus on the happy-path scenarios which should be key for MVP (I did overdo a bit towards the end of the user journey, from feature selection onwards, I think, but we can trim further as we go).

The idea here is to start wiring up the `@happypath` scenarios, leaving the rest for later, while still keeping, in the meanwhile, all the specs as a useful documentation of what we expect users to be able to do on the platform, in a language which is both machine and human friendly (I know, I never hear anyone in daily life talking in Gherkin given-when-then style, but I would say it's at least understandable if not really friendly 😅 ).

### DIFF
--- a/e2e-product-testing/cypress/integration/AdjustPlanningUnits.feature
+++ b/e2e-product-testing/cypress/integration/AdjustPlanningUnits.feature
@@ -11,18 +11,21 @@ Feature: 'Exclude' planning units
         And the user has selected 'Upload shapefile'
         When the user uploads an incorrect shapefile
         Then the user gets an error on screen
+    @happypath
     Scenario: draw lock-out area over map
         Given the user is in the Analysis tab 
         And the user has selected 'Exclude areas'
         And the user has selected 'Draw a shape on map'
         When the user draws on the map
         Then the area is displayed on the map and as a removable modal
+    @happypath
     Scenario: select individual lock-out planning units on map
         Given the user is in the Analysis tab 
         And the user has selected 'Exclude areas'
         And the user has selected 'Select planning units'
         When the user selects planning units on the map
         Then the area is displayed on the map and as a removable modal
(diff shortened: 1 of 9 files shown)


hotzevzl:
@tamaramegan could you have a look and see if you are happy with the prioritization I have suggested in this PR?

I wanted to merge your PR as it's excellent and covers a lot of testing ground, but on the other hand the devs had concerns about having to implement a lot of behaviour for tests, so I would like to use this `@happypath` tag to be able to run only the bare minimum subset of actions that we want to make sure are adequately covered in an MVP, while keeping your full set of specifications so that we can progressively implement the behaviour for more of them as needed, and at the same time have a usable and readable specification of what we want users to be able to do.

If you are happy with the suggested priorities, I suggest to merge your PR with my proposed changes and we're good to go (aka us devs will need to implement the actual behaviours bit by bit. If anything, I would be keen to make the `@happypath` subset smaller - obviously we want users to be able to do everything, but to start with we should test only key actions so that we can get product testing running without a lot of effort.

tamaramegan (approved):
@hotzevzl I think the prioritization you made makes sense as happy path. I've noticed we are leaving out all the uploads and downloads except for the marxan files, the Scheduled planning and all other extra things. We can revise once we start advancing but looks good. I am happy for you to merge, I am very happy that this was useful for you!  Also thanks for correcting typos, I've noticed that there were a few XD

hotzevzl:
@tamaramegan thanks 😃 Indeed, I have left most of the uploads out in order to focus on the bare minimum to go from zero to running Marxan and getting results, with included admin regions, WDPA PAs and features.

Ok, I'll merge all of this then 🚀

PR #342, opened 2021-07-16
documentation for scenario features (geoprocessing)

hotzevzl:


docs/README_geo-feature-processing.md
```diff
@@ -172,27 +211,12 @@ computing throwaway results in order to simulate constant-time operations).
...
-       * see above), the features generated via geoprocessing operations will be
-       * used, each with their marxanSettings as specified for each of them.
-       */
-      kind: 'plain' | 'withGeoprocessing';
-      marxanSettings: {
-        prop: number,
-        fpf: number,
-      },
-
       geoprocessingOperations: [
```
kgajowy, line 214:
```
it's an array in case we need more steps in future phases, but right now we should only accept one operation per array of geoprocessingOperations
```

kgajowy:
```
Stratification doing split: The idea in general is to be able to process a specification in a single pass, and not have any of its parts need to reference outputs of other parts
```

hotzevzl: thanks - I have added these bits and others that emerged through recent discussions here: 50da487f4e0eedbed7540d23ef09baa1dbfddee4

PR #350, opened 2021-07-20
feat(api): skip bound.dat file if no blm given or zero blm

Dyostiq:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/scenarios/marxan-run/run.service.ts
+++ b/api/apps/api/src/modules/scenarios/marxan-run/run.service.ts
@@ -61,8 +61,8 @@ export class RunService {
     );
   }
 
-  async run(scenarioId: string): Promise<void> {
-    const assets = await this.assets.forScenario(scenarioId);
+  async run(scenarioId: string, blm?: number): Promise<void> {
+    const assets = await this.assets.forScenario(scenarioId, blm);
     assertDefined(assets);
     const job = await this.queue.add(`run-scenario`, {
       scenarioId,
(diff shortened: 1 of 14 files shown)


kgajowy (commented):
If blm was not provided directly (as an override), should rather check for scenario's setting first

hotzevzl (approved):
perfect, thanks!
one single nitpick - I suggest to splurge on keystrokes in the two test descriptions:

`should return config without bound when...` -> `should return config without bound.dat when...` - we all know what this is, but let's be kind to whomever may be looking at this codebase for the first time 😉

hotzevzl:
> If blm was not provided directly (as an override), should rather check for scenario's setting first

mh, well spotted

api/apps/api/src/modules/scenarios/marxan-run/run.service.ts
```diff
@@ -42,13 +28,16 @@ export class RunService {
...
   ) {
     queueEvents.on(`completed`, ({ jobId }) => this.handleFinished(jobId));
     queueEvents.on(`failed`, ({ jobId }) => this.handleFailed(jobId));
   }
 
-  async run(scenarioId: string): Promise<void> {
-    const assets = await this.assets.forScenario(scenarioId);
+  async run(scenarioId: string, blm?: number): Promise<void> {
+    blm ??= this.blmDefault;
+    const assets = await this.assets.forScenario(scenarioId, blm);
```
kgajowy, line 40: I am still not sure about this default and place where we apply it. Right now, we decide about final BLM value in two places (run.service and scenario.service below).

```
//scenarios.service.ts
blm ?? scenario.boundaryLengthModifier ?? blmDefault
```

or maybe even force BLM to be non-nullable and insert default upon creating a scenario? (@hotzevzl ?)

Dyostiq: Initially, I've implemented it like this, but I decided to keep defaulting to the `run.service` as it was easier to test, and didn't want to make another connection between scenario service and marxan params

kgajowy (approved):

kgajowy (approved):

PR #352, opened 2021-07-20
fix planning area assignment

Dyostiq:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/projects/projects-crud.service.ts
+++ b/api/apps/api/src/modules/projects/projects-crud.service.ts
@@ -136,6 +136,7 @@ export class ProjectsCrudService extends AppBaseService<
      */
     const project = await super.setDataCreate(create, info);
     project.createdBy = info?.authenticatedUser?.id!;
+    project.planningAreaGeometryId = create.planningAreaId;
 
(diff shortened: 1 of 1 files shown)


api/apps/api/src/modules/projects/projects-crud.service.ts
```diff
@@ -136,6 +136,7 @@ export class ProjectsCrudService extends AppBaseService<
      */
     const project = await super.setDataCreate(create, info);
     project.createdBy = info?.authenticatedUser?.id!;
+    project.planningAreaGeometryId = create.planningAreaId;
```
kgajowy, line 139: 🙈 is geometry the same as area?

kgajowy (approved):

api/apps/api/src/modules/projects/projects-crud.service.ts
Dyostiq: I didn't want to change the api contract, and it looked like setting `planningAreaId` is used to set a custom planning area, whereas `planningAreaId` is a generic calculated field in the entity, and `planningAreaGeometryId` is a actual column in the entity representing the assigned custom planning area

PR #359, opened 2021-07-21
provide temporary column default

hotzevzl:
This will be needed in instances that run this migration while already having some seed data in the `output_scenarios_pu_data` table to avoid a not null constraint violation.


hotzevzl:
@kgajowy I even added this to my local copy of your PR and forgot to raise this in my review 🤦🏼

kgajowy (approved):

PR #360, opened 2021-07-21
feat(api): persist job results

Dyostiq:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


api/apps/api/src/modules/scenarios/marxan-run/run.service.ts
```diff
@@ -89,14 +92,15 @@ export class RunService {
...
-  private async handleFinished(jobId: string, eventId: string) {
+  private async handleCompleted(jobId: string, eventId: string) {
     const job = await this.getJob(jobId);
     const kind = API_EVENT_KINDS.scenario__run__finished__v1__alpha1;
     await this.apiEvents.createIfNotExists({
       topic: job.data.scenarioId,
       kind,
       externalId: eventId,
     });
+    await this.outputs.saveOutput(job);
```
kgajowy, line 103: if saving output fails we may end up in invalid state, don't we?

Dyostiq: Why?
We'll have missing output, but what we can do with that? The API event can't be deleted as it represents something that happened – and `completed` happened, we can't change the past, it is a fact here.
The only thing we can do in case of output saving failure is to take some kind of compensation action, but I am not sure what we should really do. What do you propose?

kgajowy: @Dyostiq if we have incomplete results, should most likely end as failure (the full process)

thus, if saveOutput failed, send `failed` event, otherwise `succees`

Dyostiq: fixed, added two events for saving output

kgajowy (approved):

PR #362, opened 2021-07-21
selection of pu lock status should work with single features

hotzevzl:
See: https://vizzuality.atlassian.net/browse/MARXAN-572

### DIFF
--- a/api/apps/api/src/modules/scenarios/dto/update-scenario-planning-unit-lock-status.dto.ts
+++ b/api/apps/api/src/modules/scenarios/dto/update-scenario-planning-unit-lock-status.dto.ts
@@ -1,4 +1,4 @@
-import { IsOptional, IsUUID, ValidateNested } from 'class-validator';
+import { IsArray, IsOptional, IsUUID, ValidateNested } from 'class-validator';
 import { ApiPropertyOptional } from '@nestjs/swagger';
 import { FeatureCollection, MultiPolygon, Polygon } from 'geojson';
 import { Type } from 'class-transformer';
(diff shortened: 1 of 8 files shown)


hotzevzl:
@kgajowy this is as far as I got yesterday and I will be out until back 😬 today - could you have a look and see if you think we should test anything else/differently?

I tried to make the tests kind of flexible to accommodate different kinds of setups (which I had the unfortunate idea of calling `case` in the code, with obvious risk of clashing with holy reserved words 🤷🏼) but that was before I realised that the bug I thought was _the_ bug, in fact wasn't the only nor first one in the lifecycle of these requests, so we still can't easily test for payloads like the one that the frontend was sending, but we may want to trust in `class-validator`.

Something that could be improved is the OpenAPI spec output, by bringing in the correct payload schema for `byGeoJson.include` and `byGeoJson.exclude`.

One thing I didn't have time to figure out is the behaviour in case of overlapping requests for a PU such as in the one with both an inclusion and exclusion request in the screenshot below. From a quick glance at the code I was expecting the last one to be calculated to "win" but I then had to fiddle with the expected test results to make the test work for this case. Nothing to worry about in this PR, and I may most likely just have misread things in a hurry, but let's cross check with Alicia that the behaviour we provide is correct.

planning-units

Please have a look, feel free to improve/change and merge so that Miguel can be unblocked on this (@aagm can you deploy on marxan49 when merged, please?). I have tested this on marxan49 by wrapping Miguel's example payload in an array where required and I could see the PU selection was being correctly persisted, when loading back the vector tiles for this.

hotzevzl:
@kgajowy the `singleFeature` mock data doesn't include the lower red and lower green shapes.

The original implementation would work only with exactly two features in a feature collection, and that's what the tests were testing, hence my suggestion to add a way to test the `singleFeature` case as a way to catch possible regressions if we ever touch that query again and somehow end up without the array bit.

I'll leave all this up to you folks (I'll shortly be MIA)

hotzevzl:
> @kgajowy the `singleFeature` mock data doesn't include the lower red and lower green shapes.

I mean this:

singlefeature

kgajowy (approved):

PR #364, opened 2021-07-22
Carousel: react flicking

mbarrenechea:
### Overview

I tried to fix an error in the keen slider but I end up having more. I decided to move to this library and I think it works perfectly


anamontiaga:
ok no problem. think we need to update the story deleting args we are not using

PR #365, opened 2021-07-22
feat(geoprocessing): return bbox and pu sizes during shapefile creation

Dyostiq:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/geoprocessing/src/modules/planning-area/marxan-geo-json.ts
+++ b/api/apps/geoprocessing/src/modules/planning-area/marxan-geo-json.ts
@@ -1,0 +1,28 @@
+import { BBox, GeoJSON } from 'geojson';
+import { PuSizes } from './pu-sizes';
+
+export class MarxanGeoJson {
+  private readonly geoJson: GeoJSON & { bbox: BBox } & {
+    marxanMetadata: {
+      maxPuAreaSize: number;
+      minPuAreaSize: number;
+    };
+  };
+
+  constructor(geoJson: GeoJSON, bbox: BBox, puSizes: PuSizes) {
+    this.geoJson = {
+      ...geoJson,
+      bbox: bbox,
+      marxanMetadata: {
+        maxPuAreaSize: puSizes.maxPuAreaSize,
+        minPuAreaSize: puSizes.minPuAreaSize,
+      },
+    };
+  }
+
+  toJSON(): GeoJSON & {
+    marxanMetadata?: Record<string, unknown>;
+  } {
+    return this.geoJson;
+  }
+}
--- a/api/apps/geoprocessing/src/modules/planning-area/planning-area.controller.ts
+++ b/api/apps/geoprocessing/src/modules/planning-area/planning-area.controller.ts
@@ -6,17 +6,20 @@ import {
   Post,
 } from '@nestjs/common';
 import { ApiOkResponse } from '@nestjs/swagger';
-import { plainToClass } from 'class-transformer';
 import { ApiConsumesShapefile } from '@marxan-geoprocessing/decoratos/shapefile.decorator';
-import { PlanningAreaService } from '@marxan-geoprocessing/modules/planning-area/planning-area.service';
 import { apiGlobalPrefixes } from '@marxan-geoprocessing/api.config';
 import { PlanningAreaDto } from './planning-area.dto';
+import { PlanningAreaService } from './planning-area.service';
+import { PlanningAreaSerializer } from './planning-area.serializer';
 
 @Controller(`${apiGlobalPrefixes.v1}/projects/planning-area`)
 export class PlanningAreaController {
   private readonly logger = new Logger(this.constructor.name);
 
-  constructor(private planningAreaService: PlanningAreaService) {}
+  constructor(
+    private planningAreaService: PlanningAreaService,
+    private planningAreaSerializer: PlanningAreaSerializer,
+  ) {}
 
   @ApiConsumesShapefile()
   @ApiOkResponse({ type: PlanningAreaDto })
@@ -25,10 +28,8 @@ export class PlanningAreaController {
     @Body() shapefileInfo: Express.Multer.File,
   ): Promise<PlanningAreaDto> {
     try {
-      return plainToClass<PlanningAreaDto, PlanningAreaDto>(
-        PlanningAreaDto,
-        await this.planningAreaService.save(shapefileInfo),
-      );
+      const result = await this.planningAreaService.save(shapefileInfo);
+      return this.planningAreaSerializer.serialize(result);
     } catch (error) {
       this.logger.error(
         'failed to create a planning area from shapefile: ' + error.message,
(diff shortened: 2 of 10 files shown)


api/apps/geoprocessing/src/modules/planning-area/marxan-geo-json.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  addBBox(bbox: BBox) {
+    this.geoJson.bbox = bbox;
+  }
+
+  addPuSizes(puSizes: PuSizes) {
+    this.addMarxanMetadata('maxPuAreaSize', puSizes.maxPuAreaSize);
+    this.addMarxanMetadata('minPuAreaSize', puSizes.minPuAreaSize);
+  }
```
kgajowy, line 20: that's actually `set` but its less than minor

api/apps/geoprocessing/src/modules/planning-area/marxan-geo-json.ts
```diff
@@ -0,0 +1,32 @@
+import { BBox, GeoJSON } from 'geojson';
+import { PuSizes } from './pu-sizes';
+
+export class MarxanGeoJson {
```
kgajowy, line 4: `marxan` isn't giving any idea what it is. What consumers use it for?

kgajowy: There is also another "Marxan" GeoJson which holds cost & puId of each geometry inside.

api/apps/geoprocessing/src/modules/planning-area/planning-area.controller.ts
```diff
@@ -25,10 +24,8 @@ export class PlanningAreaController {
     @Body() shapefileInfo: Express.Multer.File,
   ): Promise<PlanningAreaDto> {
     try {
-      return plainToClass<PlanningAreaDto, PlanningAreaDto>(
-        PlanningAreaDto,
-        await this.planningAreaService.save(shapefileInfo),
-      );
+      const result = await this.planningAreaService.save(shapefileInfo);
+      return PlanningAreaDto.from(result);
```
kgajowy, line 28: It isn't that I don't like it but why not `serializer` as we commonly use it around? Just because of unnecessary boilerplate?

api/apps/geoprocessing/src/modules/planning-area/marxan-geo-json.ts
```diff
@@ -0,0 +1,32 @@
...
+    this.geoJson.bbox = bbox;
+  }
+
+  addPuSizes(puSizes: PuSizes) {
+    this.addMarxanMetadata('maxPuAreaSize', puSizes.maxPuAreaSize);
+    this.addMarxanMetadata('minPuAreaSize', puSizes.minPuAreaSize);
+  }
+
+  addMarxanMetadata<V>(key: string, value: V) {
+    this.geoJson.marxanMetadata = {
```
kgajowy, line 23: shouldn't this land in `properties` of feature?

api/apps/geoprocessing/src/modules/planning-area/planning-area.dto.spec.ts
```diff
@@ -0,0 +1,117 @@
+import { PlanningAreaDto } from './planning-area.dto';
+
+const input = () => ({
```
kgajowy, line 3:
```suggestion
const input = (): GeoJSON & {puSizes?: ....}  => ({
```
sqrt(💊) as `as const` seems to be awkward workaround for GeoJSON type 🤔 nevertheless, just specs.  

(but I still think that those additional metadata should go into `properties` of a feature)

api/apps/geoprocessing/src/modules/planning-area/planning-area.dto.ts
```diff
@@ -1,10 +1,31 @@
...
   data!: GeoJSON;
+
+  static from(input: {
+    id: string;
+    data: GeoJSON;
+    bbox?: BBox;
+    puSizes?: PuSizes;
+  }) {
+    const marxanGeoJSON = new MarxanGeoJson(input.data);
+    if (isDefined(input.bbox)) marxanGeoJSON.addBBox(input.bbox);
```
kgajowy, line 21: each `GeoJSON` must have a `bbox` already. How we do ensure that bbox is actually a bbox around the shapes inside?

api/apps/geoprocessing/src/modules/planning-area/marxan-geo-json.ts
Dyostiq: no, we wanted to have it as a foreign member https://datatracker.ietf.org/doc/html/rfc7946#section-6.1

api/apps/geoprocessing/src/modules/planning-area/marxan-geo-json.ts
Dyostiq: where?

this one is specific to the planning area, adding something custom to GeoJSON and serializing it in toJSON,

just encapsulates a bit where GeoJSON keeps bbox and pu sizes

it's used in PlanningAreaDto, and can be used in other dtos when needed

api/apps/geoprocessing/src/modules/planning-area/planning-area.controller.ts
Dyostiq: do you mean `serialize` and `serializerConfig` from the base service?

yes,
it's not a regular entity, we do not keep the GeoJSON (and pu sizes) in DB, at least not in this shape
also, there is always only one entity returned at a time

api/apps/geoprocessing/src/modules/planning-area/planning-area.dto.spec.ts
Dyostiq: it will be a little more verbose but ok :)

api/apps/geoprocessing/src/modules/planning-area/planning-area.dto.ts
Dyostiq: hmm, I have no idea why I assumed it's optional.

we do not validate it, we just trust what we got from the DB

Dyostiq: fixed

api/apps/geoprocessing/src/modules/planning-area/planning-area.dto.spec.ts
Dyostiq: fixed

api/apps/geoprocessing/src/modules/planning-area/planning-area.controller.ts
Dyostiq: added to be consistent

api/apps/geoprocessing/src/modules/planning-area/marxan-geo-json.ts
Dyostiq: I've removed these methods

api/apps/geoprocessing/src/modules/planning-area/marxan-geo-json.ts
kgajowy: `api/apps/geoprocessing/src/modules/surface-cost/adapters/pu-cost-extractor.ts`

kgajowy (approved):

api/apps/geoprocessing/src/modules/planning-area/marxan-geo-json.ts
Dyostiq: so it has a different purpose

PR #369, opened 2021-07-26
feat(features): feature-specification domain

kgajowy:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/features/domain/feature-specification-revision.ts
+++ b/api/apps/api/src/modules/features/domain/feature-specification-revision.ts
@@ -1,0 +1,20 @@
+import {
+  ensure,
+  isDefined,
+  isInteger,
+  TinyType,
+  isGreaterThanOrEqualTo,
+} from 'tiny-types';
+
+export class FeatureSpecificationRevision extends TinyType {
+  constructor(public readonly value: number) {
+    super();
+    ensure(
+      'FeatureSpecificationRevision',
+      value,
+      isDefined(),
+      isInteger(),
+      isGreaterThanOrEqualTo(1),
+    );
+  }
+}
--- a/api/apps/api/src/modules/features/domain/feature-specification.spec.ts
+++ b/api/apps/api/src/modules/features/domain/feature-specification.spec.ts
@@ -1,0 +1,110 @@
+import {
+  FeatureSpecification,
+  SpecificationStatus,
+} from './feature-specification';
+import { FeatureSpecificationId } from './feature-specification.id';
+import { FeatureSpecificationRevision } from './feature-specification-revision';
+
+describe(`when specification was already created`, () => {
+  let spec: FeatureSpecification;
+  beforeEach(() => {
+    spec = new FeatureSpecification(
+      new FeatureSpecificationId(`spec-id`),
+      SpecificationStatus.Created,
+      [],
+      new FeatureSpecificationRevision(5),
+    );
+  });
+
+  describe(`when submitting again`, () => {
+    beforeEach(() => {
+      spec.submit();
+    });
+
+    it(`should emit compute event`, () => {
+      expect(spec.getUncommittedEvents()).toMatchInlineSnapshot(`
+        Array [
+          ComputeFeatureSpecificationEvent {
+            "featureSpecificationId": "spec-id",
+            "revision": 6,
+          },
+        ]
+      `);
+    });
+
+    it(`should change state`, () => {
+      expect(spec.getStatus()).toEqual(SpecificationStatus.Created);
+    });
+    it(`should set revision to 6`, () => {
+      expect(spec.getRevision()).toEqual(6);
+    });
+  });
+
+  describe(`when changing spec`, () => {
+    beforeEach(() => {
+      spec.change({
+        features: [],
+      });
+    });
+
+    it(`should not emit compute event`, () => {
+      expect(spec.getUncommittedEvents()).toEqual([]);
+    });
+    it(`should put into draft state`, () => {
+      expect(spec.getStatus()).toEqual(SpecificationStatus.Draft);
+    });
+    it(`should generate new revision number`, () => {
+      expect(spec.getRevision()).toEqual(6);
+    });
+  });
+});
+
+describe(`when creating brand new specification`, () => {
+  let spec: FeatureSpecification;
+  beforeEach(() => {
+    spec = new FeatureSpecification(
+      new FeatureSpecificationId(`spec-id`),
+      SpecificationStatus.Draft,
+      [],
+    );
+  });
+
+  describe(`when submitting`, () => {
+    beforeEach(() => {
+      spec.submit();
+    });
+    it(`should emit compute event`, () => {
+      expect(spec.getUncommittedEvents()).toMatchInlineSnapshot(`
+        Array [
+          ComputeFeatureSpecificationEvent {
+            "featureSpecificationId": "spec-id",
+            "revision": 1,
+          },
+        ]
+      `);
+    });
+    it(`should change state`, () => {
+      expect(spec.getStatus()).toEqual(SpecificationStatus.Created);
+    });
+    it(`should set revision to 1`, () => {
+      expect(spec.getRevision()).toEqual(1);
+    });
+  });
+
+  describe(`when changing configuration`, () => {
+    beforeEach(() => {
+      spec.change({
+        features: [],
+      });
+    });
+    it(`should not emit compute event`, () => {
+      expect(spec.getUncommittedEvents()).toEqual([]);
+    });
+    it(`should keep draft state`, () => {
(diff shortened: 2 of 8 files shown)


api/apps/api/src/modules/features/domain/feature-specification.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+        features: [],
+      });
+    });
+
+    it(`should not emit compute event`, () => {
+      expect(spec.getUncommittedEvents()).toEqual([]);
+    });
+    it(`should put into draft state`, () => {
+      expect(spec.getStatus()).toEqual(SpecificationStatus.Draft);
+    });
```
hotzevzl, line 55: I would say that it may be possible for a user to change a specification and submit it for processing at the same time (hence `SpecificationStatus.Created`).

api/apps/api/src/modules/features/domain/feature-specification-revision.ts
```diff
@@ -0,0 +1,20 @@
...
+    super();
+    ensure(
+      'FeatureSpecificationRevision',
+      value,
+      isDefined(),
+      isInteger(),
+      isGreaterThanOrEqualTo(1),
+    );
+  }
+}
```
hotzevzl, line 20: though I like the idea behind tiny types, I would not do this while we are already using `class-validator` extensively or w open the door to using a slightly different hammer for each slightly different shape of nail. If we don't need to add validity guarantees a `type FirstName = string` should do, and if we want validation, I would use a class and decorate its properties via `class-validator`.

maybe in this specific case, defining an interface with a single `class-validator`-annotated revision property and implementing this in the FeatureSpecification class?

api/apps/api/src/modules/features/domain/feature-specification.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+});
+
+describe(`when creating brand new specification`, () => {
+  let spec: FeatureSpecification;
+  beforeEach(() => {
+    spec = new FeatureSpecification(
+      new FeatureSpecificationId(`spec-id`),
+      SpecificationStatus.Draft,
+      [],
+    );
```
hotzevzl, line 69: this is ok but let's not forget that an initial status of `Draft` is not going to be necessarily the case - `Created` can be set straight away (the documentation uses `Draft` just for simplicity's sake)

api/apps/api/src/modules/features/domain/feature-specification.spec.ts
kgajowy: @hotzevzl I suppose this case is covered in tests.

```
when creating brand new specification > when submitting
```

so, from system/product perspective, its creating "draft" on which it will be immediately `submit`ted.

just to distinguish it, we may put some static `fresh(...payload)` method which will assign default values (or actually, it can be a domain object factory with such method)

api/apps/api/src/modules/features/domain/feature-specification.spec.ts
kgajowy: https://github.com/Vizzuality/marxan-cloud/pull/369#discussion_r676462727
Feel like the very same case.

api/apps/api/src/modules/features/domain/feature-specification.spec.ts
kgajowy: hence, this case would be handled by the application as:
```
const spec = SpecificationFactory.fresh(...payload) // including validation
spec.submit()
```

api/apps/api/src/modules/features/domain/feature-specification-revision.ts
kgajowy: `tiny-types` provide a bit more intention cases (as `type X = string` does not imply any safety when using the class, while `new Revision(3)` is a bit more expressive.

If we were to use `class-validator` it would imply more code than necessary - additional dto class and manual validation within the ValueObject (or even more, use single-propertied DTO ...)

The approach seems to be just more suitable for the needs - forcing ourselves to use library we use somewhere else isn't always the best way. Tools should help us rather than dictate how we implement things, especially when it comes to such small "cases" - we could paraphrase hammer metaphor - "why to use same heavy hammer, while just finger is needed" ? :)

hotzevzl: yes, indeed I am on the fence on this one. I would even favour a bit more verbosity/more moving pieces over more ways to do something (or something similar, I agree, the use case is different), for the sake of ongoing maintainability.

If we limit this to the indended use case, that is individual primitive types then it makes sense. Time to start `CODING_STYLE.md`, I guess 🤔

api/apps/api/src/modules/features/domain/feature-specification.spec.ts
hotzevzl: yes, but in practice we ask for all the properties to be submitted (unless they are optional, of course), and we don't need to compare with what we already have (if anything) - or do you mean to compare the stored specification with the one in a new request? If so I would just hash a normalized serialization - if hash doesn't match => new revision (could even match a previous one, we shouldn't care, I think).

`when changing spec > it should put into draft state` should not necessarily be the case - but I may be misreading the intention behind this test.

I would treat each request as a full replacement of any previous specification, and except for the hashing bit if we want to keep track of revisions to see if what is calculated matches what the user can `GET`, I don't think we need to have default values - the optional bits are really optional (a feature set can be empty, mostly by user mistake, and we will not have `prop` and `fpf` until the second step of the process in the UI, but other than that, we should always receive a `status` and enforce that is present and is one of the allowed enum values.

Then we can use per-geoprocessing operation hashing to check if we need to *recompute* an operation, but that's a separate story.

kgajowy: @hotzevzl I am a bit lost and confused - may be that we are misreading the purpose of this domain class.

It totally abstracts from what is within specification, it "only" handles draft/created states transitions according to specification and what happens with the revision once one of the actions is done. It neither knows the details, nor the user context, nor the recompute needs (as you mentioned, geo will take care of that).

Especially two cases:
```
// submit -> submit again ; which is actually creating a new versions
if the specification is submitted again with status `created`, resulting
  features will be computed again, any existing features linked to the scenario
  will be unlinked from it, and the newly calculated features will be linked to
  it
```

```
// it isn't the case from user perspective, as he wants to modify & submit but that actually creates a new version (think of Conflucence page modification)
"when changing spec > it should put into draft state" should not necessarily be the case - but I may be misreading the intention behind this test.
```

api/apps/api/src/modules/features/domain/feature-specification-revision.ts
Dyostiq: Actually, it's possible to use both libs without extra DTO class:
```
const ClassValidatorTinyTypeOf = <T>() => {
  return class extends TinyTypeOf<T>() {
    constructor(value: T) {
      super(value);
      const result = validateSync(this);
      if (result.length > 0) {
        throw result;
      }
    }
  };
};

class ATinyType extends ClassValidatorTinyTypeOf<number>() {
  @IsDivisibleBy(1)
  value!: number;
}
```

But it is artificial a bit, and I agree with @kgajowy that using the validator from tiny types is a better fit here. I'd keep class-validator to define validators for DTOs as it plays well with built-in nestjs behavior and tiny-types for domain objects. I think we need to find a balance to not use class-validator as Maslow's hammer and to not mix too many libs for validating unexpectedly.

The last important thing is that tiny-types are not MIT, but Apache 2.0. Is it a problem @hotzevzl?

api/apps/api/src/modules/features/domain/feature-specification.spec.ts
hotzevzl: yes, it's fine that this is only about the state transitions.

the Confluence example helps, thanks. I was thinking about how our own frontend is using these interim `PUT`s, but yes, if a user were to only update "something else except the status" we should indeed assume that the new copy is a draft. If we enforce `status` to be present in the payload (which we do, IIRC) this may be immaterial, but ok in any case.

api/apps/api/src/modules/features/domain/feature-specification-revision.ts
hotzevzl: @Dyostiq your example is pretty much the first thing that came to my mind when looking at the `tiny-types` documentation, but yes, the extra verbosity makes the "tiny" selling point less tiny.

Let's apply the rule of three to these contexts (vs code duplication) - if we get to a third way of doing validation then we should all get a flashing red alert lighting up.

hotzevzl: as for the package's license, from my understanding it is ok - do you have specific concerns?

PR #370, opened 2021-07-26
feat(scenarios): allow to get all planning-units with their lock-status

kgajowy:
https://vizzuality.atlassian.net/browse/MARXAN-577

image


hotzevzl (approved):
looks good, thanks!

PR #372, opened 2021-07-26
Feature/help scenario

tamaramegan:
I have added info buttons and help to the scenarios section.
I am not finished yet but there were many files piling up and I thought it was better to stop and get this merged before continuing forward.

I would like to discuss what to do about the split and intersect buttons (currently showing one per feature)


app/layout/scenarios/sidebar/wdpa/threshold/component.tsx
```diff
@@ -147,7 +147,26 @@ export const WDPAThreshold: React.FC<WDPAThresholdCategories> = ({
...
+
+                          Inside Marxan, planning units are considered as either
+                          protected
+                          or not protected. The threshold value represents a
+                          percentage of the area
+                          inside a planning unit. By setting the threshold you decide
+                          how much of a protected area needs to fall inside a
+                          planning unit to consider the whole planning unit
+                          as &quot;protected&quot;.
+                          <br />
```
mbarrenechea, line 161: I don't like `<br />`. it would be better if we wrap the paragraphs into `<p>` tags and set in the `div` above the following class. `className="space-y-2"`. If you can do this for all the `<br />` you added I'll apreciate it. Remember to pull because I just merge it from develop and fix all the conflicts

mbarrenechea (changes requested):

app/layout/scenarios/sidebar/wdpa/threshold/component.tsx
tamaramegan: Oki doki!

PR #373, opened 2021-07-26
feat: order imports eslint configs

anamontiaga:
### Overview

Test to sort imports by path groups using an eslint plugin. We would need to align if this is the order we want and add all the dependencies

### Testing instructions

- [ ] The order is generated with the autosave.
- [ ]  It has not yet been implemented in the files. It will do it as we work on each file.

### DIFF
--- a/app/.eslintrc.js
+++ b/app/.eslintrc.js
@@ -94,8 +94,6 @@ module.exports = {
         ],
         pathGroupsExcludedImportTypes: [
           'react',
-          'react-final-form',
-          'react-redux',
         ],
       },
     ],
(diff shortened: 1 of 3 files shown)


anamontiaga:
This is how it fit on a file with many types of import. It is only a proposal:

 sort imports

app/.eslintrc.js
```diff
@@ -31,5 +31,71 @@ module.exports = {
...
+            group: 'external',
+            position: 'before',
+          },
+          {
+            pattern: 'store/**',
+            group: 'external',
+          },
+          {
+            pattern: 'hooks/**',
+            group: 'unknown',
```
mbarrenechea, line 73: why `hooks` folder belongs to `unkown` group?

mbarrenechea: I don't like that internal imports come before hooks

image

mbarrenechea (changes requested):

anamontiaga:
Result after changes requested made:

order:imports fixes

app/.eslintrc.js
anamontiaga: Attached new proposal.

mbarrenechea (approved):
Let's see how it works

PR #386, opened 2021-08-02
attach planning units to scenario

hotzevzl:
see https://vizzuality.atlassian.net/browse/MARXAN-603 / https://vizzuality.atlassian.net/browse/MARXAN-623

needs tests

### DIFF
--- a/api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-linker-service.ts
+++ b/api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-linker-service.ts
@@ -26,12 +26,13 @@ export class ScenarioPlanningUnitsLinkerService {
   /** Currently we only support linking regular planning unit geometries
    * computed via ST_HexagonGrid or ST_SquareGrid */
   private isPlanningUnitGridShapeSupported(
-    shape: PlanningUnitGridShape,
+    shape: PlanningUnitGridShape | undefined,
   ): boolean {
-    return [
-      PlanningUnitGridShape.hexagon,
-      PlanningUnitGridShape.square,
-    ].includes(shape);
+    return !isNil(shape)
+      ? [PlanningUnitGridShape.hexagon, PlanningUnitGridShape.square].includes(
+          shape,
+        )
+      : false;
   }
 
   /**
@@ -40,16 +41,12 @@ export class ScenarioPlanningUnitsLinkerService {
    */
   async link(scenario: Scenario): Promise<void> {
     const project = await this.projectsRepo.findOneOrFail(scenario.projectId);
-    if (
-      !this.isPlanningUnitGridShapeSupported 
-    )
+    if (!this.isPlanningUnitGridShapeSupported(project.planningUnitGridShape))
       throw new Error(
         'Only square or hexagonal planning unit grids are supported.',
       );
-    if(
-      isNil(project.planningUnitAreakm2) ||
-      isNil(project.bbox)
-    ) throw new Error('Incomplete project data: this may be a bug.');
+    if (isNil(project.planningUnitAreakm2) || isNil(project.bbox))
+      throw new Error('Incomplete project data: this may be a bug.');
 
     const [xmin, ymin, xmax, ymax] = project.bbox;
     const query = `insert into scenarios_pu_data (pu_geom_id, scenario_id, puid)
--- a/api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-linker-service.ts
+++ b/api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-linker-service.ts
@@ -29,8 +29,6 @@ export class ScenarioPlanningUnitsLinkerService {
     private readonly projectsRepo: Repository<Project>,
   ) {}
 
-  /** Currently we only support linking regular planning unit geometries
-   * computed via ST_HexagonGrid or ST_SquareGrid */
   private isPlanningUnitGridShapeRegular(
     shape: PlanningUnitGridShape | undefined,
   ): boolean {
--- a/api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-linker-service.ts
+++ b/api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-linker-service.ts
@@ -10,7 +10,13 @@ import {
 } from '@marxan-api/modules/projects/project.api.entity';
 import { Scenario } from '../scenario.api.entity';
 import { isNil } from 'lodash';
+import { AdminAreasService } from '@marxan-api/modules/admin-areas/admin-areas.service';
 
+/**
+ * Literal query parts to be used in template query to select planning areas
+ * relevant for the current scenario, based on parent project settings.
+ */
+type QueryPartsForLinker = { planningUnitSelectionQueryPart: string, planningUnitIntersectionQueryPart: string };
 @Injectable()
 export class ScenarioPlanningUnitsLinkerService {
   constructor(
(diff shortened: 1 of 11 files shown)


api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-linker-service.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  /**
+   * Link (existing) planning unit geometries matching project settings and
+   * falling within the project bbox to the scenario being created.
+   */
+  async link(scenario: Scenario): Promise<void> {
+    const project = await this.projectsRepo.findOneOrFail(scenario.projectId);
+    if (
+      !this.isPlanningUnitGridShapeSupported
```
kgajowy, line 44: it's always false, isn't it? (`callme()`)

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-linker-service.ts
```diff
@@ -0,0 +1,65 @@
...
+      throw new Error(
+        'Only square or hexagonal planning unit grids are supported.',
+      );
+    if(
+      isNil(project.planningUnitAreakm2) ||
+      isNil(project.bbox)
+    ) throw new Error('Incomplete project data: this may be a bug.');
+
+    const [xmin, ymin, xmax, ymax] = project.bbox;
+    const query = `insert into scenarios_pu_data (pu_geom_id, scenario_id, puid)
```
kgajowy, line 55: Shouldn't it remove previous ones?

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-linker-service.ts
hotzevzl: well spotted, that was completely unfinished, just happened to please tsc in its form 😬

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-linker-service.ts
hotzevzl: it shouldn't, no - once we create a scenario, the linking stays forever.

this can change if the user edits the parent project and changes planning area, grid shape or grid size, in which case we'll need to delete & gc all the scenarios, and when a user creates a new scenario in the edited project, we redo the linking here.

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-linker-service.ts
hotzevzl: fixed (actually implemented) in d168393b1bb033fbe88a18dfa0140c65726f14f6

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-linker-service.ts
```diff
@@ -0,0 +1,63 @@
...
+      throw new Error('Incomplete project data: this may be a bug.');
+
+    const [xmin, ymin, xmax, ymax] = project.bbox;
+    const query = `insert into scenarios_pu_data (pu_geom_id, scenario_id, puid)
+    select id as pu_geom_id, '${scenario.id}' as scenario_id, row_number() over () as puid
+    from planning_units_geom pug
+    where
+      type = '${project.planningUnitGridShape}' and
+      size = '${project.planningUnitAreakm2}' and
+      st_intersects(the_geom, ST_GeomFromText('MULTIPOLYGON (((${xmin} ${ymin}, ${xmin} ${ymax}, ${xmax} ${ymax}, ${xmax} ${ymin}, ${xmin} ${ymin})))', 4326));`;
```
aagm, line 59: You are intersecting this with the bbox, we need to match the actual geometry either from the admin area table or with the custom planning area shape

aagm (changes requested):

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-linker-service.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    const { planningUnitSelectionQueryPart, planningUnitIntersectionQueryPart } = this.getQueryPartsForLinker(project);
+
+    const query = `insert into scenarios_pu_data (pu_geom_id, scenario_id, puid)
+    select id as pu_geom_id, '${scenario.id}' as scenario_id, row_number() over () as puid
+    from planning_units_geom pug
+    where
+      ${planningUnitSelectionQueryPart} and
+      st_intersects(the_geom, ${planningUnitIntersectionQueryPart});`;
+    Logger.debug(query);
```
kgajowy, line 118:
```suggestion
```

api/apps/api/src/modules/scenarios/scenarios.service.ts
```diff
@@ -86,7 +86,9 @@ export class ScenariosService {
 
   async create(input: CreateScenarioDTO, info: AppInfoDTO) {
     const validatedMetadata = this.getPayloadWithValidatedMetadata(input);
-    return this.crudService.create(validatedMetadata, info);
+    const scenario = await this.crudService.create(validatedMetadata, info);
+    await this.planningUnitsLinkerService.link(scenario);
```
kgajowy, line 90: 1 Isn't it necessary for update as well?

2 Seems like creating scenario has some "side" effects (or is just a part of creation process?). Would recommend putting it into standalone service to keep things together.

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-linker-service.ts
```diff
@@ -0,0 +1,121 @@
...
+    if(this.isProjectUsingCustomPlanningArea(project) && this.isProjectUsingRegularPlanningUnitGrid(project)) {
+      return {
+        planningUnitIntersectionQueryPart: `type = '${PlanningUnitGridShape.fromShapefile}' and project_id = '${project.id}'`,
+        planningUnitSelectionQueryPart: `type = '${project.planningUnitGridShape}' and size = ${project.planningUnitAreakm2}`,
+      }
+    }
+
+    if(this.isProjectUsingGadmPlanningArea(project) && this.isProjectUsingCustomPlanningUnitGrid(project)) {
+      const adminAreaId = project.adminAreaLevel2Id ?? project.adminAreaLevel1Id ?? project.countryId;
+      const adminAreaLevel = AdminAreasService.levelFromId(adminAreaId);
```
kgajowy, line 83: if we are only using some static method (util in this case), importing whole module shouldn't be necessary.

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-linker-service.ts
hotzevzl: @aagm 12bac7c640822068755fc86d7cdf0cf59458abec should handle all the cases correctly: does this look ok to you?

I will still need to fix what broke tests after this change, and to address other PR comments, but this commit should have the core changes that take care of correct intersection (also introducing initial support for custom PU grids, which will depend on some changes to accommodate these imports, but that's a separate story)

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-linker-service.ts
hotzevzl: do you mean to avoid static methods and promoting utility functions to module exports? e.g.

```
export const levelFromId = (areaId: string): number => {
  return areaId.match(/\./g)?.length ?? 0;
}
```

and

```
import { levelFromId } from AdminAreasService;
```

hotzevzl: if so, I'd actually quite favour this in general, since the (ES) module already works as a namespace so having the function as a class method adds namespacing redundancy, in practice.

api/apps/api/src/modules/scenarios/scenarios.service.ts
hotzevzl: on update of scenario, we shouldn't relink PUs to scenario, but it's true that we don't yet enforce that a scenario can only be created if the parent project has enough (and coherent) settings to let us attach a PU grid to the scenario. In practice this shouldn't be an issue with the current frontend flow so for the sake of getting closer to preparing demo projects and test running them in an actual instance I'd be keen to skip this for the moment.

I actually had a check for this in an earlier commit in this PR but I've dropped it now because it was not in the most appropriate place and it would break tests and I just can't update more tests and fixtures 🤯 (again, to get closer to run this whole workflow)

as for side effects, yes and no - I'd say these are side effects, but they are part of the scenario creation process. if the settings of the parent project that let us link PUs to a scenario change, we'll have (in a future iteration) to delete all the scenarios of the project that would be affected and cascade deletion donwstream.

kgajowy: similar to another comment related to keeping the create/update logic in a single place: right now it is scattered around 3 services, among those one is solely `crud.service` but it doesn't fully create a scenario (the facade which calls it has the knowledge that some other services should be called)

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-linker-service.ts
kgajowy: @hotzevzl in this particular case - yes, some util could be enough. 

What I meant is that you can still use this static method without importing the module / injecting service (as it doesn't use the DI and any of the module's services instances directly).

Although, using just a util seems a fair trade for me, as creating connection (even if not through DI) to different module is a step forward to "everything uses everything" :)

PR #387, opened 2021-08-02
calculate default pu lock status

hotzevzl:
See https://vizzuality.atlassian.net/browse/MARXAN-603 / https://vizzuality.atlassian.net/browse/MARXAN-624


api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-protection-status-calculator-service.ts
```diff
@@ -0,0 +1,46 @@
+import { Injectable, Logger } from '@nestjs/common';
+import { InjectRepository } from '@nestjs/typeorm';
+import { Repository } from 'typeorm';
+
+import { DbConnections } from '@marxan-api/ormconfig.connections';
+import { ScenariosPlanningUnitGeoEntity } from '@marxan/scenarios-planning-unit';
+import { Scenario } from '../scenario.api.entity';
+
+@Injectable()
+export class ScenarioPlanningUnitsProtectedStatusCalculatorService {
```
kgajowy, line 10: This service could be useful for geoprocessing as well (when changing inclusion status).

hotzevzl: yep, we should move this to a shared library in a second pass

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-protection-status-calculator-service.ts
```diff
@@ -0,0 +1,46 @@
...
+    console.log(wdpaIucnCategoriesForScenario);
+
+    const query = `
+    with pa as (select * from wdpa where iucn_cat in (${wdpaIucnCategoriesForScenario})),
+    pu as (
+    select spd.id, pug.the_geom
+    from scenarios_pu_data spd
+    inner join planning_units_geom pug on spd.pu_geom_id = pug.id
+    where scenario_id='${scenario.id}'),
+    pu_pa as (select pu.id, st_area(st_intersection(pu.the_geom, pa.the_geom)) as pa_pu_area, (CASE pu.the_geom && pa.the_geom WHEN true THEN 2 else 0 end) as lockin_status
```
aagm, line 34: as we discussed yesterday we need to separate pu lock status from wdpa area calculation. in this query it was together

hotzevzl: sure - is this what you had in mind? ab93e4e0ebbd26b0a2617d0d4bd995b14157bafc

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-protection-status-calculator-service.ts
```diff
@@ -0,0 +1,57 @@
...
+     * there's nothing to do. We allow to set these on update, so we should not
+     * enforce this here.
+     *
+     * @TODO We'll need to handle custom protected areas here once we add
+     * support for these.
+     */
+    if(isNil(scenario.wdpaIucnCategories) || isNil(scenario.wdpaThreshold)) return;
+
+    const wdpaIucnCategoriesForScenario = scenario.wdpaIucnCategories?.map(i => `'${i}'`).join(', ');
+    console.log(wdpaIucnCategoriesForScenario);
```
kgajowy, line 36:
```suggestion
```

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-protection-status-calculator-service.ts
kgajowy: Fancy creating a relevant task in tracking system?

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-protection-status-calculator-service.ts
```diff
@@ -0,0 +1,57 @@
...
+    inner join planning_units_geom pug on spd.pu_geom_id = pug.id
+    where scenario_id='${scenario.id}'),
+    pu_pa as (select pu.id, st_area(st_intersection(pu.the_geom, pa.the_geom)) as pa_pu_area, (CASE pu.the_geom && pa.the_geom WHEN true THEN 2 else 0 end) as lockin_status
+              from pu
+              left join pa on pu.the_geom && pa.the_geom) 
+    UPDATE scenarios_pu_data
+    SET (lockin_status) =
+        (SELECT lockin_status FROM (select id, max(lockin_status) as lockin_status
+                                                    from pu_pa group by id) as result
+         WHERE scenarios_pu_data.id = result.id) where scenario_id = '${scenario.id}';
```
kgajowy, line 52: Shouldn't we use parameters instead? Even tho we are using `Scenario` entity there.

On the second thought, we shouldn't need to pass whole `Scenario` instance.

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-protection-status-calculator-service.ts
```diff
@@ -0,0 +1,57 @@
...
+    `;
+    Logger.debug(query);
```
kgajowy, line 54:
```suggestion
```

api/apps/api/src/modules/scenarios/scenarios.service.ts
```diff
@@ -88,13 +90,16 @@ export class ScenariosService {
...
+    await this.planningUnitsStatusCalculatorService.calculatedProtectionStatusForPlanningUnitsIn(scenario);
     return scenario;
   }
 
   async update(scenarioId: string, input: UpdateScenarioDTO) {
     await this.assertScenario(scenarioId);
     const validatedMetadata = this.getPayloadWithValidatedMetadata(input);
-    return this.crudService.update(scenarioId, validatedMetadata);
+    const scenario = await this.crudService.update(scenarioId, validatedMetadata);
+    await this.planningUnitsStatusCalculatorService.calculatedProtectionStatusForPlanningUnitsIn(scenario);
```
kgajowy, line 101: As long as updating (or creating, as above) scenario requires more actions than calling a single crud method (especially, when both are linked together), the relevant change should rather be conducted in standalone service (or inside CRUD service)

api/apps/api/src/modules/scenarios/scenarios.service.ts
```diff
@@ -88,13 +90,16 @@ export class ScenariosService {
...
+    return scenario;
```
kgajowy, line 102: Another thing is that if getting service is ever extended with returning PU as well, it would return invalid (old - before calculations) data.

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-protection-status-calculator-service.ts
```diff
@@ -0,0 +1,57 @@
...
+      DbConnections.geoprocessingDB,
+    )
+    private readonly puRepo: Repository<ScenariosPlanningUnitGeoEntity>,
+  ) {}
+
+  /**
+   * Calculate default lock status of planning units given their intersection
+   * with protected areas.
+   */
+  async calculatedProtectionStatusForPlanningUnitsIn(scenario: Scenario): Promise<void> {
```
kgajowy, line 24: tests would be great addition (for both logic /unit/ and e2e/integration to verify it actually changes desired columns.

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-protection-status-calculator-service.ts
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * If no protected areas are defined, or no threshold has been defined,
+     * there's nothing to do. We allow to set these on update, so we should not
+     * enforce this here.
```
kgajowy, line 28: Could be a private method like `shouldSetDefaultLockingStatus` to express this comment.

kgajowy (changes requested):

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-protection-status-calculator-service.ts
aagm: this can be made easier as if we are separating wdpa area calculation from lockin status so i would suggest:
```query
with pu as (
    select spd.id,  (pug.protected_area/spd.area)*100 as perc_protection
    from scenarios_pu_data spd
    inner join planning_units_geom pug on spd.pu_geom_id = pug.id
    where scenario_id='${scenario.id}'),
    pu_pa as (select pu.id, (CASE pu.perc_protection > ${threshold} WHEN true THEN 2 else 0 end) as lockin_status
from pu) 
UPDATE scenarios_pu_data
    SET (lockin_status) =
        (SELECT lockin_status FROM pu_pa
         WHERE scenarios_pu_data.id = pu_pa.id) where scenario_id = '${scenario.id}'; 
```

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-protection-status-calculator-service.ts
hotzevzl: https://vizzuality.atlassian.net/browse/MARXAN-649

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-protection-status-calculator-service.ts
hotzevzl: done in 51fb372c282d7b5244bf5b2c7004e3eee37d4349

hotzevzl: (left some comments in any case because memory 🤷🏼)

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-protection-status-calculator-service.ts
hotzevzl: yes, Bobby Tables 😬 965f0d80fabe9d6e83c0682342975fddfa934d63

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-protection-status-calculator-service.ts
hotzevzl: thanks! 965f0d80fabe9d6e83c0682342975fddfa934d63

api/apps/api/src/modules/scenarios/scenarios.service.ts
hotzevzl: I'd leave this for a later pass for practical reasons, but how would your suggestion be different from the current implementation that uses a distinct service to set the inclusion status?

api/apps/api/src/modules/scenarios/scenarios.service.ts
hotzevzl: you are right - we can ignore this case right now as it's not happening in practice, but I guess a quick way out would be to keep the update as is now and fetch the scenario again after calculating the default inclusion status.

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-protection-status-calculator-service.ts
hotzevzl: another item for second pass, sorry 😢

api/apps/api/src/modules/scenarios/scenarios.service.ts
kgajowy: @hotzevzl the solely purpose of the current service is to be a kind of facade with ACL on top of it (i.e. entry point for module consumer and shared ACL checks), thus being mostly dummy, it has a decent piece of code.

However, the main point of separation is to contain a full logic of create/update in a single place - to both having common sense what creating means together with what are the side effects and required actions.

PR #388, opened 2021-08-02
Add initial Deno API bot

hotzevzl:
See https://vizzuality.atlassian.net/browse/MARXAN-622

This is a tiny bot script that creates org/proj/scenario with some hardcoded settings (including credentials, sorry - it's very MVP), expecting to find API on http://localhost:3030. Edit script if you want to customize things.

The gist is, testing the viability of this route as one of the plans for cloning demo projects (as well as the viability of using Deno for this).

To test:

* download Deno https://github.com/denoland/deno/releases/tag/v1.12.2
* `deno run --allow-net data/bots/scenario-hero/bot.ts`


kgajowy (commented):
Not sure if we need that so much but its up to you I believe.

1 Why Deno tho?
2 Why not just some insomnia/postman collections with related queries?

image
image

hotzevzl:
@kgajowy yep, what you suggest or even an in-app script would do (and an in-app script could even piggyback on proper typing) - but I don't know where and how we may end up running this stuff, so having something that can be run with a single binary from anywhere seemed like a safer bet 🤷🏼

kgajowy:
> @kgajowy yep, what you suggest or even an in-app script would do (and an in-app script could even piggyback on proper typing) - but I don't know where and how we may end up running this stuff, so having something that can be run with a single binary from anywhere seemed like a safer bet 🤷🏼

Just keep in mind that you need to adjust it each time something changes (the scripts). If it doesn't act as CLI directly to existing code, its error tolerance isn't quite high.

With postman/insomnia the only added benefit is that you base on existing swagger specification (so its with sync) and also show how clients consume it. 
Also, could be used for testing purposes.

Not pushing tho, just worth to know other options.

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-protection-status-calculator-service.ts
```diff
@@ -21,7 +21,9 @@ export class ScenarioPlanningUnitsProtectedStatusCalculatorService {
    * Calculate default lock status of planning units given their intersection
    * with protected areas.
    */
-  async calculatedProtectionStatusForPlanningUnitsIn(scenario: Scenario): Promise<void> {
+  async calculatedProtectionStatusForPlanningUnitsIn(
```
kgajowy, line 24: needs rebase?

aagm (approved):
After preparing the seed data for the demos i think i will need to set up the demo proyects from here right?

hotzevzl:
> After preparing the seed data for the demos i think i will need to set up the demo proyects from here right?

that's the idea - I can help with this... keep in mind this is a proof of concept and an aid for my own checks so far, a mini version of key e2e tests but we don't need to preload demo projects this way: let's assess quickly how we want to do this once we have at least one project more or less running in the platform, ok?

PR #390, opened 2021-08-02
add support for saving scenario UI editing progress

hotzevzl:
https://vizzuality.atlassian.net/browse/MARXAN-628


hotzevzl:
merging as discussed with Alicia 😉

PR #397, opened 2021-08-03
improve puvspr.dat queries

hotzevzl:
as discussed with Alicia - implementation based on view (which I have kept for the moment) does not allow (as far as I can think...) to prefilter by `scenarioId` - with the implementation proposed in this PR I have been able to get query times down from ~4min with test data to ~20secs (most of the deal is actually indexes, but still, we also filter out data we won't need)

### DIFF
--- a/api/apps/api/src/modules/scenarios/input-files/puvspr.dat.service.ts
+++ b/api/apps/api/src/modules/scenarios/input-files/puvspr.dat.service.ts
@@ -1,27 +1,37 @@
-import { Repository } from 'typeorm';
+import { getConnection, Repository } from 'typeorm';
 import { Injectable } from '@nestjs/common';
-import { InjectRepository } from '@nestjs/typeorm';
-
-import { ScenarioPuvsprGeoEntity } from '@marxan/scenario-puvspr';
 import { DbConnections } from '@marxan-api/ormconfig.connections';
 
 @Injectable()
 export class PuvsprDatService {
-  constructor(
-    @InjectRepository(ScenarioPuvsprGeoEntity, DbConnections.geoprocessingDB)
-    private readonly puvsprRepo: Repository<ScenarioPuvsprGeoEntity>,
-  ) {}
-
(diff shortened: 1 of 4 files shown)


kgajowy (changes requested):
PR seems to mix three different ones (deno bot, calculating protection status and improving puvspr dat). Would be great to clean it up first before jumping into review. Thanks!

aagm (approved):

api/apps/api/src/modules/scenarios/input-files/puvspr.dat.service.ts
```diff
@@ -1,27 +1,48 @@
...
 import { DbConnections } from '@marxan-api/ormconfig.connections';
-
+import { InjectConnection } from '@nestjs/typeorm';
 @Injectable()
 export class PuvsprDatService {
   constructor(
-    @InjectRepository(ScenarioPuvsprGeoEntity, DbConnections.geoprocessingDB)
-    private readonly puvsprRepo: Repository<ScenarioPuvsprGeoEntity>,
+    @InjectConnection(DbConnections.geoprocessingDB)
+    private readonly connection: Connection,
```
kgajowy, line 9: you can just use `puvsprRepo.query` instead creating new connection.

hotzevzl:
> PR seems to mix three different ones (deno bot, calculating protection status and improving puvspr dat). Would be great to clean it up first before jumping into review. Thanks!

sorry about the PR stack mess - this is properly isolated now 😬

api/apps/api/src/modules/scenarios/input-files/puvspr.dat.service.ts
hotzevzl: yep, I know - I did so because with this PR the `puvsprRepo` should become redundant and we should be able to drop it altogether: I left it in the current PR in order not to change too much in case the new approach ends up not working out in practice, but all going well we should be able to drop the repo.

kgajowy (approved):

PR #398, opened 2021-08-03
Community page redesign

pedrogpimenta:
## Applied design changes to Community page

@mbarrenechea I removed the fixed widths for the links with icon. On the community page they are different, so I just made them the same, the way it's more logical front-end wise, as you said!

I also added `transition` to all buttons, tell me if it shouldn't be like that (I think it adds a nicer hover effect with minimum effort)


app/components/button/component.tsx
```diff
@@ -70,7 +70,7 @@ function buildClassName({
   theme,
 }) {
   return cx({
-    'group relative flex items-center justify-center rounded-4xl focus:outline-none': true,
+    'group relative flex items-center justify-center rounded-4xl transition focus:outline-none': true,
```
mbarrenechea, line 73: What if we use `transition-colors` => https://tailwindcss.com/docs/transition-property#class-reference. If we only want to animate colors and backgrounds I think it has better performance. TBH, we won't see any difference, but...

app/components/icon/component.tsx
```diff
@@ -18,7 +19,7 @@ export const Icon: React.FC<IconProps> = ({
   return (
     <svg
       className={cx({
-        'fill-current': true,
+        'fill-current transition': true,
```
mbarrenechea, line 22: Same here

app/layout/community/published-projects/duplicate-button/component.tsx
```diff
@@ -71,12 +75,17 @@ export const DuplicateButton: React.FC<DuplicateButtonProps> = ({
...
       className="px-6 group"
       size="s"
-      theme="transparent-white"
+      theme={theme === 'light' ? 'transparent-white' : 'transparent-black'}
       onClick={onDuplicate}
     >
       Duplicate
       <Icon
-        className="w-3.5 h-3.5 ml-2 text-white group-hover:text-black"
+        // className="w-3.5 h-3.5 ml-2 text-white group-hover:text-black"
```
mbarrenechea, line 83: Unnecessary comment

app/layout/community/published-projects/duplicate-button/component.tsx
```diff
@@ -71,12 +75,17 @@ export const DuplicateButton: React.FC<DuplicateButtonProps> = ({
     <Button
       className="px-6 group"
       size="s"
-      theme="transparent-white"
+      theme={theme === 'light' ? 'transparent-white' : 'transparent-black'}
```
mbarrenechea, line 78: You can use classnames package for anything you want. I probably hate ternaries...
```suggestion
      theme={classnames({
        'transparent-white': theme === 'light',
        'transparent-black': theme !== 'light' 
      })}
```

app/layout/community/published-projects/detail/share/component.tsx
```diff
@@ -54,15 +54,15 @@ export const ProjectShare: React.FC<ProjectShareProps> = () => {
 
   return (
     <div>
-      <h3 className="mb-6 text-sm font-semibold text-white">Share</h3>
+      <h3 className="mb-6 text-sm font-semibold">Share</h3>
 
       <div className="space-y-5">
         <button
```
mbarrenechea, line 60: image

Can you add some hover for these links? Underline them when you hover

mbarrenechea (changes requested):
The community project detail page doesn't have the light header
image

Could you remove or put the divide border lighter?
image

mbarrenechea (approved):

PR #412, opened 2021-08-05
feat(specification): determine features application logic

kgajowy:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/specification/determine-features/application/determine-features-for-scenario.handler.ts
+++ b/api/apps/api/src/modules/specification/determine-features/application/determine-features-for-scenario.handler.ts
@@ -1,0 +1,40 @@
+import {
+  CommandHandler,
+  EventPublisher,
+  IInferredCommandHandler,
+} from '@nestjs/cqrs';
+import { Logger } from '@nestjs/common';
+
+import { SpecificationRepository } from './specification.repository';
+import { DetermineFeaturesForSpecification } from '../../commands/determine-features-for-scenario-specification.command';
+
+@CommandHandler(DetermineFeaturesForSpecification)
+export class DetermineFeaturesForScenarioHandler
+  implements IInferredCommandHandler<DetermineFeaturesForSpecification> {
+  private readonly logger: Logger = new Logger(
+    DetermineFeaturesForSpecification.name,
+  );
+
+  constructor(
+    private readonly eventPublisher: EventPublisher,
+    private readonly specificationRepository: SpecificationRepository,
+  ) {}
+
+  async execute({
+    featuresConfig,
+    specificationId,
+  }: DetermineFeaturesForSpecification): Promise<void> {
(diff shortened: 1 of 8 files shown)


api/apps/api/src/modules/specification/determine-features/application/determine-features-for-scenario.handler.ts
```diff
@@ -0,0 +1,40 @@
...
+      specificationId,
+    );
+    if (!specification) {
+      this.logger.warn(`Couldn't find specification: ${specificationId}`);
+      return;
+    }
+    specification = this.eventPublisher.mergeObjectContext(specification);
+
+    specification.determineFeatures([featuresConfig]);
+    await this.specificationRepository.save(specification);
```
Dyostiq, line 37: shouldn't the specification be found by features ids? Also, should we use a transaction/pesimistic lock here on specifications?

api/apps/api/src/modules/specification/determine-features/application/specification.repository.ts
```diff
@@ -0,0 +1,7 @@
+import { Specification } from '../../specification';
```
Dyostiq, line 1: the application is chunked too much in my opinion

this import probably leads to circular dependency

it should be:

```
specification/domain
specification/application/determine-feature
```
and so on

Dyostiq (changes requested):
please resolve the potential circular dep

api/apps/api/src/modules/specification/determine-features/application/determine-features-for-scenario.handler.ts
kgajowy: 1 I decided to separate the `find by feature ids` to standalone application as it will be reused by both `determineFeatures` and `calculateFeatures`
2) we could, do you think there is a high chance it will clash?

Dyostiq: 1. what standalone application do you mean? The repo Is a driven port here, so the repository implementation is not a part of application but it's the adapter
2. when the queue will calculate features one by one then yes, it may cause lost updates

kgajowy: 1. Sorry, I meant standalone handler using different (repo)port which will dispatch commands for each found scenario (spec)

api/apps/api/src/modules/specification/determine-features/application/specification.repository.ts
kgajowy: fixed in https://github.com/Vizzuality/marxan-cloud/pull/417/files via https://github.com/Vizzuality/marxan-cloud/pull/417/commits/94946b083a4e324560dcb1e72bdd94f5f4b46799

PR #413, opened 2021-08-05
selection from geojson over pick by

hotzevzl:
https://vizzuality.atlassian.net/browse/MARXAN-537?focusedCommentId=12010

also includes a check for https://vizzuality.atlassian.net/browse/MARXAN-651


api/apps/geoprocessing/src/modules/scenario-planning-units-inclusion/__mocks__/include-sample.ts
```diff
@@ -53,6 +53,62 @@ export const excludeSample = (): FeatureCollection<Polygon | MultiPolygon> => ({
 });
 
 export const includeSample = (): FeatureCollection<Polygon | MultiPolygon> => ({
+  type: 'FeatureCollection',
```
kgajowy, line 56: Please update the image related to sample, otherwise it will be really hard to understand what changes/is going on.

api/apps/geoprocessing/src/modules/scenario-planning-units-inclusion/scenario-planning-units-inclusion-processor.ts
```diff
@@ -121,9 +121,7 @@ export class ScenarioPlanningUnitsInclusionProcessor
 
     if (
       new Set(
-        [...uniquePuIdsToInclude].filter((i) =>
-          uniquePuIdsToExclude.has(i),
-        ),
+        [...uniquePuIdsToInclude].filter((i) => uniquePuIdsToExclude.has(i)),
```
kgajowy, line 124: The "algorithm" started to be a bit complicated, thus I would recommend extracting the selection based on inputs  (4x arrays) to standalone class with named methods so that future us could get what is going us a bit faster.

api/apps/geoprocessing/src/modules/scenario-planning-units-inclusion/scenario-planning-units-inclusion-processor.ts
```diff
@@ -59,13 +84,50 @@ export class ScenarioPlanningUnitsInclusionProcessor
...
+      ],
+    );
+    puIdsToExclude.push(
+      ...[
+        ...puIdsToExcludeFromIdsLessIdsToIncludeFromGeo,
+        ...puIdsToExcludeFromGeo,
+      ],
+    );
+    const uniquePuIdsToInclude = new Set(puIdsToInclude);
+    const uniquePuIdsToExclude = new Set(puIdsToExclude);
```
Dyostiq, line 120: it looks like there is a high computational complexity here, especially with creating a set in every iteration of the filter.

I'd suggest to try `difference` method from lodash, and convert it to a set at the end.
So:
```
const puIdsToIncludeFromIdsLessIdsToExcludeFromGeo = difference(puIdsToIncludeFromIds, puIdsToExcludeFromGeo);
const puIdsToExcludeFromIdsLessIdsToIncludeFromGeo = difference(puIdsToExcludeFromIds, puIdsToIncludeFromGeo);
```
and
```
const doInclusionAndExclusionIntersects = intersection(uniquePuIdsToInclude, uniquePuIdsToExclude).length > 0
```

Dyostiq (approved):

api/apps/geoprocessing/src/modules/scenario-planning-units-inclusion/scenario-planning-units-inclusion-processor.ts
hotzevzl: 👍🏼 more or less done in aed5b088add61ffa63805f29e8dd6cd2e8d89084

api/apps/geoprocessing/src/modules/scenario-planning-units-inclusion/scenario-planning-units-inclusion-processor.ts
hotzevzl: I agree - I decided to keep things here - I think that the use of lodash set operators as advised by Dominik and the few comments in the source could do... given the circumstances, at least 😱

api/apps/geoprocessing/src/modules/scenario-planning-units-inclusion/__mocks__/include-sample.ts
hotzevzl: good point. 501d1145f046337c25a92b37cafc27fc7f9a438c

PR #423, opened 2021-08-09
MARXAN-545 MARXAN-644 specification repository

kgajowy:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


api/apps/api/src/modules/specification/adapters/specification.repository.ts
```diff
@@ -0,0 +1,178 @@
...
+                featureId: feature.id,
+              }),
+            ),
+            featuresDetermined: configuration.featuresDetermined,
+            specificationId: snapshot.id,
+          })),
+        ),
+      }),
+    );
+    return;
```
Dyostiq, line 137:
```suggestion
```

api/apps/api/src/modules/api-events/api-event.api.entity.ts
```diff
@@ -69,7 +69,7 @@ export class ApiEvent {
    */
   @ApiProperty()
   @IsEnum(Object.values(API_EVENT_KINDS))
-  @Column('enum')
+  @Column('enum', { enum: API_EVENT_KINDS })
```
Dyostiq, line 72: not true, it's `varchar`

api/apps/api/src/modules/specification/adapters/specification-feature.api.entity.ts
```diff
@@ -0,0 +1,40 @@
+import {
+  Column,
+  Entity,
+  JoinColumn,
+  ManyToOne,
+  PrimaryGeneratedColumn,
+} from 'typeorm';
+import { SpecificationFeatureConfigApiEntity } from './specification-feature-config.api.entity';
+
+@Entity(`specification_feature`)
```
Dyostiq, line 10: the kept convention is to use plurals

Dyostiq (approved):

api/apps/api/src/migrations/api/1628585506694-ScenarioSpecification.ts
```diff
@@ -0,0 +1,47 @@
...
+      `CREATE TABLE "specification_feature" ("id" uuid NOT NULL DEFAULT uuid_generate_v4(), "specification_feature_config_id" uuid NOT NULL, "feature_id" uuid NOT NULL, "calculated" boolean NOT NULL, CONSTRAINT "PK_fe6096ea6f25997b62b8a5c5f99" PRIMARY KEY ("id"))`,
+    );
+    await queryRunner.query(
+      `CREATE TYPE "specification_feature_config_operation_enum" AS ENUM('split', 'stratification', 'copy')`,
+    );
+    await queryRunner.query(
+      `CREATE TABLE "specification_feature_config" ("id" uuid NOT NULL DEFAULT uuid_generate_v4(), "specification_id" uuid NOT NULL, "base_feature_id" uuid NOT NULL, "against_feature_id" uuid, "operation" "specification_feature_config_operation_enum" NOT NULL, "features_determined" boolean NOT NULL, CONSTRAINT "PK_df1e25cf3972f83b1ff5bcb0337" PRIMARY KEY ("id"))`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "specification" ADD CONSTRAINT "FK_2cecab1b81e5fa64f2fdbbaec76" FOREIGN KEY ("scenario_id") REFERENCES "scenarios"("id") ON DELETE CASCADE ON UPDATE NO ACTION`,
```
hotzevzl, line 20: should we cascade on update too? probably moot in practice as we don't support updating `scenarios.id`, but i'd say it doesn't hurt to have the cascade (same for the other FKs below)

api/apps/api/src/modules/specification/adapters/specification-feature.api.entity.ts
hotzevzl: Since this is a full entity now, we will also need an `order` (or something like that) column to keep the ordering of features in a specification consistent with what API clients send (or the frontend to start with may see features jumping around depending on contingent factors such as order of insert/update in db)

hotzevzl (approved):

api/apps/api/src/modules/api-events/api-event.api.entity.ts
kgajowy: Could be that actual state isn't matched to what we have in entity declaration - still wanted a short route just o keep migration-generation to return results.

Dyostiq: then it's better to update it to `varchar` :D

api/apps/api/src/modules/specification/adapters/specification-feature.api.entity.ts
kgajowy: @hotzevzl sorting on serializer/controller level should be enough, shouldn't it?

if it is needed at all (https://vizzuality.slack.com/archives/C01G7PRQB96/p1628593316134400)

api/apps/api/src/migrations/api/1628585506694-ScenarioSpecification.ts
kgajowy: TBH we should rather avoid doing such operations directly on the entity if we want to keep the system consistent.

api/apps/api/src/modules/specification/adapters/specification-feature.api.entity.ts
hotzevzl: ok, so if we are echoing back (with any augmentation) the specification as received in the request payload then we don't need to keep an order of the children of `specification_features` as stored in db.

if we are composing the response from the data we have in db, then keeping the same order of geofeatures as provided in the request payload would be preferable for a consistent user experience - we could sort at serializer level, but without other order hints we could only order alphabetically here, unless I am misunderstanding what you suggest.

api/apps/api/src/migrations/api/1628585506694-ScenarioSpecification.ts
hotzevzl: yes, we won't be updating UUID ids in general (if that's what you mean here). In general I see no reason for *not* having cascades on both delete and update unless there is a specific case for keeping FKs constant despite changes (for whatever reason) in the referenced IDs or a need for micro optimizations in terms of performance in specific contexts/use cases

hotzevzl: but yes, since we do have a peculiar setup with non-really-FKs across dbs, I agree that in this case restricting the update may be preferable, so ok.

PR #443, opened 2021-08-12
feat(scenario-specification): allow to get last updated specification

kgajowy:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/scenarios/scenarios.controller.ts
+++ b/api/apps/api/src/modules/scenarios/scenarios.controller.ts
@@ -17,6 +17,7 @@ import {
   UseInterceptors,
   ValidationPipe,
   InternalServerErrorException,
+  NotFoundException,
 } from '@nestjs/common';
 import { scenarioResource, ScenarioResult } from './scenario.api.entity';
 import { Request, Response } from 'express';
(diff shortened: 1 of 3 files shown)


kgajowy:
@Dyostiq / @hotzevzl please note the TODOs  in `scenario.controller` - at the point we have flow more (or less) ready, those endpoints should call new implementation.

hotzevzl:
> @Dyostiq / @hotzevzl please note the TODOs in `scenario.controller` - at the point we have flow more (or less) ready, those endpoints should call new implementation.

yep. I suggest to let Miguel know about the new endpoint, request DTO and any difference in behaviour (poll for status, etc.) so that he can prepare the frontend, and then switch over at the same time-ish in both API and frontend.

Can I already let Miguel know about interfaces and so on, to start preparing the transition to the full implementation, or do we still need to finish bits in terms of interface? (unfinished implementation is ok of course).

kgajowy:
> > @Dyostiq / @hotzevzl please note the TODOs in `scenario.controller` - at the point we have flow more (or less) ready, those endpoints should call new implementation.
> 
> yep. I suggest to let Miguel know about the new endpoint, request DTO and any difference in behaviour (poll for status, etc.) so that he can prepare the frontend, and then switch over at the same time-ish in both API and frontend.
> 
> Can I already let Miguel know about interfaces and so on, to start preparing the transition to the full implementation, or do we still need to finish bits in terms of interface? (unfinished implementation is ok of course).

Interfaces remain (should remain) the same once we do the switch (or can just use "v2" version for both POST/PUT)

Regarding pooling status, @Dyostiq will know a bit more how to verify that this very step is finished.

PR #445, opened 2021-08-12
Feature shapefile upload

ikas:
### Overview

This PR adds a new endpoint `POST /api/v1/projects/:id/features/shapefile` that allows users to upload feature shapefiles.

To use this endpoint, you'll need to provide a `.zip` file containing the different format files that compose a shapefile, and send it as `multipart/form-data` to the endpoint specified above.

The upload happens synchronously, so once you receive the response, the data should already be available in the database. Also, keep in mind there's a limit on the maximum size of the uploaded file, currently set to 100MB (this limit is configurable via env vars).

### Designs

[Figma](https://www.figma.com/file/hq0BZNB9fzyFSbEUgQIHdK/Marxan-Visual_V02?node-id=2991%3A2492)

_This PR does not contain the implementation of the UI for this feature, just the BE side of things._

### Testing instructions

Use the new endpoint and check that data gets correctly imported to the databases (API and geo) - check the JIRA ticket below for more details on where the data should be placed.

### Feature relevant tickets

[MARXAN-543](https://vizzuality.atlassian.net/browse/MARXAN-543)

---

- [ ] Update CHANGELOG file


api/apps/api/src/modules/geo-features/geo-feature.api.entity.ts
```diff
@@ -18,7 +19,7 @@ export interface GeoFeatureProperty {
 }
 
 @Entity('features')
-export class GeoFeature {
+export class GeoFeature extends BaseEntity {
```
ikas, line 22: had to make this to be able to access the `id` of the entity after inserting it into the DB

api/apps/api/src/modules/geo-features/geo-features.service.ts
```diff
@@ -258,4 +269,91 @@ export class GeoFeaturesService extends AppBaseService<
   ): Promise<GeoFeature> {
     return entity;
   }
+
+  public async createFeaturesForShapefile(
```
ikas, line 273: transactions here are little bit sketchy, since I'm writing data to 2 separate DBs :thinking: if you think of a better solution let me know :smile:

api/apps/api/src/modules/projects/projects.controller.ts
```diff
@@ -245,4 +250,37 @@ export class ProjectsController {
...
+  @ApiOperation({
+    description: `Upload shapefiles of species or bioregional features.`,
+  })
+  @ApiOkResponse({ type: ShapefileUploadResponse })
+  @Post(`:id/features/shapefile`)
+  @UseInterceptors(
+    FileInterceptor('shapefile', {
+      ...uploadOptions,
+      limits: {
+        // TODO: fix hardcoded value
```
ikas, line 264: fixing this right now

api/apps/api/src/modules/geo-features/geo-features.service.ts
hotzevzl: yes, it occurred to me with a huge 🤦🏼 when I saw your PR while trying to help with the makevalid bit - I think what you do here will be fine in most circumstances. if we do run into issues in a substantive way I'd think about a final check both in case of success or failure, but since I'm thinking about a grim reaper/garbage collection feature that we will need in any case to be able to invalidate stuff when users edit settings earlier on in the workflow, this - if needed at all - could be done when we implement that.

api/apps/api/src/modules/projects/projects.controller.ts
ikas: fixed!

hotzevzl (approved):
Thanks @ikas! Epic PR, thanks so much for your dedication this week 🙇🏼 

I'd ask @aagm to please cross check that there are no blind spots (which, if any, would mainly be due to blind spots in my specification, tbh) before merging, though I'm itching to merge RIGHT NOW so that I can use this feature via bots I'm working on 😅

hotzevzl:
oh well I hadn't notice the test failures - though they all seem to be identical and only due to an import lookup issue, so this should be easy to fix (if you get stuck, probably Dominik can help fixing this while blindfolded)

api/apps/api/src/modules/projects/projects.controller.ts
```diff
@@ -66,8 +70,10 @@ export class ProjectsController {
   constructor(
     private readonly projectsService: ProjectsService,
     private readonly geoFeatureSerializer: GeoFeatureSerializer,
+    private readonly geoFeatureService: GeoFeaturesService,
     private readonly projectSerializer: ProjectSerializer,
     private readonly jobsStatusSerizalizer: JobStatusSerializer,
+    private readonly shapefileService: ShapefileService,
```
Dyostiq, line 76: It should not be used here.

It's a service from the geo app, and we should not share anything between the app `as is` but by shared libs.
The second thing is the service uses mapshaper under the hood, and we wanted to keep heavy calculation on the geo side – with this, we're breaking the rule.

api/apps/api/src/modules/geo-features/geo-feature.api.entity.ts
```diff
@@ -56,6 +57,9 @@ export class GeoFeature {
   @Column('varchar')
   tag!: FeatureTag;
 
+  @Column('varchar', { name: 'creation_status' })
+  creationStatus?: JobStatus;
```
kgajowy, line 61: Why is this field necessary? Actual status usually come from ApiEvents (and in this case, should rather be related to some Scenario* event)

api/apps/api/src/modules/geo-features/geo-feature.geo.entity.ts
```diff
@@ -11,12 +12,29 @@ export const geoFeatureResource: BaseServiceResource = {
...
 export class GeoFeatureGeometry {
   @ApiProperty()
   @PrimaryColumn()
   id!: string;
 
+  @Column('geometry', { name: 'the_geom' })
+  theGeom?: Geometry;
+
+  @Column('jsonb')
+  properties?: Record<string, string | number>;
```
kgajowy, line 33: Are we able to narrow the type?

api/apps/api/src/modules/geo-features/geo-feature.geo.entity.ts
```diff
@@ -11,12 +12,29 @@ export const geoFeatureResource: BaseServiceResource = {
...
+
+  @Column('enum')
+  source?: SourceType;
```
kgajowy, line 36: It will cause issues when generating migrations. 
```
@Column(`enum`, { enum: SourceType, nullable: strue })
source?: SourceType | null
```

api/apps/api/src/modules/geo-features/geo-features.service.ts
```diff
@@ -270,4 +280,91 @@ export class GeoFeaturesService extends AppBaseService<
...
+    projectId: string,
+    data: UploadShapefileDTO,
+    features: Record<string, any>[],
+  ): Promise<void> {
+    const [apiDbConnection, geoDbConnection] = Object.values(
+      DbConnections,
+    ).map((name) => getConnection(name));
```
kgajowy, line 291: [minor] not sure if iterating gives us any profit there (could be just simple `const apiDbConnection = getConnection(DbConnections.Api...)`

api/apps/api/src/modules/geo-features/geo-features.service.ts
```diff
@@ -270,4 +280,91 @@ export class GeoFeaturesService extends AppBaseService<
...
+
+      // Store geometries in features_data table
+      for (const feature of features) {
+        await this.createFeatureData(
+          geoQueryRunner.manager,
+          geofeature.id,
+          feature.geometry,
+          feature.properties,
+        );
+      }
```
kgajowy, line 318: How about doing those in `bulk`?

api/apps/api/src/modules/geo-features/geo-feature.api.entity.ts
hotzevzl: I think this was done to keep to the current table structure, but we could probably even drop this column and related entity property, especially if doing the insert synchronously as in this initial implementation.

api/apps/api/src/modules/geo-features/geo-feature.geo.entity.ts
hotzevzl: are you asking if we can further narrow it, or if the narrowing here is accurate?

if the latter, looking at the dbase IV spec, I think we may only be missing boolean. I advised on this specific typing because of the way we import features via ETL pipelines (where bools are imported as string), but it won't hurt to add boolean here if that's what you meant.

kgajowy: @hotzevzl I meant to provide more detailed structure rather than generic "everything can be there"

PR #447, opened 2021-08-13
feat(api): add copy operation

Dyostiq:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/scenarios-features/copy/copy-data-provider.service.ts
+++ b/api/apps/api/src/modules/scenarios-features/copy/copy-data-provider.service.ts
@@ -1,0 +1,43 @@
+import { Injectable } from '@nestjs/common';
+import { InjectEntityManager } from '@nestjs/typeorm';
+import { EntityManager } from 'typeorm';
+import { assertDefined } from '@marxan/utils';
+import { ProjectsCrudService } from '@marxan-api/modules/projects/projects-crud.service';
+import { Project } from '@marxan-api/modules/projects/project.api.entity';
+import { Scenario } from '@marxan-api/modules/scenarios/scenario.api.entity';
+import { CreateFeaturesCommand } from '../create-features.command';
+
+@Injectable()
+export class CopyDataProvider {
+  constructor(
+    @InjectEntityManager()
+    private readonly apiEntityManager: EntityManager,
+    private readonly projects: ProjectsCrudService,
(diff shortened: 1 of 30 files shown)


Dyostiq:
I know that tests fail :) but still the review is welcome

api/apps/api/src/modules/scenarios-features/copy/copy-data-provider.service.ts
```diff
@@ -0,0 +1,43 @@
...
+
+@Injectable()
+export class CopyDataProvider {
+  constructor(
+    @InjectEntityManager()
+    private readonly apiEntityManager: EntityManager,
+    private readonly projects: ProjectsCrudService,
+  ) {}
+
+  async prepareData(
```
kgajowy, line 18: ~ `getProjectSetup` ?

kgajowy: (as this piece will be most likely shared across all operations)

api/apps/api/src/modules/scenarios-features/copy/copy-operation.service.ts
```diff
@@ -0,0 +1,36 @@
...
+  ) {}
+
+  async copy(command: CreateFeaturesCommand): Promise<{ id: string }[]> {
+    const {
+      project,
+      protectedAreaFilterByIds,
+      planningAreaLocation,
+    } = await this.copyDataProvider.prepareData(command);
+    const { parameters, query } = this.copyQuery.prepareStatement(
+      command,
```
kgajowy, line 25: does it need the command itself?

kgajowy (approved):

api/apps/api/src/modules/scenarios-features/copy/copy-operation.service.ts
Dyostiq: the entire service could be in fact a command handler

hotzevzl:
@Dyostiq structure and strategy look ok to me.

Since this whole über-story is a fairly tricky one, I would suggest to log queries and parameters in this phase so we can keep an eye on what gets sent to db and we can more easily debug things, and we can then fully remove in a second pass (or just rely on log levels to omit these in production)

PR #452, opened 2021-08-16
WDPA: read them from scenario instead of store in certain cases

mbarrenechea:
### Overview

This PR adds:
- withScenario HOC to prefetch scenarios from server
- use wdpaIucnCategories and wdpaThreshold from scenario data

### DIFF
--- a/app/hooks/map/index.ts
+++ b/app/hooks/map/index.ts
@@ -240,7 +240,7 @@ export function usePUGridPreviewLayer({
 
 // PUGrid
 export function usePUGridLayer({
-  active, sid, type, subtype, runId, options = {}, cache,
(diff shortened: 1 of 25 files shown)


app/hooks/map/index.ts
```diff
@@ -419,7 +420,22 @@ export function usePUGridLayer({
...
             },
           ] : [],
-
+          ...type === 'solutions' && !!solution ? [
+            {
+              type: 'line',
+              'source-layer': 'layer0',
+              filter: [
+                'all',
+                ['in', ['get', 'scenarioPuId'], ['literal', solution]],
```
anamontiaga, line 429: This is not going to work, I was just trying

PR #477, opened 2021-08-23
Feat/published project map [MARXAN-683]

anamontiaga:
### Overview

This PR adds map of the first scenario runned - in case it exists - to the published project detail page.

### Designs

[Figma Designs](https://www.figma.com/file/hq0BZNB9fzyFSbEUgQIHdK/Marxan-Visual_V02?node-id=4671%3A7069)

### Feature relevant tickets

[MARXAN-683](https://vizzuality.atlassian.net/secure/RapidBoard.jspa?rapidView=38&projectKey=MARXAN&modal=detail&selectedIssue=MARXAN-683&quickFilter=71&assignee=60d985f51613d50069adeabc&assignee=UNASSIGNED_USER_ID)

### DIFF
--- a/app/hooks/map/index.ts
+++ b/app/hooks/map/index.ts
@@ -552,7 +552,7 @@ export function useLegend({
     if (type === 'analysis' && subtype === 'analysis-adjust-planning-units') return ['wdpa-percentage', 'lock-in', 'lock-out', 'pugrid'];
     if (type === 'analysis') return ['wdpa-percentage', 'features', 'pugrid'];
     if (type === 'solutions') return ['frequency', 'solution', 'pugrid'];
-    if (type === 'published') return ['wdpa-percentage', 'frequency', 'features', 'pugrid'];
+    if (type === 'published') return ['frequency', 'features', 'pugrid'];
 
     return ['pugrid'];
   }, [type, subtype, options]);
--- a/app/hooks/map/index.ts
+++ b/app/hooks/map/index.ts
@@ -284,7 +284,6 @@ export function usePUGridPreviewLayer({
 export function usePUGridLayer({
   active, sid, include, sublayers, options = {}, cache,
 }: UsePUGridLayer) {
-  console.log('ID', sid);
   return useMemo(() => {
     if (!active || !sid) return null;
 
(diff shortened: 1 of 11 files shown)


app/hooks/map/index.ts
```diff
@@ -552,7 +552,7 @@ export function useLegend({
     if (type === 'analysis' && subtype === 'analysis-adjust-planning-units') return ['wdpa-percentage', 'lock-in', 'lock-out', 'pugrid'];
     if (type === 'analysis') return ['wdpa-percentage', 'features', 'pugrid'];
     if (type === 'solutions') return ['frequency', 'solution', 'pugrid'];
-    if (type === 'published') return ['wdpa-percentage', 'frequency', 'features', 'pugrid'];
+    if (type === 'published') return ['frequency', 'features', 'pugrid'];
```
anamontiaga, line 555: I have a add a new type 'published' independent of the tabs to be able to customize what we see on the map.

app/layout/community/published-projects/detail/map/component.tsx
```diff
@@ -2,46 +2,57 @@ import React, {
...
+import LegendTypeMatrix from 'components/map/legend/types/matrix';
 
 export interface PublishedProjectMapProps {
 }
 
 export const PublishedProjectMap: React.FC<PublishedProjectMapProps> = () => {
-  const [viewport, setViewport] = useState({});
-  const [bounds, setBounds] = useState(null);
+  const [open, setOpen] = useState(false);
   const [session] = useSession();
```
anamontiaga, line 38: In the future we would need to work without session. At the moment i didn't manage to make Mapbox API works without sending auth .

app/hooks/map/index.ts
mbarrenechea: What we should do is the same we did with the layers. This "useMemo" to get the layers should come from outside, from where you are using the useLegend instead of from inside. That way we don't care if you are in the detail, the published page, or wherever you want. 

In other words, instead of passing type & subtype, you should pass directly an attribute called `layers` calculated where you use the `useLegend`

app/layout/community/published-projects/detail/map/component.tsx
mbarrenechea: You are right... the endpoint to get the layers is authenticated. We will ask to remove that

mbarrenechea (changes requested):

mbarrenechea (approved):

PR #480, opened 2021-08-24
feat(projects): find returns only projects created by user

kgajowy:
MARXAN-582 restrict list of projects returned to user to their own projects only

migration (instead of seeds) will have a followup PR (please don't merge this one now)..

Implementation details:

Options were:
1. Add relations and cascade create
    1. pros: transactional
    2. cons: project itself knowing about roles/privileges; possible issues with adding further roles ([…roles, newRole])
2. Extracting roles/privileges into standalone (sub)module
    1. pros: clear separation
    2. cons: would require fine amount of changes to keep it in transaction; we don’t know the full shape of rbac/acl
3. Attaching to ~BaseService’s `actionAfterCreate`
    1. pros: pretty straightforward, easy to „remove”; failed non-transactional insert into user_projects may leave orphaned projects
    2. cons: overextending responsibilities; not transactional 
    3. Actually not possible as ~BaseService does not await on afterCreate (which I believe is intentional and fine…)
4. Adding additional logic within projects.service.ts

### DIFF
--- a/api/apps/api/src/modules/projects/projects-crud.service.ts
+++ b/api/apps/api/src/modules/projects/projects-crud.service.ts
@@ -1,5 +1,5 @@
 import { forwardRef, Inject, Injectable } from '@nestjs/common';
-import { isDefined } from '@marxan/utils';
+import { assertDefined, isDefined } from '@marxan/utils';
 import { InjectRepository } from '@nestjs/typeorm';
 import { AppInfoDTO } from '@marxan-api/dto/info.dto';
 import { Repository, SelectQueryBuilder } from 'typeorm';
@@ -23,6 +23,8 @@ import {
   MultiplePlanningAreaIds,
   PlanningAreasService,
 } from './planning-areas';
+import { UsersProjectsApiEntity } from './control-level/users-projects.api.entity';
+import { Roles } from '@marxan-api/modules/users/role.api.entity';
 
 const projectFilterKeyNames = [
   'name',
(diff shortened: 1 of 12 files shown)


api/apps/api/src/modules/projects/projects-crud.service.ts
```diff
@@ -250,7 +264,15 @@ export class ProjectsCrudService extends AppBaseService<
     fetchSpecification: FetchSpecification,
     info?: ProjectsInfoDTO,
   ): Promise<SelectQueryBuilder<Project>> {
+    assertDefined(info?.authenticatedUser);
```
kgajowy, line 267: (I am aware that it will need to be re-implemented a bit for a `public` projects feature)

api/apps/api/src/modules/projects/projects-crud.service.ts
```diff
@@ -272,9 +294,32 @@ export class ProjectsCrudService extends AppBaseService<
       );
     }
 
+    query
+      .andWhere(`acl.user_id = :userId`, {
+        userId: info?.authenticatedUser?.id,
+      })
+      .andWhere(`acl.role_id = :roleId`, {
+        roleId: Roles.project_owner,
+      });
```
hotzevzl, line 303: I know this is what is asked in the story, but looking at the implementation, I wonder if we should drop the check of role and just list every project the user has a role on, as we won't be setting other roles right now (so in practice even without this check the end result will be the same as in the story), with the tiny benefit of not having to remove this when we add full ACL checks (although in any case we'll have to revisit a lot of things by then)... what do you think @kgajowy ?

kgajowy: Once we will want to add full ACL (or RBAC) checks, it will require a lot more changes I believe.

Either change this one and update documentation/requirements, or let it be. Up to you @hotzevzl

hotzevzl: yep, let's leave it as is.

hotzevzl (approved):

PR #504, opened 2021-08-27
fix(scenarios): wdpa filters

kgajowy:
image

image

### DIFF
--- a/api/apps/api/test/project/create-project-wdpa-area-filter.e2e-spec.ts
+++ b/api/apps/api/test/project/create-project-wdpa-area-filter.e2e-spec.ts
@@ -1,0 +1,30 @@
+import { FixtureType } from '@marxan/utils/tests/fixture-type';
+import { getFixtures } from './create-project-wdpa-area-filter.fixtures';
+import { IUCNCategory } from '@marxan/iucn';
+
+let fixtures: FixtureType<typeof getFixtures>;
+
+beforeEach(async () => {
+  fixtures = await getFixtures();
+});
+
+afterEach(async () => {
+  await fixtures?.cleanup();
+});
+
+test(`creating multiple projects should have different protectedAreaFilterByIds`, async () => {
+  const projectIdOne: string = await fixtures.WhenProjectIsCreated(`BWA`);
+  const scenarioIdOne: string = await fixtures.WhenScenarioIsCreated(
+    projectIdOne,
+    [IUCNCategory.NotApplicable],
+  );
+  const projectIdTwo: string = await fixtures.WhenProjectIsCreated(`ZMB`);
+  const scenarioIdTwo: string = await fixtures.WhenScenarioIsCreated(
+    projectIdTwo,
+    [IUCNCategory.NotReported],
+  );
+  await fixtures.ThenProtectedAreaFiltersAreDifferent(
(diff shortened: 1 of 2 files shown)


hotzevzl:
I am really unsure about this one - the tests are ok (consistently, for me), but testing things manually gives different results, and not always.

I have created one project in BWA, one in NAM, and got the same set of protected area ids (in fact the PAs visually seem to be all across the border between these countries, so even though unlikely that they're *all* across the border, this may even be).

Then one project in EGY, and this gave a different set of protected areas.

Then one in BRA, and this resulted in the same set of PA ids as the BWA and NAM ones. Another scenario within this project resulted to the same set as the previous scenario in BRA.

Then ESP, which resulted in a new set of ids.

Then BRA again, which resulted in the same set of ids as the previous one in ESP.

Then BRA again, same as previous one.

But whatever I try in the e2e test (rotating some of these countries in the test spec), the test always passes.

Interestingly (predictably?) the UI does show the protected areas when loading tiles for the country/study area when getting to the first step of protected area selection, but once I choose the relevant IUCN categories and move to step 2/2, only the PU grid appears, with no protected areas anymore (whereas when the set of ids varies when creating a new project, these are shown and I can see the threshold slider affecting the selection of PUs that will be considered "protected").

hotzevzl:
further to the previous notes, I have run a bot first with BRA as planning area (and it picked the same set of PA ids as the previous, UI-created Brazil projects/scenarios), then a bot with ESP as planning area, and this picked a different set of PA ids than anything else before. I'm officially 🤯 at this point 😬

kgajowy:
@hotzevzl so may be this issue is somehow related to UI (or tiles?)

kgajowy:
@hotzevzl or something happens in asynchronous way (as a side effect) for what, most likely, tests do not wait?

hotzevzl:
> @hotzevzl so may be this issue is somehow related to UI (or tiles?)

@kgajowy for a moment yesterday I hoped so, but the requests all make sense, it's the sets of calculated PA ids that don't, and the fact that they do so inconsistently (or at least according to a logic/pattern I haven't been able to understand yet) doesn't help. Not ruling this out, but seems unlikely at this stage.

As for side effects... could be, though even here, I haven't been able to identify _what_ may be happening, as after all the core query is a rather simple intersect between planning area and wdpa records 🤔

hotzevzl:
@kgajowy see if you think it's worth updating these tests to cover the update case, or what to do with this PR.

kgajowy:
> @kgajowy see if you think it's worth updating these tests to cover the update case, or what to do with this PR.

updated: https://github.com/Vizzuality/marxan-cloud/pull/504/commits/b375552b1534bc64cb271876c2c1d145cea71b7f

however, the spec is not that much reliable... sometimes if fails, sometimes passes fine (locally as well)

will rebase to develop once we can observer its fail on the CI (as in the current PR's base does not contain fix)

kgajowy:
Before rebase:
image

Dyostiq:
maybe `waitForExpect` could help if there is something asynchronous

kgajowy:
@hotzevzl after rebasing to develop (which includes fixes) seems to work fine.

hotzevzl (approved):

PR #516, opened 2021-09-01
fix(marxan-run): in case of any error, handle it and use failed-reason

kgajowy:
Reason for using `string` within `new Error(...)` is that `bullmq` assumes such format so that the job data contains `failedReason` property derived from this very type (i.e. if processor finished throwing an error)

source: https://github.com/taskforcesh/bullmq/issues/537#issuecomment-836601738

### DIFF
--- a/api/apps/geoprocessing/src/marxan-sandboxed-runner/marxan-run.ts
+++ b/api/apps/geoprocessing/src/marxan-sandboxed-runner/marxan-run.ts
@@ -37,13 +37,10 @@ export class MarxanRun extends MarxanRunEmitter implements Cancellable {
     this.#process = spawn(workspace.marxanBinaryPath, {
       cwd: workspace.workingDirectory,
     });
-
     assertDefined(this.#process);
-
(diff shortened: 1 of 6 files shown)


api/apps/geoprocessing/src/marxan-sandboxed-runner/marxan-run.ts
```diff
@@ -37,20 +37,36 @@ export class MarxanRun extends MarxanRunEmitter implements Cancellable {
...
+    this.#process.on(
+      `error`,
+      (
+        error: Error & {
+          errno: number;
+          code: number;
+          path: string;
+          syscall: string;
+          spawnargs: string[];
+        },
```
Dyostiq, line 60:
```suggestion
        error: Error | {
          errno: number;
          code: number;
          path: string;
          syscall: string;
          spawnargs: string[];
        },
```
otherwise the `code` in error doesn't make sense

Dyostiq (approved):

hotzevzl (approved):

PR #522, opened 2021-09-06
fix(protected-areas): project's custom pa shapefile

kgajowy:
It seems that original implementation (which was moved by me to current processor) was different than for for `custom-planning-area`. I replaced the old one with the same one as `custom-pa` used.


kgajowy:
@hotzevzl https://vizzuality.atlassian.net/browse/MARXAN-721 please verify if the projection is fine.

api/apps/geoprocessing/src/modules/protected-areas/worker/protected-area-processor.ts
```diff
@@ -1,60 +1,48 @@
...
-      await queryRunner.commitTransaction();
-    } catch (error) {
-      await queryRunner.rollbackTransaction();
-      throw error;
-    } finally {
-      await queryRunner.release();
-    }
+      await transaction.query(
+        `
+          INSERT INTO "wdpa"("the_geom", "project_id", "full_name")
```
aagm, line 33: We are returning the inserted id, right?

aagm (approved):

api/apps/geoprocessing/src/modules/protected-areas/worker/protected-area-processor.ts
kgajowy: @aagm I haven't changed the flow right now, rather just fix the bug https://vizzuality.atlassian.net/browse/MARXAN-721 

If we need to return the IDs (assuming we cannot query them later) we should consider new issue with described use case.

PR #525, opened 2021-09-07
feat(custom-pu-grid): custom pu grid geometry

kgajowy:
- unpacking shapefile and adding geometries - https://github.com/Vizzuality/marxan-cloud/pull/525 (this)
- exposing endpoint to submit & api events - https://github.com/Vizzuality/marxan-cloud/pull/527
- exposing endpoint to get the request status - https://github.com/Vizzuality/marxan-cloud/pull/528
- updating project `planning_unit_grid_shape` - https://github.com/Vizzuality/marxan-cloud/pull/529


api/apps/geoprocessing/src/modules/planning-units-grid/planning-units-grid.processor.ts
```diff
@@ -0,0 +1,76 @@
...
+        INSERT INTO "planning_units_geom"("the_geom", "type", "project_id")
+        SELECT ST_SetSRID(
+                 ST_GeomFromGeoJSON(features ->> 'geometry'),
+                 4326)::geometry,
+               $2,
+               $3
+        FROM (
+               SELECT json_array_elements($1::json -> 'features') AS features
+             ) AS f
+        ON CONFLICT ON CONSTRAINT planning_units_geom_the_geom_type_key DO UPDATE SET type = 'irregular'
```
aagm, line 56: ` type`  should be inputted from the FE

aagm (changes requested):
In terms of validation, it is part of the characteristic of a grid that their members are adjacent geometries. in order to validate it, they should not intersect with each other, but every single geometry should be [intouch](https://postgis.net/docs/ST_Touches.html) (connected) with at least one. If we want to do that validation with javascript, i think turf will allow you to intersect and check if the result is a line in order to ensure that they only share borders.

kgajowy:
> In terms of validation, it is part of the characteristic of a grid that their members are adjacent geometries. in order to validate it, they should not intersect with each other, but every single geometry should be [intouch](https://postgis.net/docs/ST_Touches.html) (connected) with at least one. If we want to do that validation with javascript, i think turf will allow you to intersect and check if the result is a line in order to ensure that they only share borders.

Can we validate this assumption with postgis?

api/apps/geoprocessing/src/modules/planning-units-grid/planning-units-grid.processor.ts
kgajowy: Well, it's a pity we haven't mentioned this in requirements.

Are we sure this makes sense? Allowing user to provide whatever shape they want and marking the as "hex" (which isn't correct, right?) may lead system to produce unexpected results.

aagm: The user will upload their custom planning grid, which could be a regular (square, hexagon, triangles...) or an irregular one( whatever continuos group of geometries that can represent basins, parcels...). 
For those regulars, the main difference with the ones we create is related to the grid upper left corner origin. If we want to be agnostic to the type we can make use of the same categorization in the enum planning_unit_grid_shape and mark them as from_shapefile. 
But marking it as irregular is just wrong. What do you think?

kgajowy: @aagm I think following https://github.com/Vizzuality/marxan-cloud/pull/526#discussion_r705954882 at this very moment would be fine. (in future we can make the validation/checks also instead asking for a type, we will need to verify it anyway so we can also deduce it, right?) Nevertheless, as long as it isn't that much concern for this moment, `from_shapefile` sounds fair enough for current state.

kgajowy: updated in https://github.com/Vizzuality/marxan-cloud/pull/525/commits/05039600456c4312c73e78d2b4e412948b166c9b

aagm (approved):

api/apps/geoprocessing/src/modules/planning-units-grid/planning-units-grid.processor.ts
```diff
@@ -0,0 +1,76 @@
...
+        INSERT INTO "planning_units_geom"("the_geom", "type", "project_id")
+        SELECT ST_SetSRID(
+                 ST_GeomFromGeoJSON(features ->> 'geometry'),
+                 4326)::geometry,
+               $2::shape_type,
+               $3
+        FROM (
+               SELECT json_array_elements($1::json -> 'features') AS features
+             ) AS f
+        ON CONFLICT ON CONSTRAINT planning_units_geom_the_geom_type_key DO UPDATE SET type = 'from_shapefile'::shape_type
```
hotzevzl, line 56: I am not too sure about this - I guess you and Alicia discussed this so it will be ok - at a glance, I'm not sure this caters to a desired use case, but I also see no harm as we're updating existing shapes within the boundaries of a project.

hotzevzl (approved):

api/apps/geoprocessing/src/modules/planning-units-grid/planning-units-grid.processor.ts
kgajowy: @hotzevzl the only reason I update is I need to, otherwise IDs won't be returned back. Seems like there is no harm (could as well update projectId)

hotzevzl:
> > In terms of validation, it is part of the characteristic of a grid that their members are adjacent geometries. in order to validate it, they should not intersect with each other, but every single geometry should be [intouch](https://postgis.net/docs/ST_Touches.html) (connected) with at least one. If we want to do that validation with javascript, i think turf will allow you to intersect and check if the result is a line in order to ensure that they only share borders.
> 
> Can we validate this assumption with postgis?
> 
> would https://postgis.net/docs/ST_IsSimple.html be fair enough (not sure about mentioned tangency) - checked, seems like not, as touching edges make it "invalid"
> 
> https://postgis.net/docs/postgis_usage.html#eval_spatial_rel

not something I have much experience with, but I would be surprised if this wasn't the case, and it would be much more preferable than doing this at the API layer. We can deal with this in a separate validation story, if effectively needed.

PR #526, opened 2021-09-07
docs(custom-pu-grid): initial hld

kgajowy:
https://vizzuality.atlassian.net/browse/MARXAN-252

Next steps: 
- [ ] confirm HLD is correct
- [ ] agree on front <-> back communication (ideally: curls/openApi) - see LLD document


kgajowy:
@mbarrenechea would you want to declare how frontend team wants to process with uploading / fetching status? (don't mind current setup if you want)

docs/custom-planning-unit-grid-geometry.hld.md
```diff
@@ -0,0 +1,17 @@
...
+relying on generated square or hexagonal grids, for a variety of conservation
+reasons, so they should be able to supply their own PU grid via a shapefile
+upload.
+
+# Key assumptions
+
+* once set, shape cannot be changed or removed from project
+* only `FeatureCollection` with `Polygon` features can be uploaded
+* custom planning unit grid geometry belongs to given project
+* custom planning unit grid geometry has a type of `irregular`
```
aagm, line 13: `Irregular` is not for user-uploaded data, is for another kind of PU grid that will not follow a regular shape (i.e. River basins), and we will provide it in a future phase of the project. (Instead, please provide the options to the user)

aagm (changes requested):

docs/custom-planning-unit-grid-geometry.hld.md
kgajowy: @aagm what should then be the type then? if this is a new type, what possible implications it has?

kgajowy:
@mbarrenechea 
Regarding LLD (api contract), it could be:

`POST projects/:id/grid` with similar to other shapefile-uploads arguments (file attachment), returning request ID to track
`GET projects/:id/grid/:requestId` to get the status

the last one is just a proposition how we could implement it for such cases, however we can keep using `GET projects/:id/scenarios/status` to acquire all pending asynchronous jobs.

kgajowy:
@mbarrenechea having no response, implemented it with the second option (add to `.../scenario/status`) for now to keep the old style.

docs/custom-planning-unit-grid-geometry.hld.md
aagm: The user should provide the type and the 3 we have should be kept (we should also expand it with from_shapefile:  [square, hexagon, irregular, from_shapefile]) and match the enum in the api planning_unit_grid_shape ( [square, hexagon, irregular, from_shapefile]), What do you think?

kgajowy: added in https://github.com/Vizzuality/marxan-cloud/pull/526/commits/99dbd8f7e4de43afdabaf873430248257ef2880a

docs/custom-planning-unit-grid-geometry.lld.md
```diff
@@ -0,0 +1,60 @@
...
+curl -X POST "http://localhost:3030/api/v1/projects/:projectId/grid" -H  
+"accept: 
+application/json" -H  "Authorization: Bearer <<TOKEN>>" -H  "Content-Type: 
+multipart/form-data" -F "file=shape.zip;type=application/zip"
+```
+
+ID returned from request can be ignored. In the future, it may be used to 
+fetch status of the job directly.
+
+# Getting status of the processing
```
Dyostiq, line 13: it may get out of sync with the code

docs/custom-planning-unit-grid-geometry.hld.md
```diff
@@ -0,0 +1,17 @@
...
+# Key assumptions
+
+* shape can be re-uploaded as long as there are no scenarios created yet
+* only `FeatureCollection` with `Polygon` features can be uploaded
+* custom planning unit grid geometry belongs to given project
+* custom planning unit grid geometry has a type of `from_shapefile`
+* project with custom planning unit grid geometry cannot use other shape types
+* uploaded shapefile will be processed asynchronously
+* until processed, any action on project/underlying scenarios shouldn't be
+  allowed
```
aagm, line 17:
```suggestion
  allowed
  
  ## Processing requirements after a successful user upload of a Custom planning grid:
  * From the boundaries of it, we need to generate a custom planning area.
  * When a scenario is created we need to attach the custom grid with the scenario (need to expand current flow).
```

docs/custom-planning-unit-grid-geometry.hld.md
```diff
@@ -0,0 +1,17 @@
...
+
+Users may sometimes create their own planning unit geometries, instead of
+relying on generated square or hexagonal grids, for a variety of conservation
+reasons, so they should be able to supply their own PU grid via a shapefile
+upload.
+
+# Key assumptions
+
+* shape can be re-uploaded as long as there are no scenarios created yet
+* only `FeatureCollection` with `Polygon` features can be uploaded
```
aagm, line 11: I am very concerned if they end up with irregular shapes that need multipolygons. Also, we need to ensure this feature collection really represents a grid or a surface.

docs/custom-planning-unit-grid-geometry.hld.md
hotzevzl: as a corollary (ish) of the first of the suggested additions here, we also set the bbox for the project - worth making it explicit here, I think

PR #537, opened 2021-09-27
fix(cost-surface): consume uploaded file

kgajowy:
There were a few issues:
* generated template used different fields than API expected when consuming (cost property name)
* code assumed that all rows exist , thus would only update the available ones. It proved that rows are never inserted. Instead, implemented re-creating all rows based on upload
* file was never processed by `multer` thus was not available within `tmp/storage`

Note: there may be some issue with mapshaper, as it removed some PU while testing...


kgajowy:
Sadly, Extract doesn't like spaces
```
marxan-geoprocessing  | [Nest] 1153   - 09/27/2021, 8:11:52 AM   [ShapefileService] Error: 8bbeda10-b90c-45cf-a194-1e599cb0e0a9_with cost and spaces in file.zip could not be extracted: Error: ENOENT: no such file or directory, open '"/tmp/storage/8bbeda10-b90c-45cf-a194-1e599cb0e0a9_with cost and spaces in file.zip"' +118272ms
```

hotzevzl (approved):

hotzevzl:
> Sadly, Extract doesn't like spaces
> 
> ```
> marxan-geoprocessing  | [Nest] 1153   - 09/27/2021, 8:11:52 AM   [ShapefileService] Error: 8bbeda10-b90c-45cf-a194-1e599cb0e0a9_with cost and spaces in file.zip could not be extracted: Error: ENOENT: no such file or directory, open '"/tmp/storage/8bbeda10-b90c-45cf-a194-1e599cb0e0a9_with cost and spaces in file.zip"' +118272ms
> ```
> 
> wrapping it in `"` does not help. Also just "replacing" spaces within filename within multer options causes another issues, so enough for this PR.

should we reject (with a clear error message) uploads of files that contain spaces, in the meanwhile?

hotzevzl:
> Note: there may be some issue with mapshaper, as it removed some PU while testing...

Are you sure? From a quick look at the expected/received differences in Jest's output, could it be that only the order of features is different?

Received:

Screenshot 2021-09-27 at 15-00-01 geojson io

Expected:

Screenshot 2021-09-27 at 15-00-20 geojson io

kgajowy:
> > Note: there may be some issue with mapshaper, as it removed some PU while testing...
> 
> Are you sure? From a quick look at the expected/received differences in Jest's output, could it be that only the order of features is different?
> 
> Received:
> 
> Screenshot 2021-09-27 at 15-00-01 geojson io
> 
> Expected:
> 
> Screenshot 2021-09-27 at 15-00-20 geojson io

the matcher already ignores the order but indeed, only within 'features' (not their 'geometries') and apparently geometry is transformed. Not sure yet if only its pivoted order.

kgajowy:
> > Note: there may be some issue with mapshaper, as it removed some PU while testing...
> 
> Are you sure? From a quick look at the expected/received differences in Jest's output, could it be that only the order of features is different?
> 
> Received:
> 
> Screenshot 2021-09-27 at 15-00-01 geojson io
> 
> Expected:
> 
> Screenshot 2021-09-27 at 15-00-20 geojson io

image

apparently, coordinates order is also re-ordered.

kgajowy:
```
for (const geo of underlyingGeoJson.features) {
        expect(
          geoJsonFromGeometries.features.some((geoFromJson) =>
            booleanEqual(geoFromJson, geo),
          ),
        ).toBeTruthy();
      }
```
@hotzevzl the check is quite slow but I see no other way around now.

PR #549, opened 2021-09-30
feat(marxan-run): mark best solution

kgajowy:
Disclaimer: could avoid iterating twice by adding relevant code to most-different but I think it may be possible those two may become more complex in future so best would be to keep them in separation. Also, could use a decorator pattern but it does not seem to complex (yet)


hotzevzl (approved):

hotzevzl:
@kgajowy thank you. Consider if it's worth adding a test checking that in case of a tie on lowest score, the "first" such lowest score-solution is marked as best (see https://vizzuality.atlassian.net/browse/MARXAN-313?focusedCommentId=12952 for details)

kgajowy:
@hotzevzl  https://github.com/Vizzuality/marxan-cloud/pull/549/commits/cc1516295f8793047815cd14c68b490fe3e24b32

as we added one more run, clustering changed (indexes below), so did the most-different ones.
``` [ [ 3 ], [ 4 ], [ 5 ], [ 2, 6 ], [ 1, 0, 7 ] ]```

PR #553, opened 2021-10-04
fix(features): split features properly use subsets prop values

kgajowy:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


api/apps/api/src/modules/scenarios-features/split/split-query.service.ts
```diff
@@ -55,62 +55,62 @@ export class SplitQuery {
...
+        SELECT distinct fpkv.feature_data_id,
+                        fpkv.key,
+                        fpkv.value,
+                        subsets.fpf,
+                        subsets.prop,
+                        subsets.target
+        from feature_properties_kv fpkv
+               left join subsets
+                         on trim('"' FROM fpkv.value::TEXT) = trim('"' FROM subsets
+                           .sub_value::text)
```
kgajowy, line 83: ^ The only changed part

hotzevzl: this is because casting `fpkv.value` to `text` gives us a text representation of the JSON document.

I cannot test quickly right now but if you want to avoid casting and trimming (not sure about the perf penalty, likely marginal if any), IIRC `on fpkv.value->>0` *should* do.

kgajowy: @hotzevzl yep, that works for this case but not sure if this will always work (as we have jsonb, we may have something else that primitive types)

image

hotzevzl: yep, *in theory* it should work for the features we import via ETL, but without explicit guards to ensure we always have a single string or number (also when importing user-uploaded features), I agree that we can't rely on this assumption.

PR #560, opened 2021-10-08
feat(clone): domain

kgajowy:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


api/apps/api/src/modules/clone/export/domain/export/export.ts
```diff
@@ -0,0 +1,78 @@
...
+    piece?.finish(pieceLocation);
+
+    if (this.#allPiecesReady()) {
+      this.apply(new ClonePartsFinished(this.id));
+    }
+  }
+
+  complete(archiveLocation: ArchiveLocation) {
+    this.archiveLocation = archiveLocation;
+    this.apply(new ArchiveReady(this.id, this.archiveLocation));
```
Dyostiq, line 66: even if parts are not finished?

kgajowy: good point!

kgajowy: fixed in https://github.com/Vizzuality/marxan-cloud/pull/560/commits/99777f08e314427d806debeb7a85aa4606ee9549

api/apps/api/src/modules/clone/export/domain/export/export.ts
```diff
@@ -0,0 +1,107 @@
...
+import { ExportSnapshot } from './export.snapshot';
+
+export const pieceNotFound = Symbol('export piece not found');
+export const notReady = Symbol('some pieces of export are not yet ready');
+
+export class Export extends AggregateRoot {
+  private constructor(
+    public readonly id: ExportId,
+    public readonly resourceId: ResourceId,
+    public readonly resourceKind: ResourceKind,
```
Sikora00, line 26: Are you sure that kind is a good discriminator? Where it will affect the app logic?

api/apps/api/src/modules/clone/export/domain/export/clone-part/clone-part.ts
```diff
@@ -0,0 +1,51 @@
+import { v4 } from 'uuid';
+import { ClonePiece } from '../../../../shared-kernel/clone-piece';
+
+import { ResourceId } from '../resource.id';
+import { ClonePartSnapshot } from '../clone-part.snapshot';
+
+import { PieceId } from './piece.id';
+import { PieceLocation } from './piece-location';
+
+export class ClonePart {
```
Sikora00, line 10: What is that? Could you provide an example?

Sikora00 (changes requested):

api/apps/api/src/modules/clone/export/application/request-export.ts
```diff
@@ -0,0 +1,30 @@
...
+  constructor(
+    private readonly resourcePieces: ResourcePieces,
+    private readonly exportRepository: ExportRepository,
+    private readonly eventPublisher: EventPublisher,
+  ) {}
+
+  async export(id: ResourceId, kind: ResourceKind): Promise<ExportId> {
+    const parts: ClonePart[] = await this.resourcePieces.resolveFor(id, kind);
+    const exportInstance = this.eventPublisher.mergeObjectContext(
+      Export.project(id, parts),
```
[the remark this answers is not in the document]
kgajowy: Don't think so, why? Resolved pieces (components) could also be a scenario ones.

api/apps/api/src/modules/clone/export/domain/export/export.ts
kgajowy: What you would propose?

kgajowy:
> @kgajowy overall the approach looks good to me. One thing I recommend to keep in mind is that the core process here is really the cloning of projects or scenarios, with artifacts being a (fundamental) byproduct in the case of the project download/upload workflow (rather than cloning without intermediate steps).
> 
> I imagine this may mean still passing through some kind of intermediate representation of a scenario and project, of which the exportable and importable artifact may be a complex serialization - also taking into account that any such intermediate representation will often be something we should not assume that will reliably fit into memory, so persisting components' IR in some for may be unavoidable. But I'd consider the exportable/importable artifacts as a further step - all of this to be validated in terms of possible approaches, of course.

What would you propose @hotzevzl ? I am not sure if we can have any reliable data representation useful just for cloning (even the sqls) which would make it easy to then represent them in a different manner. How I imagined in there is that cloning would work exactly like export->import but without sharing the archive (thats why we keep geojsons/shapefiles rather than csv's with geometry)

api/apps/api/src/modules/clone/export/application/request-export.ts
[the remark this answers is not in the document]
Dyostiq: but `Export.project(id` suggests it is only for the project, and `id` is a project id, `kind` is not used here
and for kind === `scenario`, `Export.project` is used as well what is surprising

PR #565, opened 2021-10-11
refactor(geo-features): remove unused code

kgajowy:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


hotzevzl (approved):
Should we also "downgrade" the v2 endpoint to the path of the ones we're removing here?

This will need updating in `app/hooks/features/index.ts` too - would you be ok with this tiny cleanup on both API and app side @mbarrenechea? Just to avoid having this v2 path component sticking out for no obvious reason since we don't need to provide backwards compatibility with the "v1" endpoints.

PR #570, opened 2021-10-12
feat(planning-unit-grid): derive planning area from custom grid

kgajowy:
(the original, as well as expected planning area geojsons are added to this PR and can be viewed on github rich diff)

Please note that using non-grid geojson (like previous one, with "random" polygons) won't create a proper planning area - with the same query, geometry won't change at all (thus will remain as collection of polygons)

Uploading grid:
image

Produced planning area:
image

BBox:
image

Run result:
image

# Manual testing steps
* create new project database row (keep in mind to set NULLs for Gadm and area)
* (also add relevant entry to users_projects to make it visible on frontend)
* please note that, if you are using `restore_dump`, it can crash if you upload the same geometry more than once (missing restriction on project_id in `planning_units_geometry`)
* send post:
```
curl --request POST \
  --url http://localhost:3030/api/v1/projects/PROJECT_ID/grid \
  --header 'Authorization: Bearer TOKEN' \
  --header 'content-type: multipart/form-data; boundary=---011000010111000001101001' \
  --form file=@/SHAPE_FILE_PATH_FROM_TESTS
```

### DIFF
--- a/api/apps/api/src/modules/projects/events/custom-planning-unit-grid-set.event.ts
+++ b/api/apps/api/src/modules/projects/events/custom-planning-unit-grid-set.event.ts
@@ -1,9 +1,10 @@
 import { IEvent } from '@nestjs/cqrs';
+import { BBox } from 'geojson';
 
 export class CustomPlanningUnitGridSet implements IEvent {
   constructor(
     public readonly projectId: string,
     public readonly planningAreaId: string,
-    public readonly bbox: number[],
+    public readonly bbox: BBox,
   ) {}
 }
--- a/api/apps/geoprocessing/src/modules/planning-units-grid/planning-units-grid.processor.ts
+++ b/api/apps/geoprocessing/src/modules/planning-units-grid/planning-units-grid.processor.ts
@@ -4,6 +4,8 @@ import { plainToClass } from 'class-transformer';
 import { validateSync } from 'class-validator';
 import { InjectEntityManager } from '@nestjs/typeorm';
 import { EntityManager } from 'typeorm';
+import { isLeft } from 'fp-ts/Either';
+import { BBox } from 'geojson';
 
 import { JobInput, JobOutput, queueName } from '@marxan/planning-units-grid';
 import { ShapefileService } from '@marxan/shapefile-converter';
@@ -14,7 +16,6 @@ import {
 
 import { ShapeType } from '@marxan-jobs/planning-unit-geometry';
 import { GridGeoJsonValidator } from './grid-geojson-validator';
-import { isLeft } from 'fp-ts/Either';
 
 @Injectable()
 export class PlanningUnitsGridProcessor
@@ -81,7 +82,7 @@ export class PlanningUnitsGridProcessor
 
         const planningArea: {
           id: string;
-          bbox: number[];
+          bbox: BBox;
         }[] = await manager.query(
           `
             INSERT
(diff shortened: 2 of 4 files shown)


api/apps/geoprocessing/src/modules/planning-units-grid/planning-units-grid.processor.ts
```diff
@@ -40,27 +40,77 @@ export class PlanningUnitsGridProcessor
...
-               SELECT json_array_elements($1::json -> 'features') AS features
-             ) AS f
-        ON CONFLICT ON CONSTRAINT planning_units_geom_the_geom_type_key DO UPDATE SET type = 'from_shapefile'::shape_type
-        RETURNING "id"
-      `,
-      [result.right, ShapeType.FromShapefile, projectId],
-    );
+
+    const { puGeometriesIds, planningAreaId, bbox } = await this.entityManager
+      .transaction(async (manager) => {
```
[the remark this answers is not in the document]
hotzevzl: good point. until there are no scenarios, it's ok for API clients to keep sending a new PU grid, and we'll happily use the latest (with usual caveats about concurrency of rapid-fire identical uploads).

in practice, if a rogue API user hits the PU grid upload endpoint after a scenario has been created, chaos will ensue.

so tl;dr it's ok to delete geoms and pa here, but I think we should add a guard in `ProjectsService.setGrid()` to reject grid uploads if the project has one or more scenarios. If I haven't missed an existing guard and this sounds like a sensible plan to you, it can be done in a separate PR, in any case.

hotzevzl:
> Please note that using non-grid geojson (like previous one, with "random" polygons) won't create a proper planning area - with the same query, geometry won't change at all (thus will remain as collection of polygons)

which is the previous non-grid geojson? (I see only the new NAM grid and "unionized" PA in the repo folder in this PR)

in any case, I guess you mean non-always-touching polygons... in which case, what happens? I would expect that polygons that touch will be joined in a new merged polygon while isolated ones will stay as they are - if so, this sounds ok to me for now, until we confirm and implement (if at all) validation of contiguity of polygons for PU grid uploads.

api/apps/api/src/modules/projects/events/custom-planning-unit-grid-set.event.ts
```diff
@@ -1,5 +1,9 @@
 import { IEvent } from '@nestjs/cqrs';
 
 export class CustomPlanningUnitGridSet implements IEvent {
-  constructor(public readonly projectId: string) {}
+  constructor(
+    public readonly projectId: string,
+    public readonly planningAreaId: string,
+    public readonly bbox: number[],
```
Dyostiq, line 7:
```suggestion
    public readonly bbox: Bbox,
```

api/apps/geoprocessing/src/modules/planning-units-grid/planning-units-grid.processor.ts
```diff
@@ -40,27 +40,77 @@ export class PlanningUnitsGridProcessor
...
+          `
+            INSERT
+            INTO "planning_areas"("project_id", "the_geom")
+            VALUES ($1,
+                    (SELECT ST_MULTI(ST_UNION(the_geom))
+                     from "planning_units_geom"
+                     where project_id = $2))
+            RETURNING "id", "bbox"
+          `,
+          [projectId, projectId],
```
Dyostiq, line 95:
```suggestion
            VALUES ($1,
                    (SELECT ST_MULTI(ST_UNION(the_geom))
                     from "planning_units_geom"
                     where project_id = $1))
            RETURNING "id", "bbox"
          `,
          [projectId],
```

Dyostiq (approved):

PR #573, opened 2021-10-13
HLD - BLM calibration

hotzevzl:
https://vizzuality.atlassian.net/browse/MARXAN-890

Backporting this from the relevant draft discussed this past week, with some final curation, after agreement on the data-related requirements.


docs/features/boundary-length-modifier/blm-calibration.hld.md
```diff
@@ -0,0 +1,83 @@
...
+across each of the runs.
+
+## Requirements
+
+- Users should be able to set their BLM values.
+
+The set should be limited to a small cardinality: 6 values for the initial
+implementation, but leaving space for adjustments in the future (i.e. it is ok
+to hardcode the size of the set, but changing it in the future should not break
+other assumptions).
```
kgajowy, line 45: So the initial run should already be N-setups (each for M runs) in parallel for given scenario and those values can be manually modified by the user? (i.e. N inputs presented with pre-calculated values with formula below)

docs/features/boundary-length-modifier/blm-calibration.hld.md
```diff
@@ -0,0 +1,83 @@
...
+
+- The 6 values given by default will be `BLM = Z*sqrt(PU area)` (with `Z`  being
```
kgajowy, line 47: to paraphrase - according to below `to be decided`, it is a single PU area, not total?

docs/features/boundary-length-modifier/blm-calibration.hld.md
```diff
@@ -0,0 +1,83 @@
...
+
+- Curve fitting across the resulting scores.
+
+- Optimal value selected as the “elbow” of the curve (see Marxan manual, page
+  5-36, Box 12).
+
+## To be decided
+
+- formula to use in case of varying PU area
+- a possibly updated user flow (this won't affect backend implementation)
```
kgajowy, line 83: it may, right? now we support providing only one value to run with (if not provided, will use original setting from metadata)

docs/features/boundary-length-modifier/blm-calibration.hld.md
```diff
@@ -0,0 +1,83 @@
...
+
+- A line chart will be shown to users, but it will reflect the actual values
+  (being or not a curve).
+
+For reference, the initial functional design involved:
+
+- Using a *fixed* set of BLM values (6 of them), to be used for every
+  project/scenario.
+
+- Curve fitting across the resulting scores.
```
kgajowy, line 75:
```
alongside the 
--> cost <-- 
of the best solution foreach of them.
```
thus, cost or score?

docs/features/boundary-length-modifier/blm-calibration.hld.md
kgajowy: Also:
*  which scores are presented to user once the run above finishes?
* what about metadata setting of BLM value?

docs/features/boundary-length-modifier/blm-calibration.hld.md
```diff
@@ -0,0 +1,83 @@
...
+
+- Optimal value selected as the “elbow” of the curve (see Marxan manual, page
```
kgajowy, line 77: Let's add the pdf to the repository.

docs/features/boundary-length-modifier/blm-calibration.hld.md
kgajowy: also, this may be important and an important point in HLD, as we also should agree how it will work cross-teams.

kgajowy: marxan gets it from generated input.dat - but it can be overridden with some value when requesting a run (endpoint) - then generated "assets" will include the different BLM value within URL (😅)

Anyway the question follows in the thread above https://github.com/Vizzuality/marxan-cloud/pull/573#discussion_r728085336 so closing this one

docs/features/boundary-length-modifier/blm-calibration.hld.md
hotzevzl: it may eventually become a further option in the analysis tab, that the users can use to explore different settings before running the scenario (https://projects.invisionapp.com/share/G211WEUZFST9#/screens/459986030_Marxan_09a).

Once happy with BLM calibration results, users will typically run Marxan, either after setting BLM from one of the suggestions from the calibration, or if they want to ignore these, using either the default or a value they choose (they may have this from analysis they did outside of the platform).

docs/features/boundary-length-modifier/blm-calibration.hld.md
hotzevzl: 93b730e0d236675800e92e6f36ab76a65bb690af

PR #576, opened 2021-10-14
allow to delete projects that use custom features

hotzevzl:
https://vizzuality.atlassian.net/browse/MARXAN-897


kgajowy (commented):
@hotzevzl pretty sure if TypeOrm will not pick it up correctly without entity definitions. We shall try to match the migrations as:
1) easier to track changes over entity
2) we are not letting Orm to do its stuff on delete (for example: hooks, subscriptions)

kgajowy (commented):
https://github.com/Vizzuality/marxan-cloud/pull/576#pullrequestreview-779722415

PR #583, opened 2021-10-18
avoid dropped proxy connections

hotzevzl:
https://vizzuality.atlassian.net/browse/MARXAN-783

This was indeed happening across all the MVT endpoints, no specific patterns, **and** in fact the requests were being handled correctly by the relevant endpoints in the geoprocessing service, i.e. the service would get the data from postgis, then prepare the response, but by then the proxy (in the api service) had given up, claiming that the connection had been closed upstream.

These dropped connections were actually happening straight away. Symptom would be a red flag for non-`await`ed promises, and yet, besides things looking ok, only by swapping the `http-proxy` code with simple proxying via Node's own http module I could see these "dropped connections" going away.

For reference, I also moved the creation of the proxy server (when using `http-proxy`) to the proxy function itself (`ProxyService.proxyTileRequest()`) rather than having it instantiated in the constructor, without any difference - connections would be dropped, seemingly randomly.


api/apps/api/src/modules/proxy/proxy.service.ts
```diff
@@ -1,67 +1,27 @@
...
-    );
+  async proxyTileRequest(req: Request, res: Response) {
+    const url =
+      this.geoprocessingServiceUrl +
+      req.path +
+      new URL(req.originalUrl, this.geoprocessingServiceUrl).search;
+    get(url, { headers: req.headers }, (response) => {
+      // As we're proxying, set status code/message and headers from upstream
+      res.statusCode = response.statusCode!;
+      res.statusMessage = response.statusMessage!;
```
hotzevzl, line 22: I am not sure how to handle this otherwise - from the Node documentation my understanding is that by this stage (`response` event) we should always have `statusCode` and `statusMessage`, but the upstream typing does not agree with this.

kgajowy: Could it be that those responses are different than Express ones?

api/apps/geoprocessing/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -72,11 +72,8 @@ export class ScenariosController {
...
     @Res() response: Response,
-  ): Promise<Object> {
-    const tile: Buffer = await this.service.findTile(
-      tileRequest,
-      scenariosPuFilters,
-    );
-    return response.send(tile);
+  ): Promise<void> {
+    const tile = await this.service.findTile(tileRequest, scenariosPuFilters);
+    response.send(tile);
```
Dyostiq, line 77: BTW:
while `Res` is used, `Header`s above are ineffective (unless `@Res({passthrough: true})` and `StreamableFile` are used)

api/apps/api/src/modules/proxy/proxy.service.ts
```diff
@@ -1,67 +1,27 @@
-import { HttpStatus, Injectable, Logger } from '@nestjs/common';
+import { Injectable, Logger } from '@nestjs/common';
 
 import { AppConfig } from '@marxan-api/utils/config.utils';
 import { Request, Response } from 'express';
-/**
- * @debt i think this is not properly imported, but it was the only way for it to work
- */
-import * as Server from 'http-proxy';
+import { get } from 'http';
```
Dyostiq, line 5: alternatively we could use HttpService with `responseType: 'stream'` like @kgajowy did in `AssetFetcher`

PR #590, opened 2021-10-21
libbot: add support for number of runs

hotzevzl:
This could be achieved by setting the related property via POST/PATCH /api/v1/scenarios/:id; this PR adds a little convenience method to do so via libbot's "DSL".

The PR also updates the specification submission endpoint path removing the recently deprecated `/v2` suffix.

### DIFF
--- a/data/bots/lib/libbot/geo-feature-specifications.ts
+++ b/data/bots/lib/libbot/geo-feature-specifications.ts
@@ -24,7 +24,7 @@ export class GeoFeatureSpecifications {
     };
 
     return await this.baseHttpClient.post(
-      `/scenarios/${scenarioId}/features/specification/v2`,
+      `/scenarios/${scenarioId}/features/specification`,
       specification,
     )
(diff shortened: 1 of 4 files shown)


data/bots/lib/libbot/geo-feature-specifications.ts
```diff
@@ -24,7 +24,7 @@ export class GeoFeatureSpecifications {
     };
 
     return await this.baseHttpClient.post(
-      `/scenarios/${scenarioId}/features/specification/v2`,
+      `/scenarios/${scenarioId}/features/specification`,
```
kgajowy, line 27: ah sorry, forgot about this place when was changing!

kgajowy (approved):

data/bots/lib/libbot/geo-feature-specifications.ts
hotzevzl: well, *I* should have known better, also because it was me who pushed to drop the `v2` 😅

PR #600, opened 2021-10-25
feat(protected-area): get protected areas for scenario

kgajowy:
* untangles a bit protected-areas from project/scenario
* still have some remains (required to update/create) but those should disappear once Update is implemented
* breaks possibility to update now - uses old endpoint which does not support UUIDs

**DO NOT MERGE** until full feature is implemented. Please use this branch as a base for further PRs related to feature.


api/apps/api/test/scenario-protected-areas/fixtures.ts
```diff
@@ -0,0 +1,156 @@
...
+    GivenScenarioInsideBRA21WasCreated: async () => {
+      const { id } = await GivenScenarioExists(app, projectId, token);
+      return id;
+    },
+    WhenGettingProtectedAreas: async (scenarioId: string) =>
+      request(app.getHttpServer())
+        .get(`/api/v1/scenarios/${scenarioId}/protected-areas`)
+        .set('Authorization', `Bearer ${token}`)
+        .then((response) => response.body),
+    ThenInContainsRelevantWdpa: async (response: unknown) => {
```
hotzevzl, line 56:
```suggestion
    ThenItContainsRelevantWdpa: async (response: unknown) => {
```

hotzevzl (approved):
thanks, not least for the huge effort to clean up/refactor/repay debt :)

looks good in general, I still wonder if we could do it slightly more "cheaply" (lower SLOC 😓) but sounds good for the needs.

can we clarify in code (maybe just some comments) what's the role of the `selected` prop and how we're expecting to handle it?

also, very minor, I suggest to stick to "IUCN categories" for this concept when naming things, unless there's a good reason to use just "categories" or "WDPA categories".

otherwise, good to go IMHO, though the app part I am not sure about (is it supposed to have visible changes?)

kgajowy:
> , though the app part I am not sure about (is it supposed to have visible changes?)

yep, this is a base branch and have some frontend PR already merged, sorry for that. Not yet mergable until fully implemented tho.

PR #614, opened 2021-11-02
added new data seeds

aagm:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


kgajowy (changes requested):
Seeds fail to run :(

aagm:
Fixed @kgajowy; 
**but bear this in mind:** We need to modify how the test for scenarios are done; They rely on having data linked with scenarios that are removed from being manually set on the seeds. We need to update the full flow of setting up a project --> scenario --> wdpa selection --> feature recipe + feature calc step in `api/apps/api/test/scenario-features/` (a separate task from this)

kgajowy:
@aagm 
image

given the exit code 2 as a e2e specs results and still green light (cc @hotzevzl ?) , not sure if we can trust the "all good" from CI. Can we fix it?

hotzevzl:
> @aagm image
> 
> given the exit code 2 as a e2e specs results and still green light (cc @hotzevzl ?) , not sure if we can trust the "all good" from CI. Can we fix it?

tl;dr ok to merge this despite the flaky false positives, from my point of view.

fix and postmortem on their way, but tl;dr of that: from what I have seen while trying to track this issue down, what causes CI to report a failure is the non-zero exit code for one of the make recipes (`test-start-services`), caused by a failed migration *run*. The effects caused by the migrations will have happened, however (as the root cause is that a CI setup error causes a duplicated attempt to run migrations), and tests can still pass (bar any failing ones for unrelated reasons).

So when CI says we're green, tests were succeeding and the issue we're looking at did not trigger - in which case things look fine to me. We do have false positives, however, when tests pass but the issue above triggers, but at least in local e2e test runs the fix I prepared yesterday fixes the issue - but I will need to verify in the actual CI runners.

PR #616, opened 2021-11-02
fix(front): [MARXAN-962][MARXAN-949]

anamontiaga:
### Overview

- [x] Scroll to bottom when select a country planning area
- [x] User resets password and is redirected to an intermediate screen

### Feature relevant tickets

[MARXAN-962](https://vizzuality.atlassian.net/browse/MARXAN-962)
[MARXAN-949](https://vizzuality.atlassian.net/browse/MARXAN-949)

---

### DIFF
--- a/app/layout/projects/new/form/component.tsx
+++ b/app/layout/projects/new/form/component.tsx
@@ -119,8 +119,8 @@ const ProjectForm: React.FC<ProjectFormProps> = () => {
   };
 
   const scrollDown = useCallback((ref) => {
-    console.log('Hola', ref);
-    ref?.current?.scrollIntoView({ block: 'center', behavior: 'smooth' });
+      console.log('scroll down', ref);
+      ref?.current?.scrollIntoView({ block: 'end', behavior: 'smooth' });
   }, []);
 
   const resetPlanningArea = (form) => {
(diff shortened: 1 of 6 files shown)


app/layout/projects/new/form/country-region-selector/types.ts
```diff
@@ -2,4 +2,5 @@ export default interface CountryRegionSelectorProps {
   country?: string;
   region?: string;
   subRegion?: string;
+  onClick?: any;
```
anamontiaga, line 5: Not found the correct type as it doesn't accept ()=> void

app/layout/projects/new/form/component.tsx
```diff
@@ -303,14 +311,18 @@ const ProjectForm: React.FC<ProjectFormProps> = () => {
 
                     {PAOptionSelected === 'regular' && (
                       <>
+
                         <CountryRegionSelector
                           country={values.countryId}
                           region={values.adminAreaLevel1Id}
                           subRegion={values.adminAreaLevel2Id}
+                          onClick={scrollDown(planningAreaScrollRef)}
```
anamontiaga, line 319: It only works without void. It may be the ref, that doesn't render first time.

app/layout/projects/new/form/component.tsx
```diff
@@ -114,6 +118,10 @@ const ProjectForm: React.FC<ProjectFormProps> = () => {
     });
   };
 
+  const scrollDown = (ref) => {
+    ref?.current?.scrollIntoView({ block: 'center', behavior: 'smooth' });
+  };
+
```
kgajowy, line 124: also, `useCallback` would be good.

kgajowy (changes requested):

app/layout/projects/new/form/component.tsx
kgajowy: @anamontiaga it will trigger each time during rendering. 

`() => scrollDown(planningAreaScrollRef)`

and

` onClick?: () => void;` will work fine and as expected.

kgajowy:
Reject due to (possibly) unintended behavior: https://github.com/Vizzuality/marxan-cloud/pull/616#discussion_r741668080

PR #621, opened 2021-11-05
Brief and HLD for role based access control

ikas:
### Overview

This PR is a proposal of implementation for role based access control in Marxan. This PR should be used as a place for the discussion of the implementation.

### Feature relevant tickets

[MARXAN-961](https://vizzuality.atlassian.net/browse/MARXAN-961)


docs/features/role-based-access-control/brief.md
```diff
@@ -0,0 +1,40 @@
+# Role base access control - Brief
+
+This document aims at providing an overview of Role-based Access Control (RBAC from here on) for the Marxan platform. 
+
+In broad strokes, RBAC assumes the existence of a set of roles, which are associated with different actions that they 
+can perform in the platform. Users can then be associated with roles for specific resources, which will grant them the 
+role privileges for that specific resource.
+
+The current implementation is based on the requirements described in the [User requirements document](https://docs.google.com/document/d/16x4wQYTKov0JkO1ue6iQzJ3-RXpdgQ38ygGJWW86Zo0/edit#heading=h.58xbt2es73r2).
```
kgajowy, line 9: Once document has a final form, we should move it there (either pdf or key points required for understanding the feature.

docs/features/role-based-access-control/brief.md
```diff
@@ -0,0 +1,40 @@
...
+## Roles
+
+In the context of the Marxan platform, we will consider the following roles:
+
+* **Admin**: a generic platform administrator, who can perform **any action for any resource** in the platform.
+* **Owner**: represents a user that can perform any management actions for a specific resource of the platform (e.g. 
+a project). This role is typically attributed to the user who created the resource (i.e. they own the resource).
+* **Contributor**: represents a user with read-write access to a specific resource. Users with this role cannot manage 
+the roles of users associated with a resource or delete the resource.
+* **Viewer**: represents a user with read-only access to a given resource.
```
kgajowy, line 20:
```suggestion
* **Viewer**: represents a user with read-only access to a given resource or particular part of the resource (think of `solutions` for scenario, or map tiles)
```

docs/features/role-based-access-control/brief.md
kgajowy: Also see `Project solution viewer`

docs/features/role-based-access-control/brief.md
```diff
@@ -0,0 +1,40 @@
...
+1. Organizations
+2. Projects
+3. Scenarios
+
+The entities stated above will work in a hierarchical way: roles granted to users in organizations also apply to all 
+projects owned by said organization, as well as all scenarios in all projects owned by the organization.
+
+However, this hierarchical behavior can be overwritten in a case-by-case basis if needed - e.g. a given user can be 
+"contributor" for an organization, but have "owner" privileges for a specific project inside (or outside) that 
+organization.
```
kgajowy, line 38: I believe it may be out of the scope for the demo. It indeed usually works like that and would be nice but may increase complexity at this very moment and state of the project.

docs/features/role-based-access-control/high-level-design.md
```diff
@@ -0,0 +1,100 @@
+# Role base access control - High-level design
+
+This document aims at providing an overview of the architecture of Role-based Access Control (RBAC from here on) for
+the Marxan platform.
+
+## Roles
+
+Platform administrators should be identified by a boolean attribute in the users table (`is_admin`). This attribute can
+only be edited by other platform administrators, and it must not be editable by the user when editing their profile.
```
kgajowy, line 9: We shouldn't pack `user` table with something not related/needed for every resource. It quickly becomes overloaded with columns and hard to maintain.

More flexible approach would be "just" create a new dedicated table with only `userId` or even keep the as a common `role` type we already have.

Also not sure if this is necessary at this moment - we don't have particular requirements what platform admin can do (at least not said explicit) that would apply for the demo.

docs/features/role-based-access-control/high-level-design.md
```diff
@@ -0,0 +1,100 @@
...
+
+The remaining roles supported by the platform (`Owner`, `Contributor` and `Viewer`) will be stored in the `roles` 
+database in the `marxan-api` database.
```
kgajowy, line 12: Requirements state more than 3 roles:

* Platform admin
* Project owner
* Project contributor
* Project reviewer (read-only)
* Project solution-viewer  [nice-to-have]

docs/features/role-based-access-control/high-level-design.md
```diff
@@ -0,0 +1,100 @@
...
+
+Roles should have numerical values associated to them, to identify the hierarchical order between them. For instance:
+
+* `Owner` -> 1000
+* `Contributor` -> 500
+* `Viewer` -> 250
+
+Roles with fewer privileges should have lower numerical values than roles with more privileges. Using numerical values
+for roles allows for easily comparing if a given user has certain privileges (e.g. checking if a user has the role 
+`Contributor` can be easily done: `role >= Contributor (500)`).
```
kgajowy, line 22: Although I like the idea (deriving from rwx?), it can be misleading as we may have not overlapping capabilities - for example, how would it look when:|
* role A allows to view only solutions
* role B allows to view only scenario setup

Can be compose such values in an obvious and clear way, that "computation" on numerical values would give us relevant access to what user should be allowed, denying the other?

docs/features/role-based-access-control/high-level-design.md
```diff
@@ -0,0 +1,100 @@
...
+
+There is also an implicit relationship between platform actions and the supported roles. By default, it is assumed that:
+  * a user with role `Viewer` can read the information for the given resource (which usually includes calls to 
+`GET resource` and `GET resource/:id` endpoints).
+  * user with role `Contributor` can perform all the actions a `Viewer` can, as well as edition actions for the given
+resource (which usually includes calls to `PATCH/PUT resource/:id` endpoints).
+  * user with `Owner` role can perform all the actions a `Contributor` can, as well as deleting actions for the given
+resource (which usually includes calls to `DELETE resource/:id` endpoints).
+
+**However, it is up to each controller method to correctly check permissions for the action being performed.** This also applies
```
kgajowy, line 32: Why are we restricting ourselves to controller level? I believe it may lead to multiple checks along the way.

Imagine, some role queries the Map Tiles, but depending on the role, they may be denied, or given different set of generated tiles depending on their scope (for example, hide some layers).

Another point to consider that it may end up in duplicating checks (and/or anyway moving them to some service/query) - imagine that some feature wants to trigger another action which may have different restrictions - we would either skip the check from controller, or have to call REST call, or duplicate it.

Also, this approach makes the testing difficult and binds the verification to very given HTTP transport layer.

docs/features/role-based-access-control/high-level-design.md
```diff
@@ -0,0 +1,100 @@
...
+to custom actions that might be subject to RBAC: for instance, if publishing a project is a custom action that should be
+subject to RBAC, then it is up to the controller action to check that the user performing the request has the required role 
+(e.g. `Owner`) to perform that action.
+
+The mapping between the different platform actions and the role required to perform such action is kept in the following
+spreadsheet: https://docs.google.com/spreadsheets/d/1vz-YHdoAAG2nv_B1QUR7LQ-eU_DnKBexrqULrzMMOtw/edit?usp=sharing
```
kgajowy, line 38: As we want to keep HLD the one and only single source of truth, let's avoid linking external documents:
* they may get obsolete
* they may be no longer accessible
* they may be not open for OS contributors
* documents are easy to change without developers knowledge, thus making implementation out of date
* its hard to track changes in the external document 
* keeping it on repository makes it easier to control the flow (everyone can see PR with changes to requirements, approve it, discuss etc.)

docs/features/role-based-access-control/high-level-design.md
```diff
@@ -0,0 +1,100 @@
...
+* `users_scenarios` will store the roles of users inside scenarios.
+
+These tables should contain columns for `user_id`, `role` and the resource id (`organization_id`, `project_id` and 
+`scenario_id`, respectively).
+
+## Implementation details
+
+In terms of implementation, the existing module `projects-acl` should be renamed to a more generic name such as 
+`access-control` or `rbac` (though I'm not a fan of acronyms as names of modules). This module will be used across 
+the platform as the "single-source-of-truth" to check whether users are allowed to access given resources.
```
kgajowy, line 55: The reason it is strictly related to `projects` is to not keep more than needed. I would recommend to keep specific modules, even if they become sub-modules of some `generic` module exposed to other top-lvl modules within the application - otherwise it will become hard to maintain over time, gathering more and more knowledge over time.

docs/features/role-based-access-control/high-level-design.md
```diff
@@ -0,0 +1,100 @@
...
+Checks if the given user can perform the actions of a `Owner` for the provided resource id. This method should return 
+`true` if one of the following conditions apply:
+
+* the user is a platform admin;
+* the user has a role greater or equal to `Owner` for the resource being checked in the corresponding table (e.g. when 
+ checking if a user is owner of a project, `this.projectAclService.isOwner(projectId, userId)`, there must be a record 
+ in the `users_projects` table with role greater or equal to `Owner`).
+* the user has a role greater or equal to `Owner` for a resource hierarchically superior to the one being checked (e.g. 
+ again using projects as an example, if there is no record in the `users_projects` table, but there is a record for the 
+ given user with role greater or equal to `Owner` for the organization that owns the project).
```
kgajowy, line 72: This would require our module to be aware of relations and query for related resources (and thus ask other modules).

I would recommend keeping it as much separated as possible and self-sufficient, not importing anything.

docs/features/role-based-access-control/high-level-design.md
```diff
@@ -0,0 +1,100 @@
...
+the platform as the "single-source-of-truth" to check whether users are allowed to access given resources.
+
+This module, for now, will support checking permissions for the 3 resources subject to RBAC: organizations, projects and
+scenarios. For this, this module should provide 3 services, each of them implementing the following interface:
+
+
+`isOwner(resourceId: string, userId: string): boolean`
+
+Checks if the given user can perform the actions of a `Owner` for the provided resource id. This method should return 
+`true` if one of the following conditions apply:
```
kgajowy, line 64: I think it would be a good idea to add abstract classes/interfaces within this PR to have more "verbose" visuals instead describing them. What do you think?

docs/features/role-based-access-control/high-level-design.md
```diff
@@ -0,0 +1,100 @@
...
+
+Checks if the given user can perform the actions of a `Viewer` for the provided resource id. This method should return 
+`true` if one of the following conditions apply:
+
+* the user is a platform admin;
+* the user has a role higher than `Viewer` for the resource being checked in the corresponding table (e.g. when
+  checking if a user is contributor of a project, `this.projectAclService.isViewer(projectId, userId)`, there must
+  be a record in the `users_projects` table for the role greater or equal to `Viewer`).
+* the user has the role higher than `Viewer` for a resource hierarchically superior to the one being checked (e.g.
+  again using projects as an example, if there is no record in the `users_projects` table, but the user has a role higher
```
kgajowy, line 99: Sometimes the particular description of implementation is a good-to-have but being too strict (pointing particular fields/tables) make it not being high-level document and denies the flexibility. I would leave the implementation details to low-level-design or not put them at all.

High level module API should be fair enough for HLD (i.e. how other modules will use this root lvl module)

docs/features/role-based-access-control/brief.md
ikas: These documents describe the long-term implementation of RBAC, but the demo will be just a subset of the features described in this document (indeed, not including the item you commented on above). Do you think adding a "development-plan" doc might help clarify this?

docs/features/role-based-access-control/high-level-design.md
ikas: Well, my idea was that a platform admin might actually impact every resource in the platform, hence why I added it to the `user` table. However, I don't see any problem with keeping a dedicated database table to store platform admins.

Regarding "not sure if this is necessary at this moment" see https://github.com/Vizzuality/marxan-cloud/pull/621#discussion_r744541550

docs/features/role-based-access-control/high-level-design.md
ikas: My idea is to have generic/reusable roles, that not only cover the needs of the project but also that can be accommodate future needs - the roles described in the current requirements would be covered as such:

- "Platform admin" would be achieved with the new DB table for platform admins described here https://github.com/Vizzuality/marxan-cloud/pull/621#discussion_r744485106
- "Project owner" -> role Owner for resource Project
- "Project contributor" -> role Contributor for resource Project
- "Project reviewer (read-only)" -> role Viewer for resource Project
- "Project solution-viewer [nice-to-have]" -> role Viewer for resource "Scenarios" or "Solutions"

Do you see anything missing?

docs/features/role-based-access-control/high-level-design.md
ikas: I'm not sure I understand your question... Why are the roles you described overlapping? They are different resources (solutions vs scenarios), so they would have to be subject to different permission control, right? You could, for instance, have role "Viewer" for solutions, and role "Owner" for scenarios, or any other combination of roles...

Composition/inheritance of permissions only comes into play when you don't explicitly set anything, so if you want to explicitly allow something and deny others, you can do it in the corresponding DB tables: for instance, you could be "Owner" of a "solution" (for instance) and just "Viewer" of a scenario (though the current document is not contemplating permissions for solutions)

docs/features/role-based-access-control/brief.md
kgajowy: sounds good! @ikas

docs/features/role-based-access-control/high-level-design.md
ikas: I didn't we would be **restricting** ourselves to controller level :)

I was using controller actions as examples (since it will very likely be the most common use case), but the main idea is that **the caller** (whatever that is, controller, service, function, etc) has the responsibility to correctly check permissions - in no way that is done automatically.

And the way to do it is by calling the service method from the permission module - `this.accessControlService.isOwner(resourceId, userId)` - a good example of this is how we are applying permissions to publish projects: https://github.com/Vizzuality/marxan-cloud/pull/610/files#diff-02daf12463952aecea26345f1a5364a7fcff93b7658cc1478c80e01bbde80c0aR40. In this case, the controller has the responsibility to check the permissions - this is what I meant with the sentence above.

Does this clarify your comment?

docs/features/role-based-access-control/high-level-design.md
ikas: Makes sense, I'll move it here 👍

docs/features/role-based-access-control/high-level-design.md
ikas: I see your point, but keep the hierarchy in mind: it would be hard to keep things self-contained because when checking permissions to access projects, for instance, you also might need to check organizations. The same happens for solutions/scenarios, you could also have to check projects and organizations.

Because of that, I think that keeping separate modules for each resource would leave to an entangle of services from different modules calling each other, which is something that in the long run can also become hard to maintain IMO.

Not sure what's the best approach here - @hotzevzl do you have any thoughts?

docs/features/role-based-access-control/high-level-design.md
ikas: Same comment as above https://github.com/Vizzuality/marxan-cloud/pull/621#discussion_r744575478

docs/features/role-based-access-control/high-level-design.md
ikas: I am describing an interface here - what is your suggestion?

docs/features/role-based-access-control/high-level-design.md
ikas: Though I agree with you that this document should be high-level, I think this description is important because it describes the hierarchical nature of the permissions.

Maybe I can also add some visuals explaining how other modules would use permissions, would that help?

docs/features/role-based-access-control/high-level-design.md
kgajowy: @ikas so, solutions are different resource ? It wasn't explicite said anywhere ;) 

we may also try another example:
* role A allows to edit inclusion/exclusion of PU within scenario
* role B allows to edit cost surface within scenario
* role C allows full edit on any scenario aspect
* ... and so on.

I.e. the scope (A & B) isn't overlapping (but related to the same resource - examples are there just to visualize, don't focus on those too much). How would the proposed approach support this?

docs/features/role-based-access-control/high-level-design.md
kgajowy: @ikas as per https://github.com/Vizzuality/marxan-cloud/pull/621#discussion_r744587371 - "viewer" may have different aspects enabled/disabled , so I am afraid at some point we won't be able to refrain to have multiple "project/scenario viewer" roles.

docs/features/role-based-access-control/high-level-design.md
kgajowy: >  you also might need to check organizations
thats true, so a main exposed facade could hide those required calls to other modules. 

> would leave to an entangle of services from different modules calling each other
They should not, as above. 

> in the long run can also become hard to maintain 
both approaches may become hard to maintain - more complex (having knowledge of everything) will most likely become hard quite fast.
In the end, this is a compromise about having separated responsibilities and more control of maintaining vs having a bit few more calls.

docs/features/role-based-access-control/high-level-design.md
kgajowy: `, but the main idea is that the caller`

it depends on who you meant to be the caller - controller rather shouldn't bother about it, it should be a service responsibility to ensure that other modules using some feature also include this check.

docs/features/role-based-access-control/high-level-design.md
kgajowy: @ikas put the actual code in the repository ;)

docs/features/role-based-access-control/high-level-design.md
kgajowy: @ikas I am fine to keep the hierarchical nature of permissions (as long as we really want it) but move (or rather, remove) the technical details about particular entities.

docs/features/role-based-access-control/brief.md
ikas: Added a separate "development plan" doc describing the release plan for role based access control

docs/features/role-based-access-control/high-level-design.md
ikas: Changed implementation of platform admins to using a dedicated DB table for the purpose.

Also, added "development plan" document, identifying platform admins will only be implemented after the demo.

docs/features/role-based-access-control/high-level-design.md
ikas: Rephrased this section to make it clearer. However, the main point is of this sentence is: **permissions are not checked automatically**. They need to be explicitly verified in all places where we access resources. And the responsibility of doing that is on whoever implements the access to said resources. It's irrelevant where that access happens (controller, service, etc etc), but wherever it happens, permissions need to be checked explicitly by the caller.

kgajowy: sounds good!

docs/features/role-based-access-control/high-level-design.md
ikas: Yeah, having a main exposed facade and then responsibilities split across sub modules is the best idea, I'll rewrite this section to reflect that :+1:

docs/features/role-based-access-control/high-level-design.md
ikas: Rewriting this too according to https://github.com/Vizzuality/marxan-cloud/pull/621#discussion_r744681803

docs/features/role-based-access-control/high-level-design.md
ikas: Oh, I see... I assumed we wanted to refrain from all-things coding related before agreeing on the implementation, but in this case I think that's a good idea to better understand the interface, I'll do that, thanks :+1:

docs/features/role-based-access-control/high-level-design.md
ikas: Hum, I assumed scenarios and solutions would be the same thing, I think that's what I got wrong :sweat_smile:

Let me have a think and circle back!

docs/features/role-based-access-control/high-level-design.md
kgajowy: @ikas it could be a part of Low-level-design doc but as we already touched it, why not? also, abstract classes/interfaces aren't any real implementation ;)

docs/features/role-based-access-control/brief.md
hotzevzl: +1 to moving the requirements to the repo eventually (I have marked the google doc as archived in the past and removed editing access to avoid accidental new discussions once things have been settled and moved to the repo)

docs/features/role-based-access-control/brief.md
```diff
@@ -0,0 +1,40 @@
...
+can perform in the platform. Users can then be associated with roles for specific resources, which will grant them the 
+role privileges for that specific resource.
+
+The current implementation is based on the requirements described in the [User requirements document](https://docs.google.com/document/d/16x4wQYTKov0JkO1ue6iQzJ3-RXpdgQ38ygGJWW86Zo0/edit#heading=h.58xbt2es73r2).
+
+## Roles
+
+In the context of the Marxan platform, we will consider the following roles:
+
+* **Admin**: a generic platform administrator, who can perform **any action for any resource** in the platform.
```
hotzevzl, line 15: I recommend to split this role from the others, to avoid confusion as it will apply only to the platform as a whole rather than to specific resources such as orgs/projects/scenarios. The current explanation says so, but I'd even treat it as part of a different list, for clarity.

hotzevzl: Also, I recommend to make this role orthogonal to the other roles. The working document may not be fully clear on this, but my recommendation is that the concerns (and access rights) of a platform admin should be mostly distinct from those of entity owners/contributors/viewers/etc.

In other words, I would recommend *not* to make this a "root" role but a role with given permissions on the platform but not implicit nor explicit permissions on entities, especially projects and scenarios, *except* for being able to see title/description of projects and potentially delete them, but not necessarily being able to see project content and content of scenarios within projects.

This is partly a matter of personal preference from past experience (playing the greybeard card here 😜) but also from conversations with TNC, I know that there may be concerns about "root" users (besides the unavoidable ones such as infra admins)

docs/features/role-based-access-control/brief.md
```diff
@@ -0,0 +1,40 @@
...
+* **Owner**: represents a user that can perform any management actions for a specific resource of the platform (e.g. 
+a project). This role is typically attributed to the user who created the resource (i.e. they own the resource).
```
hotzevzl, line 17: tiny suggestion: `typically attributed initially`, to clarify that this would be the initial/default case, but more owners may be added later on (or an owner may add another one and remove themselves, for example)

docs/features/role-based-access-control/brief.md
```diff
@@ -0,0 +1,40 @@
...
+We will consider the following entities as resources subject to RBAC:
+
+1. Organizations
+2. Projects
+3. Scenarios
+
+The entities stated above will work in a hierarchical way: roles granted to users in organizations also apply to all 
+projects owned by said organization, as well as all scenarios in all projects owned by the organization.
+
+However, this hierarchical behavior can be overwritten in a case-by-case basis if needed - e.g. a given user can be
```
hotzevzl, line 36: is it worth specifying if there is an expectation (which I don't think we do have nor should have) that roles can be "narrowed", for example, granting some user a *contributor* role on an organization but restricting their role to *viewer* on a subset of the projects of the organization, while leaving the default role cascade as default (so the user would be *contributor* by default on any existing and new project within the organization, except those where their role has been reduced to *viewer*).

I would not recommend even planning this at this stage (there are several implications, such as for example being able to fully revoke any role on a specific child resource, or how to deal with cascades in many-to-many relationships between entities), but either way, I recommend to explicitly exclude this if we want to do so.

docs/features/role-based-access-control/brief.md
```diff
@@ -0,0 +1,40 @@
...
+3. Scenarios
+
+The entities stated above will work in a hierarchical way: roles granted to users in organizations also apply to all 
+projects owned by said organization, as well as all scenarios in all projects owned by the organization.
+
+However, this hierarchical behavior can be overwritten in a case-by-case basis if needed - e.g. a given user can be 
+"contributor" for an organization, but have "owner" privileges for a specific project inside (or outside) that 
+organization.
+
+**Keep in mind that permissions have to be explicitly granted for a given user to have access to a given resource.**
```
hotzevzl, line 40: Does this not contradict what we say two sentences earlier? ("roles granted to users in organizations also apply to all projects owned by said organization, as well as all scenarios in all projects owned by the organization")

Either way, this may not be fully relevant right now given that there is no plan to apply different permissions on a per-scenario basis for the foreseeable future, but also given my other note above about cascades in M-to-M setups, things may become tricky in such contexts (not an argument to avoid it, but if so, I suggest to clarify what is the expected behaviour)

docs/features/role-based-access-control/high-level-design.md
hotzevzl: looking at this from a different angle, I recommend to not hardcode in the design the assumption that different roles are supersets or subsets of another role.

I think that avoiding this assumption can give more flexibility in the modelling of access control, at the expense of more verbosity in the "roles X permissions" setup

That is, if we cannot assume that "Owner" is a proper superset of "Contributor", then we'd need to assign to "Owner" any permissions that we want the role to have from those of "Contributor", which may even be all of them, but since we'd do this once and we will only have a small handful of roles, I think the benefits of allowing for only partly-intersecting sets outweigh the convenience of defining roles as supersets/subsets.

docs/features/role-based-access-control/brief.md
hotzevzl: Ok, I should have read the whole PR first, sorry :D

This is clear from the HLD document. I'd still add something here to clarify this at the outset at the brief level - up to you.

docs/features/role-based-access-control/high-level-design.md
```diff
@@ -0,0 +1,81 @@
...
+  * user with `Owner` role can perform all the actions a `Contributor` can, as well as deleting actions for the given
+resource (which usually includes calls to `DELETE resource/:id` endpoints).
+
+**However, it is up to the functions implementing each feature (modules, controllers, services or any other kind of 
+object in the context of the API) to correctly check permissions for the action being performed.** This also applies
+to custom actions that might be subject to RBAC: for instance, if publishing a project is a custom action that should be
+subject to RBAC, then it is up to the controller action, service method, or wherever we need to do this, to check that 
+the user performing the request has the required role (e.g. `Owner`) to perform that action.
+
+Currently, these are the roles required to perform the following actions in the platform:
```
hotzevzl, line 44: This solves most of the doubts I raised in my review of brief.md 😅

However:

- should we limit permissions of "higher" roles on parent resources (e.g. Organization Owner) to read-only ones for children resources?

Obviously this is not meant as a privilege escalation measure (at least Owners could in any case update roles), but mostly as a least-privilege approach to protect against accidental changes that an Organization Owner, for example, may end up making. For roles that cannot manage roles on a resource this can also provide some protection against privilege escalation.

- should we mark the 'implicit' permissions as such in this table? I am ambivalent about this - in a way, any permission can be seen as just that, a permission, no matter whether the role has "Organization" in their name but a permission given to this role applies to a scenario; on the other hand, if it helps to reason about access levels, it may be useful

hotzevzl: Also, I think we have a chance *right now* to model many-to-many links between at least organizations and projects, which is something that was discussed much earlier on as a wish (for example, project A belongs both to organization M and organization N), with related composition of permissions (if user U has contributor role on organization M and no role in organization N, they can see the project) - later on this may be quite messy to disentangle. Happy to go with one-to-many links in any case if m-to-m adds too much right now, but let's be explicit about this maybe.

docs/features/role-based-access-control/high-level-design.md
hotzevzl: sounds good to me too.

docs/features/role-based-access-control/high-level-design.md
```diff
@@ -0,0 +1,81 @@
...
+## Architecture
+
+A main module `access-control` should be created, which will be the main point-of-contact for checking permissions. 
+Additionally, we will have specific modules for checking permissions for each of the resources supported:
+* `organizations-acl`: module that controls access to organizations;
+* `projects-acl`: module that controls access to projects;
+* `scenarios-acl`: module that controls access to scenarios;
+* `solutions-acl`: module that controls access to solutions;
+
+This `access-control` module will expose a public service, and this is the point of entry that should be used by other
```
hotzevzl, line 73: While I agree with the suggestion to have a single entry point for authorization questions we need to ask, I invite to consider using a different interface whereby instead of asking "Does user U have role R on resource M?" we would ask "Can user U do action A on resource M?".

In practice, I think just checking the role could work for the moment (and it can even be proxied via `canUserDoThis()`-style interface, if we wanted to provide more descriptive semantics). From my experience (greybeard alert!) a system that assigns roles but checks permissions is more flexible, and all in all I tend to prefer some more complexity upfront that can allow to deal with "yes. but about this edge case?" as development progresses (not necessarily in a future phase, even to wrap up the current one). A simpler setup may still well turn out to be just enough for the time being, though.

docs/features/role-based-access-control/high-level-design.md
kgajowy: @hotzevzl I agree, I was trying to get to that point in longer path ;) Inheritance overall may lead to more confusion the more complicated privileges we have (keeping in mind that our system is not JIRA and we don't have such team and years for development)

docs/features/role-based-access-control/high-level-design.md
kgajowy: @hotzevzl isn't it something we already did?

```
async canPublish(userId: string, projectId: string): Promise<boolean>
```
(instead of checking role)

api/apps/api/src/modules/access-control/access-control-service.interface.ts
```diff
@@ -0,0 +1,46 @@
...
+   * This method should return "true" if one of the following conditions apply:
+   *   - the user is a platform admin;
+   *   - the user has a role greater or equal to "Owner" for the resource being checked;
+   *   - the user has a role greater or equal to "Owner" for a resource hierarchically
+   *   superior to the one being checked.
+   *
+   * @param resourceId The id of the resource being checked.
+   * @param userId The id of the user.
+   *
+   * @returns {boolean} True/false if the provided user is owner of the provided resource.
```
kgajowy, line 13:
```
export type Allowed = true;
export type Denied = false;

export type Permit = Allowed | Denied;

type Result = Promise<Permit>

export abstract class AccessControlService {
  abstract canCreateProject(projectId: string, userId: string): Result;
  abstract canViewProject(projectId: string, userId: string): Result;
  abstract canCreateScenario(projectId: string, userId: string): Result;
  abstract canEditScenario(scenarioId: string, userId: string): Result;
  abstract canViewSolutions(scenarioId: string, userId: string): Result;
}
```

@ikas @hotzevzl thoughts about this?

docs/features/role-based-access-control/high-level-design.md
ikas: While I agree this approach is more flexible, one could also argue it's less reusable, harder to maintain, and even redundant, in some cases. For every new action in the platform subject, we will have to add a new method to check permissions for that action. The purpose of having roles (IMO) is exactly to avoid having to check each action individually. Each role has a set of actions they can do (identified in the permissions matrix), and that way we can take advantage of the existing methods to check permissions for a role, instead of having to create a new method for every new action.

However, I'll accept the suggestion, mostly because of the greybeard card :sweat_smile: if flexibility is key, then this approach is more adequate in this case.

docs/features/role-based-access-control/high-level-design.md
ikas: As discussed during today's daily meeting, I'll remove the "hierarchical" bits from this implementation, which will also imply we remove permissions granted implicitly.

This gives us a more flexible approach in the long term, and we still have the possibility of programatically granting permissions with events. Plus, it doesn't constrain us to having to model many-to-many links right now, we can do it later down the road without compromising the existing permissions structure. 

This also addresses your point of "marking implicit permissions in this table" - without implicit permissions in place, this is not a concern because all permissions will have to be explicitly set in the permissions table :+1:

docs/features/role-based-access-control/high-level-design.md
ikas: As discussed during today's daily meeting, I'll remove the "hierarchical" bits from this implementation.

docs/features/role-based-access-control/brief.md
ikas: As discussed during today's daily meeting, I'll remove the "hierarchical" bits from this implementation,

docs/features/role-based-access-control/high-level-design.md
kgajowy: @ikas there is no silver bullet. As you mentioned, this approach has it downsizes but on the other hand, leaking the  roles lead to multiple modules/places/services to check permissions on their own - and it also will eventually become duplicated many times, which makes the changes much more harder to maintain when compared to proposed solution. Again, there is no silver bullet ;)

docs/features/role-based-access-control/brief.md
ikas: I'll move the requirements we have for now, to set a base for this implementation, and because I assume there won't be many changes from here on.

If in the future they change, I'll update this part of the document.

docs/features/role-based-access-control/brief.md
ikas: Done and done :+1:

docs/features/role-based-access-control/brief.md
ikas: With the recent changes and the removal of the hierarchical bits, this sentence is now the golden-rule of this implementation, so it should not sound contradictory :) plus, concerns related to m-2-m relationships don't apply with the removal of hierarchical bits, so I'll resolve this conversation :)

docs/features/role-based-access-control/high-level-design.md
ikas: @kgajowy during the final PR review, can you check if this still applies? I think with the latest changes to the implementation plan this should not be a problem anymore, but I'd appreciate if you could double check :+1:

docs/features/role-based-access-control/development-plan.md
```diff
@@ -0,0 +1,22 @@
+# Role base access control - development plan
+
+This document describes the initial plan for the implementation of Role based access control for Marxan.
+
+## Release 1 (early December 2021)
```
kgajowy, line 5: so, it assumes we will manually edit roles for the platform?

also: for the demo, it will only be effective for project, while scenario/solutions will still be "public"?

docs/features/role-based-access-control/development-plan.md
```diff
@@ -0,0 +1,22 @@
...
+* Creating the `access-control` module, exposing a facade that allows checking permissions for different resources.
+* Update the already existing `projects-acl` implementation module to implement the defined interface.
+* Create point-of-entry service in the `access-control` module that wraps the `projects-acl` module.
+* Update existing calls of the `projects-acl` module to use `access-control` instead.
+* Call the `access-control` service to check permissions for projects in the required places in the code.
+
+## Release 2 (later)
+
+* Implement platform admin users.
+* Create `organizations-acl` module with a service to check access implementing the IAccessControlService interface
```
kgajowy, line 17: `IAccessControlService` may not exist ; )

docs/features/role-based-access-control/high-level-design.md
```diff
@@ -0,0 +1,76 @@
...
+* "Project reviewer (read-only)" -> role Viewer for resource Project;
+* "Project solution-viewer" -> role Viewer for resource "Solutions";
+
+There is an implicit relationship between platform actions and the supported roles. By default, it is assumed that:
+* a user with role `Viewer` can read the information for the given resource (which usually includes calls to 
+`GET resource` and `GET resource/:id` endpoints).
+* user with role `Contributor` can perform all the actions a `Viewer` can, as well as edition actions for the given
+resource (which usually includes calls to `PATCH/PUT resource/:id` endpoints).
+* user with `Owner` role can perform all the actions a `Contributor` can, as well as deleting actions for the given
+resource (which usually includes calls to `DELETE resource/:id` endpoints).
```
kgajowy, line 26: What about `create` action? (POST project , POST scenario) 

What about `run marxan` action?

docs/features/role-based-access-control/high-level-design.md
```diff
@@ -0,0 +1,76 @@
...
+in the context of the platform) to correctly check permissions for the action being performed.** Checking permissions
+should be done by calling the appropriate function in the Access Control service class for the resource being accessed.
+
+For example, if a user with id `1` is trying to publish a `project` with id `2`, then the `publish` function
+needs to check if that user can publish that project. This can be done by calling the `canPublish` function in the 
+access control service in the `projects-acl` module.
+
+Currently, these are the roles required to perform the following actions in the platform:
+
+| Actions / Roles | Solution Viewer | Solution Contributor | Solution Owner | Scenario Viewer | Scenario Contributor | Scenario Owner | Project Viewer | Project Contributor | Project Owner | Organization Viewer | Organization Contributor | Organization Owner | Platform admin |
```
hotzevzl, line 38: looking at the final shape of the permissions matrix, I think we should add a few more permissions such as _create project_ (should be available to organization contributors?), _create scenario_ (project contributors?) as well as granting of roles (it's explained in the document already but it may be useful to add it to the matrix maybe?) on projects (project owners can do so?) and organizations (organization owners?)

docs/features/role-based-access-control/high-level-design.md
```diff
@@ -0,0 +1,76 @@
...
+* `organizations-acl`: module that controls access to organizations;
+* `projects-acl`: module that controls access to projects;
+* `scenarios-acl`: module that controls access to scenarios;
+* `solutions-acl`: module that controls access to solutions;
+
+This `access-control` module will expose a public service, and this is the point of entry that should be used by other 
+services to control access to resources. The `access-control` module will also contain the interface that should be 
+implemented the services in the resource-specific modules, which these should implement.
+
+This interface is called `IAccessControlService` and it is located in `api/apps/api/src/modules/access-control/access-control-service.interface.ts`.
```
kgajowy, line 72: needs update

hotzevzl, line 72: very minor - this looks like the kind of detail that may go out of sync at some point with code (which is hard to avoid in general, and I wouldn't stress about it - we can review things every now and then), and it actually already is out of sync - I'd remove it altogether, and keep the general points about the implementation strategy above.

docs/features/role-based-access-control/high-level-design.md
hotzevzl: > @hotzevzl isn't it something we already did?
> > async canPublish(userId: string, projectId: string): Promise<boolean>

yep, that's what I meant.

ACL consumers should not need to know about roles, but about subjects and permissions. To be fair, for the time being these abstractions and separations of concerns may be more academic than pragmatic, but depending on the implementation, checking permissions vs checking roles could be a negligible computational overhead.

Things would be quite different if, say, platform admins had an app UI where they could reconfigure which roles are granted which permissions - in this case, as long as no new permissions are added, the "grant roles, check permissions" strategy would allow to reshape roles without code changes. But indeed, this is not a requirement for the time being.

hotzevzl (approved):

docs/features/role-based-access-control/high-level-design.md
kgajowy: I agree, also the initial access points/interfaces/classes are within the PR there already, so easy to find.

docs/features/role-based-access-control/development-plan.md
ikas: Hum, that's a very valid point, we should have some way to edit this - I'll include CRUD actions for `projects-acl` in the scope of the initial release, and add CRUD actions for the remaining resources in the following release :+1:

> it will only be effective for project, while scenario/solutions will still be "public"?

Yes, that is the plan!

docs/features/role-based-access-control/high-level-design.md
ikas: This is a "conventions list", something I believe we will follow in most cases (I'll make this clearer). The detailed permissions matrix (a few lines below) is the place to go to check the comprehensive list of actions for each resource.

I didn't include `create` actions here, as the permissions for them might not be straightforward (i.e. we might want to allow anyone to create projects, or maybe just members of certain organizations, etc etc).

In any case, I'll add the actions you mentioned to the permission matrix :+1:

docs/features/role-based-access-control/high-level-design.md
ikas: Added the actions you mentioned, as well as "run Marxan" as suggested by Kamil in a comment above, thanks both :bow:
* Org owners and contributors can create projects, proj owners and contributors can create scenarios, and so on;
* Owners of the resource can manage roles of users for said resource;
* "run Marxan" can only be done by project owners and contributors.

docs/features/role-based-access-control/high-level-design.md
ikas: Removed this line :+1:

kgajowy (approved):

PR #624, opened 2021-11-08
derive BLM calibration values from min,max range [MARXAN-890]

hotzevzl:
This PR updates the BLM calibration brief describing the current/latest functional requirements, whereby users are asked to supply only a `[min, max]` range, from which we derive the intermediate values on a linear scale.

### DIFF
--- a/docs/features/boundary-length-modifier/brief.md
+++ b/docs/features/boundary-length-modifier/brief.md
@@ -48,15 +48,32 @@ across each of the runs.
 
 ## Requirements
 
-- Users should be able to set their BLM values.
+- Users should be able to set their BLM values, by supplying a `[min, max]`
+  range, from which the BLM calibration process will derive a set of BLM values
+  to use by splitting the domain in equal intervals, according to the formula
+  below.
 
-The set should be limited to a small cardinality: 6 values for the initial
-implementation, but leaving space for adjustments in the future (i.e. it is ok
-to hardcode the size of the set, but changing it in the future should not break
-other assumptions).
+Whereas the user input consists in exactly two values, the derived set of values
+to be used for the calibration process should be limited to a small cardinality:
+6 values for the initial implementation, but leaving space for adjustments in
+the future (i.e. it is ok to hardcode the size of the set, but changing it in
+the future should not break other assumptions).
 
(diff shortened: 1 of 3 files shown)


kgajowy (approved):

docs/features/boundary-length-modifier/brief.md
```diff
@@ -48,15 +48,32 @@ across each of the runs.
...
+  min,
+  (min + ((max-min) / N-1)) * 1,
+  (min + ((max-min) / N-1)) * 2,
+  ...,
+  (min + ((max-min) / N-1)) * N-1
+]
+```
+
+- The 6 values given by default will be `BLM = Z*sqrt(PU area)` (with `Z ∈
+  blmValues(0.001, 100)`).
```
hotzevzl, line 76: @tamaramegan I am not sure about this bit - given the set of values we specified in the previous version of this document (`[0.001, 0.01, 0.1, 1, 10, 100]`), if we now use the same `min` and `max`, we would get `[0.001, 20.0008, 40.0006, 60.0004, 80.0002, 100]` which IMO may look a bit puzzling to a user. Should we ad-hoc this somehow?

kgajowy: may be some log(n) usage within the formula to keep the slope?

hotzevzl: that's the issue Tamara and I raised when the idea of only letting users supply the min, max bounds was raised - depending on the actual range, different scales may likely make sense rather than a linear one, but we also didn't want to add the scale as a possible input (even if "controlled", for example from a dropdown)...

In a way, even a linear scale will get users an idea of where diminishing returns start, but probably with more trial-and error (=more BLM calibration runs). Maybe we could at least round any values that are >1, and keep the fractional values for anything in the `[0,1]` range? @tamaramegan what do you think?

hotzevzl: ok, thanks everyone for your input - after some final discussion with Tamara, we'll keep this as plain as possible, with initial default `[0.001, 100]`, linear scale and splitting the domain in 6 identical intervals.

PR #625, opened 2021-11-08
Minor adjustments to get the project running natively

ikas:
Minor adjustments to get the project running natively.

### DIFF
--- a/api/apps/api/config/custom-environment-variables.json
+++ b/api/apps/api/config/custom-environment-variables.json
@@ -29,7 +29,8 @@
    }
   },
   "api": {
-    "url": "API_SERVICE_URL"
+    "url": "API_SERVICE_URL",
+    "port": "API_SERVICE_PORT"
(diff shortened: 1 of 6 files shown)


kgajowy (commented):
@hotzevzl please take a look - you are more suitable to see what env changes are needed afterwards.

@ikas shouldn't we have some defaults provided through 'config' jsons instead hardcoding them in code?

ikas:
I've updated the PR to use default config values set in `default.json` :+1: 

Re env changes: I'm only making sure that we can change values that were already set in the code through env vars. So, _in theory_, we don't need to change anything in the current env vars (it's always worth having a sanity check though :sweat_smile:)

hotzevzl (approved):
thanks @ikas, the changes make sense and as far as I can see all keeps working as expected when running in containers.

kgajowy (approved):

PR #626, opened 2021-11-09
feat(blm): project defaults skeleton

kgajowy:
This PR includes a "skeleton" of application shape. The main purpose of it is to fulfill requirements related to calculating default BLM values after project has been created (as its grid/planning units). 

Once its merged, the following tasks/pieces should be relatively easy to complete in isolation:
https://vizzuality.atlassian.net/browse/MARXAN-996
https://vizzuality.atlassian.net/browse/MARXAN-997
https://vizzuality.atlassian.net/browse/MARXAN-998
https://vizzuality.atlassian.net/browse/MARXAN-999
https://vizzuality.atlassian.net/browse/MARXAN-1000 (🚀)

### DIFF
--- a/api/apps/api/src/modules/blm/values/project-blm-repository.ts
+++ b/api/apps/api/src/modules/blm/values/project-blm-repository.ts
@@ -10,8 +10,20 @@ export type GetFailure = typeof unknownError | typeof projectNotFound;
 
 export interface ProjectBlm {
   id: string;
+
+  /**
+   * User-supplied or defaults if none provided,
+   */
   range: [number, number];
+
+  /**
+   * Calculated from range.
+   */
   values: number[];
+
+  /**
+   * Sets once at the project creation chain - once PU are known.
+   */
   defaults: number[];
 }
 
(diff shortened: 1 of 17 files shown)


api/apps/api/src/modules/blm/values/project-blm-repository.ts
```diff
@@ -0,0 +1,33 @@
...
+export type CreateFailure = typeof unknownError | typeof alreadyCreated;
+export type SaveFailure = typeof unknownError | typeof projectNotFound;
+export type GetFailure = typeof unknownError | typeof projectNotFound;
+
+export interface ProjectBlm {
+  id: string;
+  range: [number, number];
+  values: number[];
+  defaults: number[];
+}
```
hotzevzl, line 16: minor, but I would find it useful to add some inline documentation at least for this core interface.

A reference to the design documents may be enough, but having at least some core traits of this interface documented for quick in-IDE reference may a way to be kinder to our future selves, e.g. defaults is set-once, values is calculated from range, range is user-supplied... -  all this can also be gleaned from the interfaces in the rest of this PR, of course.

hotzevzl (approved):

api/apps/api/src/modules/blm/values/project-blm-repository.ts
kgajowy: @hotzevzl 👍  https://github.com/Vizzuality/marxan-cloud/pull/626/commits/32a887dcbbf9d986b3ca9284a4c032f4bcfc0a0d

PR #628, opened 2021-11-10
feat(blm): calibration runner

kgajowy:
This part mainly takes care to determine which processor should be attached to which queue (worker works now on abstraction)
image

The interfaces aren't finished - sometimes on purpose to leave some flexibility for implementation.
Also, I haven't yet tested it the right way - will do next week.


hotzevzl (approved):
the plan looks solid, thank you!

From this PR, I take that the main departures from the possible implementation ideas outlined in the HLD are around how we instantiate the runner service and around the use of partial results, which seems a good approach to me.

It may be worth doing a light revisit of the design documents after the fact (once the BLM calibration is complete) to clear up inconsistencies that may trip up future readers (including our future selves).

kgajowy:
> the plan looks solid, thank you!
> 
> From this PR, I take that the main departures from the possible implementation ideas outlined in the HLD are around how we instantiate the runner service and around the use of partial results, which seems a good approach to me.
> 
> It may be worth doing a light revisit of the design documents after the fact (once the BLM calibration is complete) to clear up inconsistencies that may trip up future readers (including our future selves).

yes, hope this is clear enough. May need a bit further moves, as most likely the BLM will use Single- under the hood.

also, added missing parts https://github.com/Vizzuality/marxan-cloud/pull/628/commits/c36cbc94e90e9292ba1e0880d4b0de0f9b84fc68 of attaching relevant adapters (through DynamicModule) connected to respective queue.

PR #629, opened 2021-11-11
Updated functions and data for most different solution

aagm:
### Overview

The details of the implementation are described in the notebook. and in the [task details](https://vizzuality.atlassian.net/browse/MARXAN-810)

Missing bit is reformat how data is passed to the function as we need to also provide the data as:
```typescript
ResultRow & {puValues: number[]}
```
where puValues is an array of the binary solution for that run.

### Testing instructions

`yarn test most-different`

---

- [ ] If this PR adds a feature that should be tested for regressions when
- [ ] Update CHANGELOG file


api/apps/geoprocessing/src/marxan-sandboxed-runner/adapters/solutions-output/most-different.service.ts
```diff
@@ -92,4 +94,39 @@ export class MostDifferentService {
     cluster: SolutionPerClusterGroup<ResultRow | undefined>,
   ): cluster is SolutionPerClusterGroup =>
     cluster.every((element) => isDefined(element));
+
+  #getJaccardBinaryDistance = (setA: number[], setB: number[]): number => {
+    if (setA.length !== setB.length) throw new Error('Need to be equal length');
```
hotzevzl, line 99: in theory we should never get empty sets, but just in case, should we throw an error in this case too? at a quick glance (I may be wrong) such situation may lead to a division by zero on the return statement (which could be ok too - either way is a situation that shouldn't be allowed)

api/apps/geoprocessing/src/marxan-sandboxed-runner/adapters/solutions-output/most-different.service.ts
```diff
@@ -92,4 +94,39 @@ export class MostDifferentService {
...
+    setB: number[],
+    distances: number[][],
+  ): number => {
+    let distance = 0;
+    // methodology for linkage https://en.wikipedia.org/wiki/UPGMA
+    for (const a of setA) {
+      for (const b of setB) distance += distances[a][b];
+    }
+
+    return distance / (setA.length * setB.length);
```
hotzevzl, line 130: same as above for empty sets - do we need to care?

api/apps/geoprocessing/src/marxan-sandboxed-runner/adapters/solutions-output/most-different.service.ts
kgajowy: https://github.com/Vizzuality/marxan-cloud/pull/629/files#diff-6d118bda04a900bbf9d5a32482a01bbe56e1972c45b6cfd4fb1fc8c21158ce76R26-R42

also there are some (possibly) related things to verify.

kgajowy (approved):

PR #630, opened 2021-11-11
use a distinct envvar to set daemon listen ports when running Express natively

hotzevzl:
https://vizzuality.atlassian.net/browse/MARXAN-1028

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,10 @@ details.
    * `API_AUTH_X_API_KEY` (string, required): a secret used as API key for
      requests from the Geoprocessing service to the API; can be generated
      similarly to `API_AUTH_JWT_SECRET`
-   * `API_SERVICE_PORT` (number, required): the port on which the API service
-     should listen on the local machine
+   * `API_SERVICE_PORT` (number, required): the port exposed by Docker for the
+     API service; when running an instance under Docker Compose, NestJS will
+     always be listening on port 3000 internally, and this is mapped to
+     `API_SERVICE_PORT` when exposed outside of the container
    * `API_SERVICE_URL` (URL, optional, default is http://api:3000): the internal
      (docker-compose or k8s cluster) where the API service can be reached by
      other services running in the cluster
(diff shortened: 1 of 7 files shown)


ikas (commented):
I'm wondering though if it wouldn't be simpler not to hardcode `3000` and use the same port in both sides... Something like:

```
ports:
      - "${API_SERVICE_PORT}:${API_SERVICE_PORT}"
```

I'm not sure if this would fix the current problem, but at least locally (using Docker compose) it works fine. But seems like a simpler setup, and we wouldn't need to have 2 separate env vars and 2 separate config values (plus less hardcoded values)

hotzevzl:
> I'm wondering if it would be simpler not to hardcode `3000` and use the same port in both sides... Something like:
> 
> ```
> ports:
>       - "${API_SERVICE_PORT}:${API_SERVICE_PORT}"
> ```
> 
> I'm not sure if this would fix the current problem, but at least locally (using Docker compose) it works fine. Seems like a simpler setup, we wouldn't need to have 2 separate env vars and 2 separate config values, and we would have fewer hardcoded values :)

that's a possible solution, but I'd prefer to keep as they are things that can be ok with defaults (thinking about running these services in k8s pods, for example, or the existing setup for CI and ETL pipelines), and configure things specifically for the non-"production" scenarios (dev environments), if needed.

ikas (approved):
:rocket:

hotzevzl:
> > I'm wondering if it would be simpler not to hardcode `3000` and use the same port in both sides... Something like:
> > ```
> > ports:
> >       - "${API_SERVICE_PORT}:${API_SERVICE_PORT}"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm not sure if this would fix the current problem, but at least locally (using Docker compose) it works fine. Seems like a simpler setup, we wouldn't need to have 2 separate env vars and 2 separate config values, and we would have fewer hardcoded values :)
> 
> that's a possible solution, but I'd prefer to keep as they are things that can be ok with defaults (thinking about running these services in k8s pods, for example, or the existing setup for CI and ETL pipelines), and configure things specifically for the non-"production" scenarios (dev environments), if needed.

on the other hand, that `<SERVICE>_SERVICE_PORT` are required env vars is a bit of a lie - it was true until now because that's needed in the Docker Compose setup, but if running natively, all that one needs to set after this PR are `<SERVICE>_DAEMON_LISTEN_PORT` envvars, so both devs using Docker or native setups only need to configure one thing. I haven't updated the README to reflect this because I ~would like to keep the Docker setup as default~ I am a nasty person 😬

PR #632, opened 2021-11-11
MARXAN-1004 Updates in projects ACL

rubvalave:
### Overview

Updates the "projects-acl" module to implement the interface defined in the "access-control" module document.

There are a couple of methods established in the main access-control module that were used just to avoid any complains by the system but that are not actually fully implemented yet (solutions, scenarios).

Create project access control is for the moment open to both all users and non-users. Future work needed here.

### Feature relevant tickets

[MARXAN-1004](https://vizzuality.atlassian.net/browse/MARXAN-1004)


ikas (changes requested):
The roles we currently have defined don't match the ones defined in the HLD - we should refactor that at this point, otherwise later it will be much more complex!

api/apps/api/src/modules/projects-acl/project-acl.service.ts
```diff
@@ -5,26 +5,89 @@ import { intersection } from 'lodash';
...
   ];
 
   constructor(
     @InjectRepository(UsersProjectsApiEntity)
     private readonly roles: Repository<UsersProjectsApiEntity>,
-  ) {}
+  ) {
+    super();
+  }
+  // TODO: create a proper workflow to check project creation capabilities
```
ikas, line 35:
```suggestion
  // TODO: this will be changed in the following release of user requirements.
  // For now, anyone should be able to create projects, regardless of having roles or not.
  // In the future project creation will be limited to organization contributors, so this logic will be moved to the access control module
```

api/apps/api/src/modules/projects-acl/project-acl.service.ts
```diff
@@ -5,26 +5,91 @@ import { intersection } from 'lodash';
...
+
+  async canViewProject(userId: string, projectId: string): Promise<Permit> {
+    const roles = (
+      await this.roles.find({
+        where: {
+          projectId,
+          userId,
+        },
+      })
+    ).flatMap((role) => role.roleName);
```
kgajowy, line 50: Very minor but we don't have to select whole `role` entity but just the roleName.

api/apps/api/src/modules/projects-acl/project-acl.service.ts
```diff
@@ -5,26 +5,91 @@ import { intersection } from 'lodash';
...
+  }
+  async canEditScenario(userId: string, projectId: string): Promise<Permit> {
+    const roles = (
+      await this.roles.find({
+        where: {
+          projectId,
+          userId,
+        },
+      })
+    ).flatMap((role) => role.roleName);
```
kgajowy, line 88: Another minor: could be a good fit to extract this call to standalone private function like `getUserRolesWithinProject(...)` to avoid repetition.

api/apps/api/src/modules/projects-acl/project-acl.service.ts
```diff
@@ -5,35 +5,102 @@ import { intersection } from 'lodash';
...
-  constructor(
-    @InjectRepository(UsersProjectsApiEntity)
-    private readonly roles: Repository<UsersProjectsApiEntity>,
-  ) {}
-
-  async canPublish(userId: string, projectId: string): Promise<boolean> {
-    const roles = (
-      await this.roles.find({
+  private async getRolesWithinProject(
+    roles: Repository<UsersProjectsApiEntity>,
```
kgajowy, line 30: can just use `this.roles` instead

kgajowy (approved):
✅ from code perspective

api/apps/api/src/modules/access-control/projects-acl/project-access-control.ts
```diff
@@ -0,0 +1,21 @@
+export type Allowed = true;
+export type Denied = false;
+export type Permit = Allowed | Denied;
```
hotzevzl, line 3: I imagine we'll end up using these types in other `<ResourceType>AccessControl` services: maybe worth moving them to their own file and importing them here?

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+  await fixtures.ThenCanCreateProject();
+  await fixtures.ThenCanViewProject();
+});
+
+test(`project contributor role assigned`, async () => {
+  fixtures.GivenProjectContributorRoleIsAssigned();
+  await fixtures.ThenCannotPublishProject();
+  await fixtures.ThenCanCreateProject();
+  await fixtures.ThenCanViewProject();
+});
```
hotzevzl, line 44: I LOVE this

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -0,0 +1,82 @@
...
+  private readonly canEditScenarioRoles = [Roles.project_owner];
+  private readonly canViewSolutionRoles = [Roles.project_owner];
+  private readonly canPublishProjectRoles = [Roles.project_owner];
+  private readonly canViewProjectRoles = [
+    Roles.project_owner,
+    Roles.project_contributor,
+    Roles.project_viewer,
+  ];
+
+  private async getRolesWithinProject(
```
hotzevzl, line 27: you may have noticed already that I am fastidious about naming things 😱  - tiny detail, and feel free to ignore: what about the slightly more verbose name below?

```suggestion
  private async getRolesWithinProjectForUser(
```

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -0,0 +1,82 @@
...
+  constructor(
+    @InjectRepository(UsersProjectsApiEntity)
+    private readonly roles: Repository<UsersProjectsApiEntity>,
+  ) {}
+
+  // TODO: this will be changed in the following release of user requirements.
+  // For now, anyone should be able to create projects, regardless of having roles or not.
+  // In the future project creation will be limited to organization contributors, so this logic will be moved to the access control module
+  async canCreateProject(_userId: string, _projectId: string): Promise<Permit> {
+    return true;
```
hotzevzl, line 52: though structurally equivalent to `true` and `false` by definition, should we make use of `Allow` and `Deny` in these ACL functions, since we're using `Permit` in the function signature?

this would however imply a slightly more verbose syntax when we check the intersection of user roles with roles who can carry out an action.

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  async canViewProject(userId: string, projectId: string): Promise<Permit> {
+    const roles = await this.getRolesWithinProject(userId, projectId);
+
+    return intersection(roles, this.canViewProjectRoles).length > 0;
```
hotzevzl, line 58: this is very personal opinion/preference - I too do expose some raw "computation" snippets in my own code, but these lines repeated in each can*Project() method to me seem to be screaming asking to be wrapped in a more expressively-named function that we can use in each can<Action><Resource>() method across the app.

```suggestion
    return doesUserHaveRole(
        await this.getRolesWithinProject(userId, projectId),
        this.canViewProjectRoles) ? Allow : Deny;
```

What I mean, repeating the mechanics of checking the cardinality of set intersections sounds tedious. If we can identify a good enough abstraction for the time being, I'd abstract it.

But again, it's also partly a matter of preference.

(Also, it can be abstracted in several ways - above I have suggested a light touch, but the function could also be something like `doUserRolesAllow(roles: Roles[], wantedRoles: Role[]): Promise<Permit>`, for example)

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -0,0 +1,82 @@
...
+  private readonly canViewProjectRoles = [
+    Roles.project_owner,
+    Roles.project_contributor,
+    Roles.project_viewer,
+  ];
+
+  private async getRolesWithinProject(
+    userId: string,
+    projectId: string,
+  ): Promise<Array<string>> {
```
hotzevzl, line 30: Should we keep stricter typing here and return `Promise<Array<Role>>` instead?

hotzevzl (approved):

PR #634, opened 2021-11-12
Access control module

ikas:
### Overview

This PR adds the `access-control` module, which is the main point of entry for permission check in the application. It also changes the existing ACL checks across the application to use the `AccessControlService` to check if a given user can perform certain actions.

This PR does not include (yet):
* Tests for the access control module
* Adding the remaining checks for the other actions that should be subject to access control (like editing or deleting a project).

### Testing instructions

Full testing instructions will be added to #632 

### Feature relevant tickets

[MARXAN-1003](https://vizzuality.atlassian.net/browse/MARXAN-1003)


api/apps/api/src/modules/access-control/access-control.module.ts
```diff
@@ -0,0 +1,11 @@
...
+
+import { AccessControlService } from './access-control.service';
+import { ProjectAclModule } from '@marxan-api/modules/projects-acl';
+
+@Module({
+  imports: [ProjectAclModule],
+  providers: [AccessControlService],
+  exports: [AccessControlService],
+})
+export class AccessControlModule {}
```
kgajowy, line 11: Think of:

```
// keep abstract class
abstract class ProjectAccessControl {}

// implementation
export class ProjectAclService extends ProjectAccessControl { }

// inject
@Module({
  providers: [
    {
         provide: ProjectAccessControl,
         useClass: ProjectAclService
    }
  ],
  exports: [ProjectAccessControl]
})
```

Thanks to that we may be able to get rid of "dummy" facade.

Also "types" of response would be still within module contract instead of particular implementation.

ikas: sounds like a better approach, thanks!

ikas: @kgajowy https://github.com/Vizzuality/marxan-cloud/pull/634/commits/5dd9ac0e3a3d4a482286538125e23839ea9725a2

ikas:
Will merge this PR into #632 and ask for further reviews there :)

api/apps/api/src/modules/projects-acl/project-acl.module.ts
```diff
@@ -1,17 +1,10 @@
...
 
 @Module({
-  imports: [
-    CqrsModule,
-    // this entity most likely shouldn't be under `projects`
-    TypeOrmModule.forFeature([UsersProjectsApiEntity]),
-  ],
+  imports: [CqrsModule],
   providers: [ProjectAclService],
   exports: [ProjectAclService],
```
kgajowy, line 9: Shouldn't it export `ProjectAccessControl` ?

ikas: yeah, true, fixing!

api/apps/api/src/modules/projects-acl/project-acl.service.ts
```diff
@@ -5,17 +5,18 @@ import { intersection } from 'lodash';
 import { UsersProjectsApiEntity } from '@marxan-api/modules/projects/control-level/users-projects.api.entity';
 import { Roles } from '@marxan-api/modules/users/role.api.entity';
 import { Injectable } from '@nestjs/common';
+
 import {
-  AccessControlService,
+  ProjectAccessControl,
   Permit,
-} from '../access-control/access-control-service.interface';
+} from '@marxan-api/modules/access-control/project-access-control';
```
kgajowy, line 12: Circular module imports/dependency

kgajowy (changes requested):
Let's get rid of it! 💪

PR #639, opened 2021-11-18
Add calculation of PU usage for each run

ikas:
### Overview

This PR adds the calculation of PU usage on a per run basis to the parsing of calculation results.

### Feature relevant tickets

[MARXAN-810](https://vizzuality.atlassian.net/browse/MARXAN-810)

### DIFF
--- a/api/apps/geoprocessing/src/marxan-sandboxed-runner/adapters/solutions-output/result-parser.service.ts
+++ b/api/apps/geoprocessing/src/marxan-sandboxed-runner/adapters/solutions-output/result-parser.service.ts
@@ -3,7 +3,6 @@ import { plainToClass } from 'class-transformer';
 import { ExecutionResult, ResultRow } from '@marxan/marxan-output';
 import { isDefined } from '@marxan/utils';
 import { validateSync } from 'class-validator';
-import { chunk } from 'lodash';
 
 import { MostDifferentService } from './most-different.service';
 import { BestSolutionService } from './best-solution.service';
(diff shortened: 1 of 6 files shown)


api/apps/geoprocessing/src/marxan-sandboxed-runner/adapters/solutions-output/result-parser.service.ts
```diff
@@ -81,46 +75,12 @@ export class ResultParserService {
...
-              // `MostDifferentService`.
-              distinctFive: false,
-              puValues: planningUnitSelection['1'].values.map((b) =>
-                b ? 1 : 0,
-              ),
-            });
+
+            const entry = ResultParserService.calculatePuValues(
+              ResultParserService.parseRowToResultRow(row),
+              planningUnitSelection,
```
kgajowy, line 81: I may be not noticing something obvious - sorry 😬 - is there any reason why we cannot "just" assign that value within `parseRowToResultRow`? If possible, would be a bit more straightforward to understand.

kgajowy:
Also:
image

kgajowy:
E2E on real marxan run seems to fail as well during parsing:
https://github.com/Vizzuality/marxan-cloud/runs/4250556227?check_suite_focus=true#step:4:22285

api/apps/geoprocessing/src/marxan-sandboxed-runner/adapters/solutions-output/result-parser.service.ts
ikas: I was aiming at maintaining the existing behavior, where we only set `puValues` in the second call of this function (line 120 of the old file), whereas in the initial call (line 64 of the old file), `puValues` was initialized as empty - but I agree this way doing this is more confusing :upside_down_face:

I'll apply your suggestion and I'll set `puValues` straight away on `parseRowToResultRow` and check the tests - if nothing breaks, we should be ok doing this :) (unless ofc you think we should keep the previous behavior!)

ikas:
Feedback has been addressed, will merge the PRs and have a look at why tests are failing in the target PR (#629)

PR #640, opened 2021-11-18
Feat marxan 1005 endpoint permission projects

rubvalave:
### Feature relevant tickets

[MARXAN-1005](https://vizzuality.atlassian.net/browse/MARXAN-1005)


api/apps/api/src/app.module.ts
```diff
@@ -65,6 +66,7 @@ import { BlmValuesModule } from '@marxan-api/modules/blm';
     ScenarioSpecificationModule,
     PublishedProjectModule,
     BlmValuesModule,
+    ProjectAclModule,
```
kgajowy, line 69: `project-acl` module isn't indented to be used directly anywhere (also take look at the import path).

api/apps/api/src/modules/access-control/projects-acl/dto/update-user-role-project.dto.ts
```diff
@@ -0,0 +1,14 @@
+import { Roles } from '@marxan-api/modules/users/role.api.entity';
```
kgajowy, line 1: It might be a good moment to move this entity directly to `access-control` module and get rid of its dependencies in other modules. Reasoning: we have dedicated module now, so there isn't any reason left to keep polluting other modules (like projects, users) with anything about roles or access control.

api/apps/api/src/modules/access-control/projects-acl/dto/update-user-role-project.dto.ts
```diff
@@ -0,0 +1,14 @@
...
+  @IsUUID()
+  userId!: string;
+
+  @IsString()
+  @IsEnum(Object.values(Roles))
+  roleName?:
+    | Roles.project_viewer
+    | Roles.project_contributor
+    | Roles.project_owner;
+}
```
kgajowy, line 14: 1 Shouldn't it include the `projectId` we are changing role for?
2 also, `role?: Roles` should be enough unless we want to restrict which roles are possible

api/apps/api/src/modules/access-control/projects-acl/project-acl.controller.ts
```diff
@@ -0,0 +1,80 @@
...
+import { ProjectAclService } from './project-acl.service';
+import { RequestWithAuthenticatedUser } from '@marxan-api/app.controller';
+import { ApiBearerAuth, ApiTags } from '@nestjs/swagger';
+import { UsersProjectsApiEntity } from '@marxan-api/modules/projects/control-level/users-projects.api.entity';
+import { UpdateUserRoleinProject } from './dto/update-user-role-project.dto';
+
+@UseGuards(JwtAuthGuard)
+@ApiBearerAuth()
+@ApiTags('Projects Users')
+@Controller(`${apiGlobalPrefixes.v1}/projects`)
```
kgajowy, line 24: A couple of thoughts there:

1. `/projects` - it will most likely override other controller (not sure Nest can "merge" routes)
2. `/projects` isn't the resource we change in REST terms (we change privileges/roles)
3. Sub-modules should be rather "clean" in terms of functionality and not bind to particular transport layer. Think of having them "pure" for doing one thing - access to roles within a project in this case
4. Thus, I think it *could* be good to move it to `access-control` or even outside of the whole `access-control` module. (as controller is "only" a way to access this module capabilities for system consumers)

api/apps/api/src/modules/access-control/projects-acl/project-acl.controller.ts
```diff
@@ -0,0 +1,80 @@
...
+export class ProjectAclController {
+  constructor(private readonly projectAclService: ProjectAclService) {}
+
+  @Get(':projectId/users')
+  async findUsersInProject(
+    @Param('projectId', ParseUUIDPipe) projectId: string,
+    @Req() req: RequestWithAuthenticatedUser,
+  ): Promise<UsersProjectsApiEntity[]> {
```
kgajowy, line 32: Controller should never operate on Entities - dto/entity/pojos have different purpose and even tho it may seem overengineering, we don't want to force REST API be that coupled to entities.

api/apps/api/src/modules/access-control/projects-acl/project-acl.controller.ts
```diff
@@ -0,0 +1,80 @@
...
+    const isOwner = await this.projectAclService.checkUserIsOwner(
+      req.user.id,
+      projectId,
+    );
```
kgajowy, line 36: Seems like `controller` is keeping some logic (check Acl, find something) which would be great to move to the service.

Controller is only one of the ways to access the feature. If any other module would like to get the usersIds within a project, would it need to use it through controller? otherwise, it would skip the "check" the module is developed for.

api/apps/api/src/modules/access-control/projects-acl/project-acl.controller.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    if (!isOwner) {
+      throw new ForbiddenException();
+    }
+
+    await this.projectAclService.updateUserInProject(projectId, dto);
```
kgajowy, line 60: as above.

api/apps/api/src/modules/access-control/projects-acl/project-acl.module.ts
```diff
@@ -0,0 +1,20 @@
...
+@Module({
+  imports: [
+    CqrsModule,
+    // this entity most likely shouldn't be under `projects`
+    TypeOrmModule.forFeature([UsersProjectsApiEntity]),
+  ],
+  providers: [ProjectAclService],
+  exports: [ProjectAclService],
+  controllers: [ProjectAclController],
+})
```
kgajowy, line 19: Like for roles, seems we should move the `UsersProjectsApiEntity` under access control module and get rid of relation.

kgajowy: also, keeping in-code relation to `users` causes kind of circular relation/dependency. We don't want the access-control to care about particular users, not even to be able to identify them - why would we?

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -79,4 +80,56 @@ export class ProjectAclService implements ProjectAccessControl {
...
+        projectId,
+        userId,
+        roleName,
+      });
+    } else {
+      await this.roles.update({ projectId, userId }, { roleName });
+    }
+  }
+
+  async deleteUserFromProject(
```
kgajowy, line 129: as we are within access control domain, the right word would be `revokeAccess`

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -79,4 +80,56 @@ export class ProjectAclService implements ProjectAccessControl {
...
+    });
+    if (!existingUserInProject) {
+      await this.roles.save({
+        projectId,
+        userId,
+        roleName,
+      });
+    } else {
+      await this.roles.update({ projectId, userId }, { roleName });
+    }
```
kgajowy, line 126: It may be rare but imagine two parallel requests hitting:
- one to delete
- second to update

so, it may happen that the `update` first part (getting existingUserInProject) executes, then the delete happens, then we are back to update user... which may not work as we expect.

Most suitable would be to wrap find/update within transaction, locking the row till finished.

api/apps/api/test/access-control/project-acl.fixtures.ts
```diff
@@ -0,0 +1,41 @@
...
+  const app = await bootstrapApplication();
+  const randomUserToken = await GivenUserIsLoggedIn(app);
+  const publishedProjectsRepo: Repository<PublishedProject> = app.get(
+    getRepositoryToken(PublishedProject),
+  );
+  const cleanups: (() => Promise<void>)[] = [];
+
+  return {
+    cleanup: async () => {
+      await Promise.all(cleanups.map((clean) => clean()));
```
kgajowy, line 19: most likely you would want to `reverse` cleanup (like, creating project->scenario->roles would have intention to be "cleaned" in reverse way, starting with roles, not project)

api/apps/api/test/access-control/project-acl.fixtures.ts
```diff
@@ -0,0 +1,41 @@
+import * as request from 'supertest';
+import { bootstrapApplication } from '../utils/api-application';
+import { GivenUserIsLoggedIn } from '../steps/given-user-is-logged-in';
+import { GivenProjectExists } from '../steps/given-project';
+import { Repository } from 'typeorm';
+import { getRepositoryToken } from '@nestjs/typeorm';
+import { PublishedProject } from '@marxan-api/modules/published-project/entities/published-project.api.entity';
```
kgajowy, line 7: Minor, but we may want to create some utility step (or extend the current one) to avoid accessing another unrelated module directly

api/apps/api/test/access-control/project-acl.fixtures.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export const getFixtures = async () => {
+  const app = await bootstrapApplication();
+  const randomUserToken = await GivenUserIsLoggedIn(app);
+  const publishedProjectsRepo: Repository<PublishedProject> = app.get(
```
kgajowy, line 12: Also, it is not used - what was your intention there?

kgajowy (changes requested):
Blocking mainly due to submodule mixed responsibilities and failing CI

api/apps/api/test/access-control/project-acl.fixtures.ts
rubvalave: What do you mean here? Doesn't `cleanup` just clean everything up in the DB? Or am I understanding this wrong?

I didn't make any changes in here as it looks like it was working properly (deleting what I need for every test on every cycle). But if you can explain it a little bit I would appreciate it very much 🙇

api/apps/api/src/modules/access-control/projects-acl/project-acl.module.ts
rubvalave: The thing is that this entity was used also in a couple of places around the projects directory to be able to do some extended queries to the DB while looking for some particular users. Is this not needed anymore with the ACL-module? I was going to move it, but it involved changing `project-module` and some other files and I wanted to be careful before doing so.

api/apps/api/src/modules/access-control/projects-acl/project-acl.controller.ts
```diff
@@ -0,0 +1,56 @@
...
+@Controller(`${apiGlobalPrefixes.v1}/roles/projects`)
+export class ProjectAclController {
+  constructor(private readonly projectAclService: ProjectAclService) {}
+
+  @Get(':projectId/users')
+  async findUsersInProject(
+    @Param('projectId', ParseUUIDPipe) projectId: string,
+    @Req() req: RequestWithAuthenticatedUser,
+  ): Promise<UserRoleInProjectDto[]> {
+    await this.projectAclService.checkUserIsOwner(req.user.id, projectId);
```
ikas, line 31: I think what Kamil meant in [this comment](https://github.com/Vizzuality/marxan-cloud/pull/640#discussion_r752952683) - and I agree - is that we should move the `checkUserIsOwner` call to **inside** the service methods themsleves (like, e.g. inside `findUsersInProject`).

With this, we are making sure that every time `findUsersInProject` is called, the `checkUserIsOwner` validation is performed without the caller having to worry about this :) plus, the controller becomes even cleaner :ok_hand:

api/apps/api/src/modules/access-control/projects-acl/project-acl.controller.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    return await this.projectAclService.findUsersInProject(projectId);
+  }
+
+  @Patch('/users')
```
ikas, line 36: Any reason why we're not using `projectId` as part of the route here? It would be more consistent with the other endpoints in this controller :)

api/apps/api/src/modules/access-control/projects-acl/project-acl.controller.ts
```diff
@@ -0,0 +1,56 @@
...
+    @Body() dto: UserRoleInProjectDto,
+    @Req() req: RequestWithAuthenticatedUser,
+  ): Promise<void> {
+    await this.projectAclService.checkUserIsOwner(req.user.id, dto.projectId);
+
+    await this.projectAclService.updateUserInProject(dto);
+  }
+
+  @Delete(':projectId/users/:userId')
+  async deleteUserFromProject(
```
ikas, line 47: Same comment as above, this endpoint should also return 204

api/apps/api/test/access-control/project-acl-get.e2e-spec.ts
```diff
@@ -0,0 +1,50 @@
...
+  const response = await fixtures.WhenGettingProjectUsersAsNotOwner(projectId);
+  fixtures.ThenForbiddenIsReturned(response);
+});
+
+test(`getting project users as the owner`, async () => {
+  const projectId = await fixtures.GivenProjectWasCreated();
+  const response = await fixtures.WhenGettingProjectUsersAsOwner(projectId);
+  fixtures.ThenSingleOwnerUserInProjectIsReturned(response);
+});
+
```
ikas, line 24: Let's add another (or 2) E2E test(s) here making sure that users with role `project_viewer` or `project_contributor` are not able to get project users :)

ikas (changes requested):
We're close :muscle: overall looks great, awesome job :clap: 

I left some comments with some suggestions of improvements, let me know if you need any help with them :)

api/apps/api/src/modules/access-control/projects-acl/project-acl.controller.ts
rubvalave: It's mainly because the DTO before didn't include the `projectId` and Kamil asked why it shouldn't. I included it on the `DTO` and figured it was not necessary in the `param` anymore (as it was not fetched from there) and the `updateUserInProject` method in the service was using the dto itself. But do you think for consistency I should add it back?

api/apps/api/src/modules/access-control/projects-acl/project-acl.controller.ts
rubvalave: Oh damn, good idea, but I didn't understood it well. Thanks for pointing it out!

api/apps/api/src/modules/access-control/projects-acl/project-acl.module.ts
kgajowy: We can shift it to another PR but in general yes, `project` itself shouldn't care much. As you wrote, there is a challenge how to make listing projects, for example.

api/apps/api/test/access-control/project-acl.fixtures.ts
kgajowy: @rubvalave I like the sandwich process creation as an example:

bread > butter > cheese > ham > tomato

now if you will try to remove bread first (as in the code), everything else may fall on the floor - thats why cleanup `reverse` (in terms of order) is more safe to execute.

api/apps/api/src/modules/access-control/projects-acl/project-acl.controller.ts
kgajowy: Yes, keeping those alike seems a good reason.

I may overlooked that DTO does not have it from `service` perspective - as it needs it to complete the check. 

So - in controller layer, `projectId` may remain in path, while the data passed down to service layer should include it.

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -92,4 +93,53 @@ export class ProjectAclService implements ProjectAccessControl {
...
+  async checkUserIsOwner(userId: string, projectId: string): Promise<void> {
+    const userIsProjectOwner = await this.roles.findOne({
+      where: {
+        projectId,
+        userId,
+        roleName: Roles.project_owner,
+      },
+    });
+    if (!userIsProjectOwner) {
+      throw new ForbiddenException();
```
kgajowy, line 106: Service layer doesn't know about REST or HTTP - it cannot assume the consumer is using those. 

consider:
`async isOwner(userId: string, projectId: string): Promise<Permit>`

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -92,4 +93,53 @@ export class ProjectAclService implements ProjectAccessControl {
...
+  async findUsersInProject(
+    projectId: string,
+  ): Promise<UsersProjectsApiEntity[]> {
+    const usersInProject = await this.roles.find({
+      where: { projectId },
+      select: ['roleName'],
+      relations: ['user'],
+    });
+    return usersInProject;
+  }
```
kgajowy, line 119: As we are peeking on `user` relation, we are close to https://refactoring.guru/pl/refactoring/smells/couplers smell.
Let's ask ourselves a question: Why would AccessControl need to know user details? It pollutes its boundaries, doesn't it?

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -92,4 +93,53 @@ export class ProjectAclService implements ProjectAccessControl {
...
+    if (!existingUserInProject) {
+      await this.roles.save({
+        projectId,
+        userId,
+        roleName,
+      });
+    } else {
+      await this.roles.update({ projectId, userId }, { roleName });
+    }
+  }
```
kgajowy, line 140: Correct me if I am wrong but it seems we are not checking if the request contains allowed roles: for example, if user is `viewer` / `collaborator`, can he update this rights to `owner`? (relevant spec would help us to discover it)

kgajowy (changes requested):
👍 great job so far!

Blocking due to possible security issue (updating roles) and emerging issue with extending boundaries to `users` details.

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
ikas: good point, we'll need to check (and add tests) for this as well:
- only owners can add other owners;
- we can only remove owners if there's at least one leftover after removing (no orphan projects);

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
ikas: @kgajowy on one hand, I understand why you say that - it would be ideal to keep concepts (modules) as separated as possible. Additionally, there's a possibility in the future that we'll introduce `access-control` to the `users` module, and that might cause circular dependencies in the future.

On the other hand, if we don't return _any_ user information here, then the FE will not be able to build a proper list of project contributors (at least not without making unnecessary requests to find the user information). Also, the purpose of `access-control` is to grant roles on things to _users_ - so one could argue that it needs to know user details...

One idea that might be a good common ground to address both concerns would be do a plain old `left join` with the `users` table, and just pick the user `name` (and other fields the FE might need) in a query builder. This way, we would prevent from having an explicit dependency between the `access-control` module, but we would still return all the data the FE needs to properly render this list - what are your thoughts about this?

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
rubvalave: What I understood in this [comment](https://github.com/Vizzuality/marxan-cloud/pull/640#discussion_r754446378) by @ikas was that it was better to do the checks for the ownership of the projects inside each method on the service rather than the controller, thus avoiding repetition in the controller and having the same effect in the end.

Also, if I move this again to the controller by returning the boolean/Permit, wouldn't that mean that I would need to add mixed return types? Like if the method fails and returns a boolean instead of the actual purpose of the method, would it be a little bit confusing?

ikas: @rubvalave yes, you should do the checks inside the service, but you should not throw a `ForbiddenException` :)

Like @kgajowy mentioned, it doesn't make sense to throw HTTP exceptions (like `ForbiddenException`) from inside the service, since the caller might not be a controller, and it might not be prepared to handle it.

Instead, you can just return `false` here (aka `Denied`, from the `Permit` type), and on the controller, if you receive `false`, then you throw the `ForbiddenException` :+1:

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
kgajowy: We can either: 
* return just the ids and compose the results on `project` module
* make a compromise - i.e. not mess with modules, but make a dedicated read model, utilizing both tables (could be a db view) - such module would use both (indirectly) just for reads

kgajowy: The main aspect is there to not extend module just because the view needs something - we want it to be "clean" (purist alert) and do what we designed it for.

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
rubvalave: I can't believe I overlooked this because what you said makes total sense and would be the first thing I should have realized. I added some extra tests to check that this can't be done in the e2e tests and also a condition to avoid deleting the last owner in the project.

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
rubvalave: I think I understood it better now and I did the appropriate movements regarding this (or so I think) by moving the `ForbiddenException` into the controller after passing the check in the service.

Any comment about this would be much appreciated.

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
rubvalave: Correct me if I'm wrong (that I will probably be) but are we actually extending the module here instead of just accessing the joined table for `users`, `projects`  and `roles` and just adding some information from there?

I mean, the thing I did (now) was use a `createQueryBuilder` method to add only the information needed by FE to do some kind of listing of users (displayName, id, avatarDataUrl), and maybe even the id can be skipped.

kgajowy: I believe it would be a standalone module

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -92,4 +94,117 @@ export class ProjectAclService implements ProjectAccessControl {
...
+        projectId,
+        userId,
+        roleName: Roles.project_owner,
+      },
+    });
+    if (!userIsProjectOwner) {
+      return false;
+    }
+
+    return true;
```
ikas, line 110:
```suggestion
    return !!(await this.roles.findOne({
      where: {
        projectId,
        userId,
        roleName: Roles.project_owner,
      },
    }));
```

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -92,4 +94,117 @@ export class ProjectAclService implements ProjectAccessControl {
...
+        roleName: Roles.project_owner,
+      },
+    });
+    if (
+      allOwnersInProject.length === 1 &&
+      allOwnersInProject[0].userId === userId
+    ) {
+      return false;
+    }
+    return true;
```
ikas, line 126: I can't really explain why, but I'm not the biggest fan of this... We could add the `userId` condition to the where clause and then just straight away `return allOwnersInProject.length <= 1`

But again, this is mostly a preference of syntax, nothing blocking :sweat_smile:

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -92,4 +94,117 @@ export class ProjectAclService implements ProjectAccessControl {
...
+  }
+
+  async findUsersInProject(
+    projectId: string,
+    userId: string,
+  ): Promise<UsersProjectsApiEntity[] | Permit> {
+    const isOwner = await this.checkUserIsOwner(userId, projectId);
+    if (!isOwner) {
+      return isOwner;
+    }
```
ikas, line 136:
```suggestion
    if (!(await this.checkUserIsOwner(userId, projectId))) {
      return false;
    }
```

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -92,4 +94,117 @@ export class ProjectAclService implements ProjectAccessControl {
...
+  async updateUserInProject(
+    projectId: string,
+    updateUserInProjectDto: UserRoleInProjectDto,
+    loggedUserId: string,
+  ): Promise<void | Permit> {
+    const { userId, roleName } = updateUserInProjectDto;
+    const isOwner = await this.checkUserIsOwner(loggedUserId, projectId);
+    if (!isOwner) {
+      return isOwner;
+    }
```
ikas, line 162:
```suggestion
    if (!(await this.checkUserIsOwner(loggedUserId, projectId))) {
      return false;
    }
```

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -92,4 +94,117 @@ export class ProjectAclService implements ProjectAccessControl {
...
+      await apiQueryRunner.release();
+    }
+  }
+
+  async revokeAccess(
+    projectId: string,
+    userId: string,
+    loggedUserId: string,
+  ): Promise<void | Permit> {
+    const isOwner = await this.checkUserIsOwner(loggedUserId, projectId);
```
ikas, line 199: same suggestion as above here

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -92,4 +94,117 @@ export class ProjectAclService implements ProjectAccessControl {
...
+    if (!isOwner) {
+      return isOwner;
+    }
+
+    const isLastOwner = await this.checkLastOwner(userId, projectId);
```
ikas, line 204: same suggestion as above here

ikas (approved):
Left a few comments, mostly syntax suggestions, but I think overall looks good :rocket: 

Let's wait for Andrea and Kamil's review before pulling the trigger!

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
ikas: looks good, great job :+1:

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
ikas: @rubvalave please add a `/** @debt */` comment to the docs of this method, explaining what the problem is, how we plan to tackle it, and let's refactor this later down the road once we revisit this module after the demo. 

This will resurface once we tackle organization/scenario/solution permissions, so let's merge this PR for now and tackle this issue after the demo :+1:

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
rubvalave: If we add the `userId` condition to the allOwners it would only always find one as long as it exists (thus making impossible to remove other owners from the project).

If I do the same but I don't add the `userId` to the where clause you would have the possibility of removing yourself even if you are the last one in the project.

What I would do (and I'm going to commit is changing what is returned into `return allOwnersInProject.length > 1` so it would only allow to delete owners as long as there is more than 1, and it would allow to remove yourself as owner (that I doubt if it should be possible, but this is for another time)

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
kgajowy: I would actually recommend original code style. A few more lines but much easier to read , because:
1) we have named condition and we know the intentions behind
2) double bang operator is always hard to understand

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
kgajowy: @rubvalave @ikas is there a reason why we cannot return count ?

(pseudocode)
``` 
select count(....) where projectId = ... and role = ... and userId != ....
 ```

File: api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts (hunk 25 above)
Comment (kgajowy) [reply api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts]: actually if we are at minor styling:

`is....` usually indicates a function (`isBoolean`)

`const owner = await this.isOwner(...)` (i.e. `checkUser` is also redundant)

as of https://github.com/Vizzuality/marxan-cloud/pull/640#discussion_r757312393 and other suggestions, I would actually recommend leaving direct assignment.

File: api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts [hunk 27]
```diff
@@ -92,4 +94,109 @@ export class ProjectAclService implements ProjectAccessControl {
...
+  /**
+   * @debt This module should not involve user details and it should deal with it
+   * using a standalone module that will access the data just to read it. We
+   * have to get back to it once scenarios, organizations and solutions are included
+   * inside the access-module.
+   */
+  async findUsersInProject(
+    projectId: string,
+    userId: string,
+  ): Promise<UsersProjectsApiEntity[] | Permit> {
```
kgajowy, line 127: Consider returning `Either<SomeErrorSymbols | ProjectUsers[]`

Also: we shouldn't leak entities to controller to avoid unwanted interactions (why would controller need to know about entity and make actions on it?)

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -92,4 +94,109 @@ export class ProjectAclService implements ProjectAccessControl {
       this.canViewSolutionRoles,
     );
   }
+
+  async checkUserIsOwner(userId: string, projectId: string): Promise<Permit> {
```
kgajowy, line 98: `isOwner`, the rest seems redundant, given we are in ACL scope.

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -92,4 +94,109 @@ export class ProjectAclService implements ProjectAccessControl {
...
+    return !!(await this.roles.findOne({
+      where: {
+        projectId,
+        userId,
+        roleName: Roles.project_owner,
+      },
+    }));
+  }
+
+  async checkLastOwner(userId: string, projectId: string): Promise<Permit> {
```
kgajowy, line 108: I don't fully understand what intends to do, name isn't self explanatory.
Why would be need to find `last owner`?

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -92,4 +94,109 @@ export class ProjectAclService implements ProjectAccessControl {
...
+    const { userId, roleName } = updateUserInProjectDto;
+    if (!(await this.checkUserIsOwner(loggedUserId, projectId))) {
+      return false;
+    }
+
+    const apiDbConnection = getConnection(DbConnections.default);
+    const apiQueryRunner = apiDbConnection.createQueryRunner();
+
+    await apiQueryRunner.connect();
+    await apiQueryRunner.startTransaction();
```
kgajowy, line 161: Consider using `EntityManager` (`manager.transaction( transaction => .... )` ) to get rid of boilerplate code, including manually getting the connection.

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
kgajowy: ah, judging from code I get it now.

How about: `hasAnotherOwner` ?

api/apps/api/test/access-control/project-acl-get.e2e-spec.ts
```diff
@@ -0,0 +1,132 @@
...
+  const response = await fixtures.WhenGettingProjectUsersAsViewer(projectId);
+  fixtures.ThenForbiddenIsReturned(response);
+});
+
+test(`getting project users with user in project but is contributor`, async () => {
+  const projectId = await fixtures.GivenProjectWasCreated();
+  await fixtures.WhenAddingANewContributorToTheProjectAsOwner(projectId);
+  const response = await fixtures.WhenGettingProjectUsersAsContributor(
+    projectId,
+  );
```
kgajowy, line 39: Usually there should be only one "action", the other things should be either conditions (Given) or checks (Then). As in this scope we check the ACLs, `WhenGettingProjectUsersAsContributor` could be merged with `Then` (and be like check of particular permission).

Also the description isn't clear to me. Did you mean, `getting project users as contributor` ? (also even if so, test seems to be wrong 🤔 )

api/apps/api/test/access-control/project-acl-get.e2e-spec.ts
```diff
@@ -0,0 +1,132 @@
...
+  fixtures.ThenForbiddenIsReturned(response);
+});
+
+test(`getting project users with user in project but is other owner`, async () => {
+  const projectId = await fixtures.GivenProjectWasCreated();
+  await fixtures.WhenAddingANewOwnerToTheProjectAsOwner(projectId);
+  const response = await fixtures.WhenGettingProjectUsersAsOtherOwner(
+    projectId,
+  );
+  fixtures.ThenAllUsersinProjectAfterAddingAnOwnerAreReturned(response);
```
kgajowy, line 49: I suppose we can skip complicated descriptions and keep `ThenProjectUsersAreAvailable` , as the rest of information is represented by steps.

api/apps/api/test/access-control/project-acl-get.e2e-spec.ts
```diff
@@ -0,0 +1,132 @@
...
+  const response = await fixtures.WhenGettingProjectUsersAsOwner(projectId);
+  fixtures.ThenSingleOwnerUserInProjectIsReturned(response);
+});
+
+test(`delete last owner as the only owner`, async () => {
+  const projectId = await fixtures.GivenProjectWasCreated();
+  const revokeResponse = await fixtures.WhenRevokingAccessToLastOwnerFromProjectAsOwner(
+    projectId,
+  );
+  fixtures.ThenForbiddenIsReturned(revokeResponse);
```
kgajowy, line 131: [minor] Is it forbidden operation due to ACL? (would consider BadRequest)

api/apps/api/test/access-control/project-acl-get.e2e-spec.ts
```diff
@@ -0,0 +1,132 @@
...
+    projectId,
+  );
+  fixtures.ThenNoContentIsReturned(revokeResponse);
+  revokeResponse = await fixtures.WhenRevokingAccessToOwnerFromProjectAsOwner(
+    projectId,
+  );
+  fixtures.ThenNoContentIsReturned(revokeResponse);
+  const response = await fixtures.WhenGettingProjectUsersAsOwner(projectId);
+  fixtures.ThenSingleOwnerUserInProjectIsReturned(response);
+});
```
kgajowy, line 124:
```
fixtures.GivenProjectWasCreated();
const viewerId = fixtures.GivenUserInProject(projectId, `viewer`);
fixtures.GivenUserInProject(projectId, `contributor`);
fixtures.GivenUserInProject(projectId, `owner`);

fixtures.WhenRevokingAccess(projectId, viewerId /* or other...*/ );

fixtures.Then.... /** (get & verify under the hood) **/
```


also would recommend to split those tests (if there are too much within single file, split into multiple files - there, `project-acl-revoke`)

api/apps/api/test/access-control/project-acl.fixtures.ts
```diff
@@ -0,0 +1,242 @@
...
+  const userProjectsRepo: Repository<UsersProjectsApiEntity> = app.get(
+    getRepositoryToken(UsersProjectsApiEntity),
+  );
+  const cleanups: (() => Promise<void>)[] = [];
+
+  return {
+    cleanup: async () => {
+      await userProjectsRepo.delete({
+        userId: ownerUserId,
+      });
```
kgajowy, line 33: `userProjectsRepo.delete({ userId: In([....])})`

api/apps/api/test/access-control/project-acl.fixtures.ts
```diff
@@ -0,0 +1,242 @@
+import * as request from 'supertest';
```
kgajowy, line 1: Seems that fixtures/tests are trying to test too many things within one place

api/apps/api/src/modules/access-control/projects-acl/project-acl.controller.ts
```diff
@@ -0,0 +1,96 @@
...
+@Controller(`${apiGlobalPrefixes.v1}/roles/projects`)
+export class ProjectAclController {
+  constructor(private readonly projectAclService: ProjectAclService) {}
+
+  @Get(':projectId/users')
+  @ApiOperation({ summary: 'Get all users with roles in project' })
+  @ApiOkResponse({
+    status: 200,
+    description: 'User roles found',
+    type: [UserRoleInProjectDto],
```
kgajowy, line 39: one-element array of `UserRoleInProjectDto` type?

api/apps/api/test/access-control/project-acl-get.e2e-spec.ts
rubvalave: The intention with all this tests was to be able to check that no other users were able to access the users of a given project except those with `project_owner` role.

I think this convoluted way of doing the tests (not only this one but the others as well) was because I didn't find an easy way of creating new users from the beginning, and for some of the tests I had some problems if I tried to add them in the `beforeEach()` statement.

I just noted down the general idea on how to improve them (splitting them up into revoke-get-add, less actions in the middle) and  added it to a TO-DO-FUTURE-PR note.

For the time being, and as we discussed earlier, I think it would be better if I try to finish all the tasks related to project-users access-control to use it as kind of chaotic approach to all this (as long as there is not a big security issue, or something directly wrong)

api/apps/api/test/access-control/project-acl-get.e2e-spec.ts
rubvalave: I was wondering the same thing and I almost used BadRequest instead of Forbidden, but when I looked up the definition of 403 i found this (emphasis mine):

The request contained valid data and was understood by the server, but the server is refusing action. This may be due to the user not having the necessary permissions for a resource or needing an account of some sort, or **attempting a prohibited action** (e.g. creating a duplicate record where only one is allowed). This code is also typically used if the request provided authentication by answering the WWW-Authenticate header field challenge, but the server did not accept that authentication. The request should not be repeated.

As it was trying to do a prohibited action (removing the last owner from the project) I thought 403 was enough.

api/apps/api/test/access-control/project-acl-get.e2e-spec.ts
rubvalave: Same thing as https://github.com/Vizzuality/marxan-cloud/pull/640#discussion_r757509916

api/apps/api/test/access-control/project-acl.fixtures.ts
rubvalave: Same as https://github.com/Vizzuality/marxan-cloud/pull/640#discussion_r757509916

api/apps/api/src/modules/access-control/projects-acl/project-acl.controller.ts
rubvalave: I did not know exactly how to indicate to Swagger that this was an array and I wrongly assumed this was the correct way. Thanks for pointing it out, I think it is solved now.

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -92,4 +94,113 @@ export class ProjectAclService implements ProjectAccessControl {
...
+   * have to get back to it once scenarios, organizations and solutions are included
+   * inside the access-module.
+   */
+  async findUsersInProject(
+    projectId: string,
+    userId: string,
+  ): Promise<UserRoleInProjectDto[] | Permit> {
+    if (!(await this.isOwner(userId, projectId))) {
+      return false;
+    }
```
hotzevzl, line 134: I am not a big fan of returning data or an error condition. I would not change it now, because it works and it's clear enough, but I suggest to eventually either switch to using `Either` from `fp-ts` as we do in several other places, or to use exceptions. Either may be preferable.

api/apps/api/test/access-control/project-acl-get.e2e-spec.ts
```diff
@@ -0,0 +1,132 @@
...
+let fixtures: FixtureType<typeof getFixtures>;
+
+beforeEach(async () => {
+  fixtures = await getFixtures();
+});
+afterEach(async () => {
+  await fixtures?.cleanup();
+});
+
+test(`getting project users when user is not added to project`, async () => {
```
hotzevzl, line 13: very minor, but the "added" bit feels off
```suggestion
test(`getting project users when user has no role in the project`, async () => {
```

api/apps/api/test/access-control/project-acl-get.e2e-spec.ts
```diff
@@ -0,0 +1,132 @@
...
+  const projectId = await fixtures.GivenProjectWasCreated();
+  const response = await fixtures.WhenGettingProjectUserAsNotInProject(
```
hotzevzl, line 15:
```suggestion
  const response = await fixtures.WhenGettingProjectUsersAsNotInProject(
```

api/apps/api/test/access-control/project-acl-get.e2e-spec.ts
```diff
@@ -0,0 +1,132 @@
...
+
+test(`getting project users as the owner and unique user`, async () => {
+  const projectId = await fixtures.GivenProjectWasCreated();
+  const response = await fixtures.WhenGettingProjectUsersAsOwner(projectId);
+  fixtures.ThenSingleOwnerUserInProjectIsReturned(response);
+});
+
+test(`getting project users with user in project but is viewer`, async () => {
+  const projectId = await fixtures.GivenProjectWasCreated();
+  await fixtures.WhenAddingANewViewerToTheProjectAsOwner(projectId);
```
hotzevzl, line 29: as Kamil pointed out below, this should not be an action but another condition - a name change alone may make the setup clearer:

```suggestion
  await fixtures.GivenAUserWasGivenViewerRoleOnProject(projectId);
```
Also, we shouldn't care to encode in the condition description that this was done by an owner, I think - if this were not the case, the role granting operation should be rejected, and we test that already. Incidentally, the actions in the tests for this (only owners can grant roles), you have correctly used `When` and encoded that the fact that the operation is carried out by a user with a given role.

api/apps/api/test/access-control/project-acl.fixtures.ts
hotzevzl: @rubvalave what Kamil suggests can be as simple as in the snippet below (lifted wholesale from `input-file.e2e-spec.ts`). What you have here may work now because there may be no dependencies, but if when adding new tests/setups dependencies are introduced, a LIFO approach should be safer.

```
      for (const cleanup of cleanups.reverse()) {
        await cleanup();
      }
```

api/apps/api/test/access-control/project-acl-get.e2e-spec.ts
hotzevzl: dealing with ACL mechanics in tests is often tricky, and often things such as users need to be treated differently than in tests for other parts of the domain - what you have done here doesn't feel necessarily convoluted to me; nevertheless the test does seem off to me too: I am concerned that with most things user-related going on "behind the scenes", we can accidentally mess things up.

Partly matter of preference, but even just returning the new contributor user here and then passing it in to the then-check (which should then be named differently) would help. Or skipping the second condition (creation of a user) and encoding that in the then-check.

Pragmatically, I am also happy to keep these tests as they are now, since it's a tricky part and trying to get absolutely everything "right" would be too much for a single PR, in my opinion. But in order for us to keep confidence in these access-related tests over time, I think some refactor would be advisable in a future PR.

hotzevzl: actually, having looked now at the fixtures file, I would go directly for the second approach I suggested above (which should be the same suggested by Kamil earlier), so
given project was created
when getting users as contributor
then...

the fact that a contributor user was created at some point doesn't have much to do with what we are testing in this specific test, I think

hotzevzl: and so on for the rest of the tests and related fixtures - I'm not going to comment further given that they stick to similar patterns throughout.

hotzevzl (commented):
thanks @rubvalave! I think this PR could be improved on some fronts and I tried to add some suggestions besides what Kamil has done, but I'm also happy for us to merge this (bar any typos/quick fixes) for the sake of pragmatically adding the functionality and then leaving some time to refactor what we may agree on.

api/apps/api/src/modules/access-control/projects-acl/project-acl.controller.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  @Delete(':projectId/users/:userId')
+  @HttpCode(204)
+  @ApiOperation({ summary: 'Revoke access to user from project' })
+  @ApiResponse({ status: 204, description: 'User was deleted correctly' })
+  async deleteUserFromProject(
+    @Param('projectId', ParseUUIDPipe) projectId: string,
+    @Param('userId', ParseUUIDPipe) userId: string,
+    @Req() req: RequestWithAuthenticatedUser,
+  ): Promise<void | boolean> {
```
kgajowy, line 86: `void`?

rubvalave:
Merging and taking smaller PR's for the comments in this one

PR #641, opened 2021-11-19
Feat marxan 996 calculate default blm

AaronPerezPerez:
### Overview

After the creation of the project, now we are calculating and persisting the default BLM values

### Testing instructions

*WIP*

### Feature relevant tickets

[MARXAN-996
](https://vizzuality.atlassian.net/browse/MARXAN-996)
---

- [ ] Update CHANGELOG file


api/apps/api/src/modules/blm/values/blm-values.module.ts
```diff
@@ -1,20 +1,15 @@
...
 
-import { ProjectBlmRepository } from './project-blm-repository';
-import { TypeormProjectBlmRepository } from './typeorm-project-blm-repository';
+import { ProjectBlmRepository } from '@marxan-api/modules/blm';
+import { TypeormProjectBlmRepository } from './repositories/typeorm-project-blm-repository';
+import { TypeOrmModule } from '@nestjs/typeorm';
+import { ProjectBlm } from '@marxan-api/modules/blm/values/repositories/project-blm.api.entity';
+import { MemoryProjectBlmRepository } from '@marxan-api/modules/blm/values/repositories/memory-project-blm-repository';
+import { ProjectBlmRepositoryToken } from '@marxan-api/modules/blm/values/repositories/project-blm-repository';
+import { ProjectBlmRepositoryModule } from '@marxan-api/modules/blm/values/repositories/project-blm-repository.module';
```
kgajowy, line 9: minor: would be good to keep things that are related to this very module as relative imports - just for a sake of reading brevity

api/apps/api/src/modules/blm/values/blm-values.module.ts
```diff
@@ -1,20 +1,15 @@
 import { Module } from '@nestjs/common';
 
-import { ProjectBlmRepository } from './project-blm-repository';
-import { TypeormProjectBlmRepository } from './typeorm-project-blm-repository';
+import { ProjectBlmRepository } from '@marxan-api/modules/blm';
+import { TypeormProjectBlmRepository } from './repositories/typeorm-project-blm-repository';
+import { TypeOrmModule } from '@nestjs/typeorm';
+import { ProjectBlm } from '@marxan-api/modules/blm/values/repositories/project-blm.api.entity';
+import { MemoryProjectBlmRepository } from '@marxan-api/modules/blm/values/repositories/memory-project-blm-repository';
+import { ProjectBlmRepositoryToken } from '@marxan-api/modules/blm/values/repositories/project-blm-repository';
```
kgajowy, line 8: Seems those are not used

api/apps/api/src/modules/blm/values/repositories/memory-project-blm-repository.ts
```diff
@@ -0,0 +1,51 @@
...
+    return right(true);
+  }
+
+  async update(
+    projectId: string,
+    range: ProjectBlm['range'],
+    values: ProjectBlm['values'],
+  ): Promise<Either<SaveFailure, true>> {
+    return right(true);
+  }
```
kgajowy, line 50: I know this is only "temporary" but just for being consistent, could actually update.

api/apps/api/src/modules/blm/values/repositories/memory-project-blm-repository.ts
```diff
@@ -0,0 +1,51 @@
...
+  async get(projectId: string): Promise<Either<GetFailure, ProjectBlm>> {
+    const blm = this.memory[projectId];
+    if (!blm) return left(projectNotFound);
+
+    return right(blm);
+  }
+
+  async create(
+    projectId: string,
+    defaults: ProjectBlm['defaults'],
```
kgajowy, line 30: Could use the defaults as well actually. Even if kind of "temporary", still would have a great use even for testing.

api/apps/api/src/modules/blm/values/repositories/project-blm-repository.ts
```diff
@@ -14,7 +14,7 @@ export interface ProjectBlm {
   /**
```
kgajowy, line 14: (file location). Why did we move to `repositories` ? Is it giving any particular value? The "port" itself may be used by other applications/use-cases within `blm/values` submodule 🤔

api/apps/api/src/modules/blm/values/repositories/project-blm-repository.ts
```diff
@@ -14,7 +14,7 @@ export interface ProjectBlm {
...
    * User-supplied or defaults if none provided,
    */
-  range: [number, number];
+  range: number[];
```
kgajowy, line 17: Would you mind to elaborate? Range is restricted to be 2 numbers.

api/apps/api/src/modules/blm/values/repositories/project-blm-repository.ts
```diff
@@ -29,6 +29,8 @@ export interface ProjectBlm {
 
 type SaveSuccess = true;
 
+export const ProjectBlmRepositoryToken = 'ProjectBlmRepositoryToken';
```
kgajowy, line 32: Seems to be not needed. `ProjectBlmRepository` as a token should be sufficient.

api/apps/api/src/modules/blm/values/repositories/project-blm.api.entity.ts
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * The project to which this BLM Values belongs.
+   */
+  @ApiProperty()
+  @PrimaryColumn('uuid')
+  @OneToOne(() => Project)
+  id!: string;
+}
+
+export class JSONAPIProjectBlmData {
```
kgajowy, line 41: We don't need/want the CRUD there I believe - there is no particular reason to extend internal values storage implementation with full CRUD capabilities.
Also, mixing DTO and DB entity should rather be avoided (unless we are really wanting to build fast) - exposing entity (including all methods) to module consumers can lead to serious issues; also this starts to pollute the entities with possibly unrelated/unwanted related entities.

api/apps/api/src/modules/blm/values/repositories/typeorm-project-blm-repository.ts
```diff
@@ -24,6 +33,15 @@ export class TypeormProjectBlmRepository extends ProjectBlmRepository {
...
     defaults: ProjectBlm['defaults'],
   ): Promise<Either<CreateFailure, true>> {
+    if (await this.repository.findOne(projectId)) return left(alreadyCreated);
+
+    const projectBlm = await this.repository.create();
+    projectBlm.id = projectId;
+    projectBlm.defaults = defaults;
+    projectBlm.values = [];
+    projectBlm.range = [0, 0];
+    await this.repository.insert(projectBlm);
```
kgajowy, line 43: I am afraid it may not work - AFAIR `id` would be still overridden by TypeORM. (`.query` would work tho)

api/apps/api/src/modules/projects/blm/change-blm-range.handler.ts
```diff
@@ -3,11 +3,16 @@ import { Either, right } from 'fp-ts/Either';
...
+import { Inject } from '@nestjs/common';
+import { ProjectBlmRepositoryToken } from '@marxan-api/modules/blm/values/repositories/project-blm-repository';
 
 @CommandHandler(ChangeBlmRange)
 export class ChangeBlmRangeHandler
   implements IInferredCommandHandler<ChangeBlmRange> {
-  constructor(private readonly blmRepository: ProjectBlmRepository) {}
+  constructor(
+    @Inject(ProjectBlmRepositoryToken)
+    private readonly blmRepository: ProjectBlmRepository,
```
kgajowy, line 14: As above, token shouldn't be necessary:

```
{
 provide: ProjectBlmRepository,
 useClass: ....,
}

//
constructor(private readonly rpo: ProjectBlmRepository) { ... }
```

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
```
kgajowy, line 4: probably leftover

api/apps/api/src/modules/projects/projects-crud.service.ts
```diff
@@ -224,6 +225,9 @@ export class ProjectsCrudService extends AppBaseService<
           projectId: model.id,
           planningAreaGeometryId: createModel.planningAreaId,
         }),
+        this.commandBus.execute(
+          new SetProjectBlm(model.id, createModel.planningUnitAreakm2),
+        ),
```
kgajowy, line 230: Does not seems right, please see:

api/apps/api/src/modules/projects/blm/set-project-blm-handler.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+
+describe('set-project-blm-handler', () => {
+  const projectId = 'fooId';
+  let blmRepository: ProjectBlmRepository;
+  let setProjectBLMHandler: SetProjectBlmHandler;
+
+  beforeEach(() => {
+    blmRepository = new MemoryProjectBlmRepository();
+    setProjectBLMHandler = new SetProjectBlmHandler(blmRepository);
+  });
```
kgajowy, line 15: Would be great to keep using Nest's `Test.` module and provide dependencies.
```
providers: [{
 provide: BlmRepo,
 useClass: MemoryProjectBlmRepository,
}, SetProjectBlmHandler]
```

api/apps/api/src/modules/projects/blm/set-project-blm-handler.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+        defaults: [
+          0.03872983346207417,
+          606.799665767047,
+          1213.5606017006319,
+          1820.3215376342168,
+          2427.0824735678016,
+          3033.8434095013868,
+        ],
+        id: projectId,
+        range: [0, 0],
```
kgajowy, line 32: Shouldn't range be defined (with hardcoded default range)?

api/apps/api/src/modules/blm/values/repositories/project-blm.api.entity.ts
kgajowy: also, extending project/scenario entity (BTW docs/requirements say about scenario, not project, but not sure which is valid tbh, @hotzevzl ?) could leverage `extendGetByIdQuery`-like methods of CRUD abstraction to fetch the values directly.

kgajowy (changes requested):
Nice job overall!

`reject` (or rather, 'hold on') mainly due to:
* Nest DI
* CRUD
* setting project BLM trigger

api/apps/api/src/modules/projects/blm/set-project-blm-handler.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+  const projectId = 'fooId';
+  let blmRepository: ProjectBlmRepo;
+  let setProjectBLMHandler: SetProjectBlmHandler;
+
+  beforeEach(async () => {
+    const module = await Test.createTestingModule({
+      providers: [
+        {
+          provide: ProjectBlmRepo,
+          useFactory: () => new MemoryProjectBlmRepository(),
```
kgajowy, line 18: [minor] Could be just `useClass: MemoryProjectBlmRepository`

api/libs/planning-units-grid/src/planning-unit-set.event.ts
```diff
@@ -9,5 +9,8 @@ import { IEvent } from '@nestjs/cqrs';
  *
  */
 export class PlanningUnitSet implements IEvent {
-  constructor(public readonly projectId: string) {}
+  constructor(
+    public readonly projectId: string,
+    public readonly planningUnitArea: number,
```
kgajowy, line 14: Just to keep event purity, we should refrain from attaching more data than needed - it would require the publisher to know what to emit. I would say the BLM-handler should get the `planningUnitArea` value on its own.
(think of custom-grid in which we may don't even have this value)

api/apps/api/src/modules/projects/blm/set-project-blm-handler.spec.ts
AaronPerezPerez: Not really sure why, but using `useClass` throws an error about dependencies

api/apps/api/src/modules/blm/values/repositories/project-blm-repository.module.ts
```diff
@@ -0,0 +1,17 @@
+import { ProjectBlmRepo } from '@marxan-api/modules/blm/values/project-blm-repo';
+import { Module } from '@nestjs/common';
+import { TypeOrmModule } from '@nestjs/typeorm';
+import { ProjectBlm } from '@marxan-api/modules/blm/values/repositories/project-blm.api.entity';
+import { TypeormProjectBlmRepository } from '@marxan-api/modules/blm/values/repositories/typeorm-project-blm-repository';
```
kgajowy, line 5: [minor] relative imports

api/apps/api/src/modules/projects/blm/project-blm.module.ts
```diff
@@ -6,9 +6,17 @@ import { BlmModule } from '@marxan-api/modules/blm';
...
+import { Project } from '@marxan-api/modules/projects/project.api.entity';
+import { DbConnections } from '@marxan-api/ormconfig.connections';
 
 @Module({
-  imports: [CqrsModule, BlmModule],
+  imports: [
+    CqrsModule,
+    BlmModule,
+    TypeOrmModule.forFeature([Project]),
+    TypeOrmModule.forFeature([], DbConnections.geoprocessingDB),
```
kgajowy, line 18: Shouldn't we load concrete entities?

api/apps/api/src/modules/projects/blm/set-project-blm-handler.spec.ts
```diff
@@ -0,0 +1,49 @@
+import { SetProjectBlmHandler } from '@marxan-api/modules/projects/blm/set-project-blm-handler';
+import { ProjectBlmRepo } from '@marxan-api/modules/blm';
+import { SetProjectBlm } from '@marxan-api/modules/projects/blm/set-project-blm';
+import { right } from 'fp-ts/Either';
+import { Test } from '@nestjs/testing';
+import { MemoryProjectBlmRepository } from '@marxan-api/modules/blm/values/repositories/memory-project-blm-repository';
+
+// TODO: Should we remove this unitary test and make a proper e2e test?
+describe.skip('set-project-blm-handler', () => {
```
kgajowy, line 9: We would extract calculations to standalone dependency and keep unit tests for it, while e2e/integration for whole module.

api/apps/api/src/modules/projects/blm/set-project-blm-handler.spec.ts
```diff
@@ -0,0 +1,49 @@
+import { SetProjectBlmHandler } from '@marxan-api/modules/projects/blm/set-project-blm-handler';
+import { ProjectBlmRepo } from '@marxan-api/modules/blm';
+import { SetProjectBlm } from '@marxan-api/modules/projects/blm/set-project-blm';
+import { right } from 'fp-ts/Either';
+import { Test } from '@nestjs/testing';
+import { MemoryProjectBlmRepository } from '@marxan-api/modules/blm/values/repositories/memory-project-blm-repository';
```
kgajowy, line 6: [minor] relative import

api/apps/api/src/modules/projects/blm/set-project-blm-handler.ts
```diff
@@ -1,17 +1,76 @@
...
+    return (
+      project?.planningUnitAreakm2 ??
+      (await this.entityManager
+        .query(
+          `SELECT AVG(size) from "planning_units_geom" WHERE "project_id" = $1`,
+          [project.id],
+        )
+        .then((res) => res[0].avg))
+    );
   }
```
kgajowy, line 75: @hotzevzl @aagm would you mind to double check if this is right? Seems legit to me.

api/apps/api/src/modules/projects/blm/project-blm.module.ts
AaronPerezPerez: Here we just need the connection to do a raw SQL query

api/apps/api/src/modules/projects/blm/set-project-blm-handler.ts
hotzevzl: @AaronPerezPerez @kgajowy this won't work - but almost 😅 

```suggestion
  private async getPlanningUnitArea(
    project: Project,
  ): Promise<number | undefined> {
    return (
      project?.planningUnitAreakm2 ??
      (await this.entityManager
        .query(
          `SELECT AVG(area)/1e6 from "planning_units_geom" WHERE "project_id" = $1`,
          [project.id],
        )
        .then((res) => res[0].avg))
    );
  }
```

if we want to avoid looking at fractions of square km, `SELECT ROUND(AVG(area))...`

hotzevzl (commented):
@AaronPerezPerez with the exception of the small change required for calculation of average PU area, this looks good to me after you've taken into account Kamil's suggestions - so this is basically an approve, once the average area bit is updated. Thanks!

api/apps/api/src/modules/projects/blm/set-project-blm-handler.spec.ts
hotzevzl: @AaronPerezPerez I agree with Kamil's suggestion here (for a later PR), however, this test looks useful already, in the meanwhile - can we un-skip it (and maybe give it a clearer description)?

api/apps/api/src/modules/projects/blm/set-project-blm-handler.ts
AaronPerezPerez: Added!

api/apps/api/src/modules/projects/blm/set-project-blm-handler.spec.ts
AaronPerezPerez: In the following PRs, I refactorized this and added a test to the domain service in charge of the calculations, and also an e2e test for the related endpoint. Should we still keep this?

api/apps/api/src/modules/projects/blm/set-project-blm-handler.ts
kgajowy: @AaronPerezPerez can't see the change being effectively committed 🤔

AaronPerezPerez: @kgajowy  I'm making all the changes in the last PR, as it has the previous requirements

kgajowy (approved):

PR #643, opened 2021-11-22
feat(project): once pu are set, trigger relevant event

kgajowy:

### DIFF
--- a/api/apps/api/src/modules/planning-units/planning-units.service.ts
+++ b/api/apps/api/src/modules/planning-units/planning-units.service.ts
@@ -1,4 +1,4 @@
-import { Inject, Injectable } from '@nestjs/common';
+import { Inject, Injectable, Logger } from '@nestjs/common';
 
 import { Queue } from 'bullmq';
 import { API_EVENT_KINDS } from '@marxan/api-events';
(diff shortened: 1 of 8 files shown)


api/apps/api/src/modules/planning-units/planning-units.service.ts
```diff
@@ -1,79 +1,32 @@
...
-    const { projectId } = job.data;
-    await this.events.createIfNotExists({
-      kind: API_EVENT_KINDS.project__planningUnits__finished__v1__alpha,
-      topic: projectId,
-      externalId: eventId,
-    });
-  }
+    // bad typing - may happen that job wasn't added
+    if (!job) {
+      return;
```
ikas, line 24: I'd suggest adding some logging here - if adding jobs starts to fail for some reason, we'd want to easily be able to track where that happens

ikas (approved):
Minor comment, overall LGTM :+1:

api/apps/api/src/modules/planning-units/planning-units.service.ts
kgajowy: @ikas Added in https://github.com/Vizzuality/marxan-cloud/pull/643/commits/bedc454297d448da6ba67299a25837e7b9cd1058

ikas: nice :ok_hand:

PR #645, opened 2021-11-22
feat: limit uploaded files size

kgajowy:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/decorators/file-interceptors.decorator.ts
+++ b/api/apps/api/src/decorators/file-interceptors.decorator.ts
@@ -1,0 +1,35 @@
+import { applyDecorators, UseInterceptors } from '@nestjs/common';
+import { FileInterceptor } from '@nestjs/platform-express';
+import { uploadOptions } from '@marxan-api/utils/file-uploads.utils';
+import {
+  complexGeometry,
+  complexGeometryWithProperties,
+  simpleGeometry,
+} from '@marxan-api/modules/uploads';
+import { MulterOptions } from '@nestjs/platform-express/multer/interfaces/multer-options.interface';
+
+export enum GeometryKind {
+  Simple = 'simple',
+  Complex = 'complex',
+  ComplexWithProperties = 'complex-with-properties',
+}
+
+const kindToSizeLimit: Record<GeometryKind, () => MulterOptions['limits']> = {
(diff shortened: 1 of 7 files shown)


api/apps/api/src/decorators/file-interceptors.decorator.ts
```diff
@@ -0,0 +1,35 @@
...
+  return applyDecorators(
+    ...[
+      UseInterceptors(
+        FileInterceptor(fileName, {
+          ...uploadOptions,
+          limits,
+        }),
+      ),
+    ],
+  );
```
kgajowy, line 34: could have standalone decorators per type or different approach - but no point in building exaggerated abstraction

api/apps/api/src/modules/uploads/upload-limits.ts
```diff
@@ -0,0 +1,23 @@
+import { MulterOptions } from '@nestjs/platform-express/multer/interfaces/multer-options.interface';
+import { AppConfig } from '@marxan-api/utils/config.utils';
+
+export const simpleGeometry = (): MulterOptions['limits'] => ({
+  fileSize: (() =>
+    AppConfig.get<number>('fileUploads.limits.singleGeometry', 1048576))(),
```
ikas, line 6: I'd suggest setting default values in `default.json`, instead of declaring them here (same for the other 2 values below)

ikas (commented):
Minor comment, otherwise looks great :ok_hand:

api/apps/api/src/modules/uploads/upload-limits.ts
kgajowy: @ikas the reason I removed those from `json` is pretty simple - due to nature of `AppConfig`, we would anyway need to define the `default` value when `.get`ting value (or it would crash the whole application) - repeating the same value  makes it harder to maintain.

As long as we have a single place to keep the default - together with keeping in mind the issue above, as well as the fact that .json is anyway kept in repository - decision to keep those in code seems to be right. (not to mention its easier to find out defaults navigating in code).

Does this sound fair?

ikas: yep, that's fine :+1:

hotzevzl (approved):

PR #646, opened 2021-11-22
User account confirmation after signing up

ikas:
### Overview

This PR adds the user account confirmation flow, as described [in this document](https://docs.google.com/document/d/1kEEgckbv6K7MoCpUSClJMUl3AV9hNUsjPN52CZte9DY/edit?usp=sharing).

### Testing instructions

After signing-up, an email should be sent to the new user with instructions on how to confirm their account:
1. If the user does not click on the confirm link in the email, they should not be able to log in to the platform;
2. If the user clicks on the confirm link in the email, they should be taken to the front end of the platform, where their account will be activated and they will be able to log in.

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/authentication/password-recovery/mailer.ts
+++ b/api/apps/api/src/modules/authentication/password-recovery/mailer.ts
@@ -10,6 +10,12 @@ import { CreateTransmission, Recipient } from 'sparkpost';
 import { AppConfig } from '@marxan-api/utils/config.utils';
 import { UsersService } from '@marxan-api/modules/users/users.service';
 
+enum SparkpostTemplate {
+  PasswordChangedConfirmation = 'confirmation-password-changed',
+  PasswordRecovery = 'marxan-reset-password',
+  SignUpConfirmation = 'confirmation-account',
+}
+
 export abstract class Mailer {
   abstract sendRecoveryEmail(userId: string, token: string): Promise<void>;
   abstract sendPasswordChangedConfirmation(userId: string): Promise<void>;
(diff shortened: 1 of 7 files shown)


rubvalave (approved):
It looks good to me! 👍 🥇

api/apps/api/src/modules/authentication/password-recovery/mailer.ts
```diff
@@ -25,52 +34,68 @@ const passwordResetPrefixToken = Symbol('password reset prefix token');
...
+    @Inject(forwardRef(() => UsersService))
     private readonly usersService: UsersService,
     @Inject(passwordResetPrefixToken)
     private readonly passwordResetPrefix: string,
   ) {}
-  async sendPasswordChangedConfirmation(userId: string): Promise<void> {
-    const user = await this.usersService.getById(userId);
+
+  private async sendEmail(
+    template: string,
```
hotzevzl, line 60: this suggestion borders on (or may be firmly within the domain of) overengineering, but - given that we (and any admin of a self-hosted instance) are bound to manually sync template names between application and sparkpost, could it be prudent to narrow the typing of this parameter to enum values so that we can have one single point where we list all the templates (and can change them if ~needed~ wanted - the template names we have already cause my ocd to surface given their lack of consistency, even if this is practically irrelevant)?

hotzevzl (approved):
thanks!

I love the `sendEmail()` refactor 🚀 

Consider if you want to do something about my only inline suggestion, otherwise feel free to merge.

api/apps/api/src/modules/authentication/password-recovery/mailer.ts
ikas: I think that makes perfect sense, thanks for the suggestion Andrea :bow: I wish we had done this in RW API for instance, it was a pain in the *** to find a list of all the templates being used across the API, so for sure this will be useful in the future :+1: 

I'll make the change :rocket:

PR #649, opened 2021-11-24
Feat marxan 999 calculate blm values from range

AaronPerezPerez:
### Overview

When the command is handled, it calculates the BLM values from a given range using a new domain service

### Testing instructions

*WIP*

### Feature relevant tickets

[MARXAN-999
](https://vizzuality.atlassian.net/browse/MARXAN-999)
---

- [ ] Update CHANGELOG file


api/apps/api/src/modules/projects/blm/change-blm-range.command.ts
```diff
@@ -3,8 +3,14 @@ import { Either } from 'fp-ts/Either';
 
 export const invalidRange = Symbol(`invalid range`);
 export const unknownError = Symbol(`unknown error`);
+export const queryFailure = Symbol(
+  `could not query planning unit area for project`,
+);
```
kgajowy, line 8: How would consumer of this command understand it? (imagine like you as a developer, are using this command via sdk). Does that mean that planning unit could not be found? (then it isn't a query failure, right?)

api/apps/api/src/modules/projects/blm/change-blm-range.handler.ts
```diff
@@ -1,18 +1,82 @@
...
+  queryFailure,
+} from './change-blm-range.command';
 import { ProjectBlmRepo } from '@marxan-api/modules/blm';
+import { InjectEntityManager, InjectRepository } from '@nestjs/typeorm';
+import { Project } from '@marxan-api/modules/projects/project.api.entity';
+import { EntityManager, Repository } from 'typeorm';
+import { DbConnections } from '@marxan-api/ormconfig.connections';
+import { Logger } from '@nestjs/common';
+import { SetProjectBlm } from '@marxan-api/modules/projects/blm/set-project-blm';
+import { BlmValuesCalculator } from '@marxan-api/modules/projects/blm/domain/blm-values-calculator';
```
kgajowy, line 16: Please pay attention to imports. As we are using barrel files, it may lead to unintended issues during compilation - also, it is nice to ensure what we import and from where ;)

api/apps/api/src/modules/projects/blm/change-blm-range.handler.ts
```diff
@@ -1,18 +1,82 @@
...
+    if (this.isInValidRange(range)) {
+      this.logger.error(
+        `Received range [${range[0]},${range[1]}] for project with ID: ${projectId} is invalid`,
+      );
+
+      return left(queryFailure);
+    }
+
+    const project = await this.projectRepository.findOneOrFail(projectId);
+    const area = await this.getPlanningUnitArea(project);
```
kgajowy, line 44: seems like a duplicated code in both set & change. Would be great to consider moving related things as a dependency

api/apps/api/src/modules/projects/blm/change-blm-range.handler.ts
```diff
@@ -1,18 +1,82 @@
...
+
+    if (!area) {
+      this.logger.error(
+        `Could not get Planning Unit area for project with ID: ${projectId}`,
+      );
+
+      return left(queryFailure);
+    }
+
+    const blmValues = BlmValuesCalculator.with(range, area);
```
kgajowy, line 54: I would recommend keeping NestJS DI system instead "hardcoded" utilities - otherwise we cannot change this behavior (or mock it)

api/apps/api/src/modules/projects/blm/domain/blm-values-calculator.ts
```diff
@@ -0,0 +1,26 @@
+export class BlmValuesCalculator {
```
kgajowy, line 1: While the algorithm of calculation could indeed belong do domain (although rather technical one), we would rather need to introduce it as an aggregate (gettable, save-able). As it may not be worth the effort atm, I would stick to keeping it within application.

api/apps/api/src/modules/projects/blm/change-blm-range.command.ts
AaronPerezPerez: Yep, much clearer!

api/apps/api/src/modules/projects/blm/change-blm-range.handler.ts
AaronPerezPerez: Yep, fixed!

api/apps/api/src/modules/projects/blm/domain/blm-values-calculator.ts
AaronPerezPerez: If you mean making it an Aggregate Root, as (correct me if I'm wrong) Project could be, I don't think this calculator could be one.
As far as I know, we can also have domain services which encapsulates domain logic (as I thought this was). Also that's why I didn't  make it a NestJS provider, to keep it clean from libraries as I see it as business logic code.
If we make it a provider, we could mock up the results, for example, but I don't really see an scenario where we wouild need that 🤔

api/apps/api/src/modules/projects/blm/change-blm-range.handler.ts
AaronPerezPerez: Initially this was hardcoded as a private method, but to reuse the logic I extracted it to domain service
(Further explained in the comment below)

api/apps/api/src/modules/projects/blm/change-blm-range.handler.ts
AaronPerezPerez: Great idea, done!

api/apps/api/src/modules/projects/blm/change-blm-range.handler.ts
```diff
@@ -1,18 +1,82 @@
...
+      this.logger.error(
+        `Could not update BLM for project with ID: ${projectId}`,
+      );
+
     return right(true);
   }
+
+  private isInValidRange(range: [number, number]) {
+    return range.length !== 2 && range[0] < 0 && range[0] > range[1];
+  }
```
hotzevzl, line 67: Given how I had to ask colleagues to cross-check unrelated things for me this afternoon, I think it's safe to assume that my brain has already left for the weekend, however, from the checks we're making here and the way the method is used above, should this not actually be as below?

```suggestion
  private isInValidRange(range: [number, number]) {
    return range.length !== 2 || range[0] < 0 || range[0] > range[1];
  }
```

I would actually go a bit further, clarify the naming (contentious issue, I know, so this is just a minor suggestion) and avoiding the use of the negative form (invalid vs valid), thus:

```suggestion
  private isRangeValid(range: [number, number]) {
    return range?.length == 2 && range[0] > 0 && range[0] < range[1];
  }
```
This would however require inverting any checks on the return value in consumers of the method - but in general, and this is bordering on personal preference of course, I suggest to check the positive form of something (is this valid?) vs the negative (is this invalid?), as possible uses such as `!isInvalid()` quickly become harder to mentally parse because of double negations, etc.

hotzevzl (approved):
with the caveat (which you have already fixed in downstream PR as discussed) of the logical operators in the validity check method (and if you wish, the small refactor I suggested), this looks good to me, thanks.

I have no strong opinions about the discussions within this PR related to domain logic.

api/apps/api/src/modules/projects/blm/change-blm-range.handler.ts
AaronPerezPerez: Yep, solved, it was old garbage that I didn't realize because the first validation is done in the DTO, but already fixed

api/apps/api/src/modules/projects/blm/domain/blm-values-calculator.ts
kgajowy: @AaronPerezPerez you may be right about aggregate.

Calculator rather sounds as `domain policy` so we can keep it within domain.

However, we should try to avoid `static` thing - as otherwise we cannot "replace" policies in application, as we strictly hardcode usage via concrete class.

api/apps/api/src/modules/projects/blm/change-blm-range.handler.ts
kgajowy, line 16: imports 🙃

api/apps/api/src/modules/projects/blm/change-blm-range.handler.ts
kgajowy: Given the comment below, if we keep it as "dependency", it could be enough:
` constructor( private readonly blmValuesGetter: Type = new BlmValuesCalculator )`

api/apps/api/src/modules/projects/blm/domain/blm-values-calculator.ts
AaronPerezPerez: Okeey, refactoring it!

api/apps/api/src/modules/projects/blm/change-blm-range.handler.ts
AaronPerezPerez: @kgajowy  When passing it via constructor, NestJS tries to inject a provider and cannot resolve the dependencies. Should we move it to a provider even being a domain policy?

kgajowy: @AaronPerezPerez we can make a pretty dummy factory which will construct such policy, what do you think?

AaronPerezPerez: Great idea!

PR #650, opened 2021-11-25
Feat marxan 1000 patch calibration endpoint

AaronPerezPerez:
### Overview

Implements an endpoint to update the BLM range of a project, calculating its BLM values

### Testing instructions

*WIP*

### Feature relevant tickets

[MARXAN-1000
](https://vizzuality.atlassian.net/browse/MARXAN-1000)
---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/projects/dto/project-blm-values-response.dto.ts
+++ b/api/apps/api/src/modules/projects/dto/project-blm-values-response.dto.ts
@@ -1,6 +1,6 @@
 import { ApiProperty } from '@nestjs/swagger';
 
-export class ProjectBlmValuesResponseDTO {
+export class ProjectBlmValuesResponseDto {
   @ApiProperty({
     description: 'An ID of the project that owns this BLM values',
     example: '6fbec34e-04a7-4131-be14-c245f2435a6c',
@@ -29,7 +29,8 @@ export class ProjectBlmValuesResponseDTO {
   defaults!: number[];
 
   @ApiProperty({
-    description: 'Calculates BLM values with given range',
+    description:
+      'Calculated BLM values according to given range and planning unit area',
     example: [
       38.72983346207417,
       316.2936399402723,
--- a/api/apps/api/src/modules/projects/dto/project-blm-values-response.dto.ts
+++ b/api/apps/api/src/modules/projects/dto/project-blm-values-response.dto.ts
@@ -1,0 +1,44 @@
+import { ApiProperty } from '@nestjs/swagger';
+
+export class ProjectBlmValuesResponseDTO {
+  @ApiProperty({
+    description: 'An ID of the project that owns this BLM values',
+    example: '6fbec34e-04a7-4131-be14-c245f2435a6c',
+  })
+  id!: string;
+
+  @ApiProperty({
+    description: 'Range used to calculate values',
+    example: [0.001, 100],
+    isArray: true,
+  })
+  range!: [number, number];
+
+  @ApiProperty({
+    description: 'Default BLM values',
+    example: [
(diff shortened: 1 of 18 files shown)


api/apps/api/src/modules/projects/dto/update-project-blm-range.dto.ts
```diff
@@ -0,0 +1,10 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { IsArray } from 'class-validator';
+import { IsValidRange } from '@marxan-api/decorators/is-valid-range.decorator';
+
+export class UpdateProjectBlmRangeDTO {
+  @ApiProperty({ example: [0.001, 100] })
+  @IsArray()
+  @IsValidRange()
```
kgajowy, line 8: Isn't it already validated in previous PR ?

AaronPerezPerez: Yes, it might be a bit overkill to validate it at controller and application levels. As I implemented the application code first, I didn't have this validation yet, so when implementing this endpoint I thought this would be better. Then, before removing the next validation I thought "well, a double check won't hurt anybody", but giving it a second thought now, if we change the validation details of a range we have two places to do so now.
So, we can remove the application level validation, or we could also extract the validation rules to a utils where both validations check the range.

api/apps/api/src/modules/projects/dto/project-blm-values-response.dto.ts
```diff
@@ -0,0 +1,44 @@
...
+      1820.3215376342168,
+      2427.0824735678016,
+      3033.8434095013868,
+    ],
+    isArray: true,
+  })
+  defaults!: number[];
+
+  @ApiProperty({
+    description: 'Calculates BLM values with given range',
```
hotzevzl, line 32:
```suggestion
    description: 'Calculated BLM values according to given range and planning unit area',
```

api/apps/api/src/modules/projects/dto/update-project-blm-range.dto.ts
hotzevzl: ok, I see this now what we were discussing in DM 😅 

My suggestion would be to remove the application level validation and to *not* extract the validation rules - actually extracting them was my first reaction when I looked at the decorator's logic, but if we don't actually need to validate downstream (=if we don't currently expect these ranges to come in through any other route than controllers using this DTO), then I would not bother. If we do need to account for other data ingress routes at some point, I'd definitely recommend to extract the logic at that point and share it between ingress routes as needed.

hotzevzl (approved):

api/apps/api/src/modules/projects/dto/project-blm-values-response.dto.ts
AaronPerezPerez: Oops, fixed

api/apps/api/src/modules/projects/dto/update-project-blm-range.dto.ts
AaronPerezPerez: Perfect, I deleted the inner validation.

kgajowy: Sorry to reopen.

@hotzevzl @AaronPerezPerez as this is our kind of valid data rule, I think its better to have it "inner", as not every single update request will come through REST API.

AaronPerezPerez: Okeey, restoring it!

PR #653, opened 2021-11-26
Sign-up confirmation - remove need to provide userId

ikas:
### Overview

This PR removes the need to provide the `userId` in the `validate` endpoint on the sign-up confirmation flow. At that point in time, when the `validate` endpoint is used by the FE, the FE does not know the `userId` yet (because the login didn't happen yet).

This PR makes sure we store the `userId` to be validated in the API event data of the validation token, so that we can validate users based on the token only.

### Testing instructions

`POST /auth/validate` providing only `validationToken` should be possible.

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/authentication/authentication.service.ts
+++ b/api/apps/api/src/modules/authentication/authentication.service.ts
@@ -40,6 +40,16 @@ export interface AccessToken {
 }
 
 /**
+ * Access confirmation token for the app
+ */
+export interface ValidationToken {
+  /**
+   * Token that should be used to confirm the user account.
+   */
+  validationToken: string;
+}
+
+/**
  * JWT payload (decoded)
  */
 export interface JwtDataPayload {
@@ -109,7 +119,7 @@ export class AuthenticationService {
    *
    * @todo Allow to set all of a user's data on signup, if needed.
    */
-  async createUser(signupDto: SignUpDto): Promise<Partial<User>> {
+  async createUser(signupDto: SignUpDto): Promise<ValidationToken> {
     const user = new User();
     user.displayName = signupDto.displayName;
     user.passwordHash = await hash(signupDto.password, 10);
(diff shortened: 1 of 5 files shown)


api/apps/api/src/modules/authentication/authentication.service.ts
```diff
@@ -126,16 +136,15 @@ export class AuthenticationService {
     });
     const validationToken = v4();
     await this.apiEventsService.create({
```
kgajowy, line 138: Let's discuss it how we can avoid messing with ApiEvents, as those purpose was most likely something else. We could even grab dedicated module handling account activation if necessary.

hotzevzl (commented):
thanks @ikas!

I am not sure what to advise - I noted in my only comment to this PR (which otherwise looks good, if we wanted to go this route) that I think we may keep API events as they are and only touch the earlier part of the flow (my fault for not having conveyed the original intent accurately, apologies for this).

I have no strong opinions on this.

I don't particularly like the tweaking of the original API event with userid as topic to actually carry the validation token as topic, because we're pulling the semantics of API events a bit, but that's a minor concern.

If we want to go this route (instead of passing both userid and token to sparkpost, and using both), I would maybe create a different `kind` of API event just for that, *or* (preferable still, from my point of view) create a new method to be used instead of `ApiEvents.getLatestEventForTopic()` which looks for the latest (and likely only, unless we have bugs somewhere) event where `data.validationToken` matches the `validationToken` received on the `POST /auth/validate` endpoint. which may make this current PR even simpler and move less pieces around.

api/apps/api/src/modules/authentication/authentication.service.ts
ikas: yeah, that sounds good, I'll do that :+1: thanks Andrea :bow:

PR #654, opened 2021-11-26
feat(marxan-run): blm calibration

kgajowy:
* contains a few changes related to modules / moves of shared things
* leaves space for detail implementations of input.dat values (BLM)


kgajowy:
> looks good, thank you!
> 
> do we need to backport some pending decisions to the design documents? (or maybe remove pending decisions from there altogether or even better mark them as things we considered, and leave things mostly alone, just to keep a reference to design discussions while avoiding confusion for whomever may read those documents and wonder if anything got out of sync at some point)

@hotzevzl I think we are following the HLD now - single-fetcher, construction of N workspaces, executing marxan runs sequentially - I haven't found something that would require HLD adjustment 🤔

AaronPerezPerez (approved):

PR #659, opened 2021-11-30
test: adds fixtures, improvements in project-acl e2e tests

rubvalave:
### Overview

- Split tests into get / patch / delete tests.
- Usage of already created projects to check GET actions without unnecessary steps
- Changes how cleanup is processed for e2e tests

Part of [MARXAN-1076](https://vizzuality.atlassian.net/browse/MARXAN-1076)

### DIFF
--- a/api/apps/api/test/access-control/project-acl-delete.e2e-spec.ts
+++ b/api/apps/api/test/access-control/project-acl-delete.e2e-spec.ts
@@ -1,0 +1,38 @@
+import { FixtureType } from '@marxan/utils/tests/fixture-type';
+import { getFixtures } from './project-acl.fixtures';
+
+let fixtures: FixtureType<typeof getFixtures>;
+
+beforeEach(async () => {
+  fixtures = await getFixtures();
+});
+afterEach(async () => {
+  await fixtures?.userRoleCleanup();
+  await fixtures?.cleanup();
+});
+
+test(`delete every type of user from the project as owner`, async () => {
+  const projectId = await fixtures.GivenProjectExistsAndHasUsers();
+  let revokeResponse = await fixtures.WhenRevokingAccessToViewerFromProjectAsOwner(
+    projectId,
+  );
+  fixtures.ThenNoContentIsReturned(revokeResponse);
+  revokeResponse = await fixtures.WhenRevokingAccessToContributorFromProjectAsOwner(
+    projectId,
+  );
+  fixtures.ThenNoContentIsReturned(revokeResponse);
+  revokeResponse = await fixtures.WhenRevokingAccessToOwnerFromProjectAsOwner(
+    projectId,
+  );
+  fixtures.ThenNoContentIsReturned(revokeResponse);
+  const response = await fixtures.WhenGettingProjectUsersAsOwner(projectId);
+  fixtures.ThenSingleOwnerUserInProjectIsReturned(response);
+});
+
+test(`delete last owner as the only owner`, async () => {
+  const projectId = await fixtures.GivenProjectWasCreated();
+  const revokeResponse = await fixtures.WhenRevokingAccessToLastOwnerFromProjectAsOwner(
+    projectId,
+  );
+  fixtures.ThenForbiddenIsReturned(revokeResponse);
+});
--- a/api/apps/api/test/access-control/project-acl-delete.e2e-spec.ts
+++ b/api/apps/api/test/access-control/project-acl-delete.e2e-spec.ts
@@ -7,26 +7,30 @@ beforeEach(async () => {
   fixtures = await getFixtures();
 });
 afterEach(async () => {
-  await fixtures?.userRoleCleanup();
   await fixtures?.cleanup();
 });
 
 test(`delete every type of user from the project as owner`, async () => {
-  const projectId = await fixtures.GivenProjectExistsAndHasUsers();
-  let revokeResponse = await fixtures.WhenRevokingAccessToViewerFromProjectAsOwner(
+  const projectId = await fixtures.GivenProjectWasCreated();
+  await fixtures.GivenViewerWasAddedToProject(projectId);
+  await fixtures.GivenContributorWasAddedToProject(projectId);
+  await fixtures.GivenOwnerWasAddedToProject(projectId);
+  const viewerRevokeResponse = await fixtures.WhenRevokingAccessToViewerFromProjectAsOwner(
     projectId,
   );
-  fixtures.ThenNoContentIsReturned(revokeResponse);
-  revokeResponse = await fixtures.WhenRevokingAccessToContributorFromProjectAsOwner(
+  const contributorRevokeResponse = await fixtures.WhenRevokingAccessToContributorFromProjectAsOwner(
     projectId,
   );
-  fixtures.ThenNoContentIsReturned(revokeResponse);
-  revokeResponse = await fixtures.WhenRevokingAccessToOwnerFromProjectAsOwner(
+  const ownerRevokeResponse = await fixtures.WhenRevokingAccessToOwnerFromProjectAsOwner(
     projectId,
   );
-  fixtures.ThenNoContentIsReturned(revokeResponse);
-  const response = await fixtures.WhenGettingProjectUsersAsOwner(projectId);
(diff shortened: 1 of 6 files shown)


api/apps/api/test/access-control/project-acl-delete.e2e-spec.ts
```diff
@@ -0,0 +1,38 @@
+import { FixtureType } from '@marxan/utils/tests/fixture-type';
+import { getFixtures } from './project-acl.fixtures';
+
+let fixtures: FixtureType<typeof getFixtures>;
+
+beforeEach(async () => {
+  fixtures = await getFixtures();
+});
+afterEach(async () => {
+  await fixtures?.userRoleCleanup();
```
kgajowy, line 10: Do we have any particular reason why `userRoleCleanup` cannot be a part of `cleanup`? 

The idea of `cleanup` is to remove all "noise" from the test itself

api/apps/api/test/access-control/project-acl-delete.e2e-spec.ts
```diff
@@ -0,0 +1,38 @@
...
+  revokeResponse = await fixtures.WhenRevokingAccessToContributorFromProjectAsOwner(
+    projectId,
+  );
+  fixtures.ThenNoContentIsReturned(revokeResponse);
+  revokeResponse = await fixtures.WhenRevokingAccessToOwnerFromProjectAsOwner(
+    projectId,
+  );
+  fixtures.ThenNoContentIsReturned(revokeResponse);
+  const response = await fixtures.WhenGettingProjectUsersAsOwner(projectId);
+  fixtures.ThenSingleOwnerUserInProjectIsReturned(response);
```
kgajowy, line 29: in general GWT syntax approach, it is advised to keep G -> W -> T order - this is, without mixing the action (When) with expectation (Then).

We could either group `When`s together or split the actions (as single action would be rather expected). In this case I think it is fine to just group them.

api/apps/api/test/access-control/project-acl.fixtures.ts
```diff
@@ -21,37 +22,95 @@ export const getFixtures = async () => {
...
-      });
-      await Promise.all(cleanups.map((clean) => clean()));
+      for (const cleanup of cleanups.reverse()) {
+        await cleanup();
+      }
       await app.close();
     },
 
+    userRoleCleanup: async () => {
+      if (projectCreatedId) {
```
kgajowy, line 40: [minor, style] consider 
```
if (!projectCreatedId) return
```
to have less indentation

api/apps/api/test/access-control/project-acl.fixtures.ts
```diff
@@ -21,37 +22,95 @@ export const getFixtures = async () => {
...
+      return result.projectId;
+    },
+
     GivenProjectWasCreated: async () => {
       const { cleanup, projectId } = await GivenProjectExists(
         app,
         ownerUserToken,
       );
       cleanups.push(cleanup);
+      projectCreatedId = projectId;
```
kgajowy, line 92: Second thought: we can extract it as a `background`  (see https://cucumber.io/docs/gherkin/reference/) and just put project creation after boostrapping the application

api/apps/api/test/access-control/project-acl.fixtures.ts
```diff
@@ -21,37 +22,95 @@ export const getFixtures = async () => {
...
+          roleName: projectOwnerRole,
+        },
+      });
+
+      if (!result) {
+        throw new Error('Project does not exist');
+      }
+
+      return result.projectId;
+    },
```
kgajowy, line 84: Seems like a "background" (see below) - and actually it is not creating/setting up anything  (`Given something happened`)

api/apps/api/test/access-control/project-acl.fixtures.ts
```diff
@@ -21,37 +22,95 @@ export const getFixtures = async () => {
...
-          ownerUserId,
-          contributorUserId,
-          viewerUserId,
-          otherOwnerUserId,
-        ]),
-      });
-      await Promise.all(cleanups.map((clean) => clean()));
+      for (const cleanup of cleanups.reverse()) {
+        await cleanup();
+      }
```
kgajowy, line 35: `await Promise.all(cleanups.reverse().map((clean) => clean()));`

kgajowy (commented):
Nothing critical, only minor stuff. Well done!

api/apps/api/test/access-control/project-acl-delete.e2e-spec.ts
rubvalave: I did it to separate it from the regular cleanup and only apply it when a project is created, so it does not run everytime I apply a cleanup.

Now that I think of, maybe it is better if I include it in cleanup but I remove cleanups from the before/after GET tests that use the basic seeded data. 🤔

kgajowy: you can always `push` to `cleanups` within given step implementation, how about this?

api/apps/api/test/access-control/project-acl.fixtures.ts
rubvalave: I don't know if I get it completely but, I have two different things `GivenProjectWasCreated` and `GivenProjectExistsAndHasUsers`.

I'm using the first one for the addition of new users and to check who can or can't add any new users. I create this project from scratch because I can't use the ones already in the seeds.

The second one is used for the GET actions, as all the users are already part of the project, as well as for the DELETE of every type of user. This avoids not having to create a new project with every type of user for every test.

However, as I got from the Background link you posted, are those two not `Given` steps that repeat a lot but on different `Scenarios`? First method for when you need a project from scratch, the other already created projects with already established roles.

The only thing I realized is that by using second method for the DELETE of every type of user, I'm effectively not cleaning up/restoring the state of the seeds to its previous state, as the user are deleted from a specific project and not added again.

api/apps/api/test/access-control/project-acl-delete.e2e-spec.ts
rubvalave: You mean like including it in the `GivenProjectWasCreated` for example?

api/apps/api/test/access-control/project-acl-delete.e2e-spec.ts
rubvalave: I think I didn't understand this well. You mean grouping like I did here (all under one test) or grouping every endpoint request in a single `When` method?

Would not make that impossible to check the `ThenNoContentReturned` or also checking those inside the When Method and only using the last `Then` for the latest endpoint request the `When` method returns?

api/apps/api/test/access-control/project-acl-delete.e2e-spec.ts
kgajowy: @rubvalave yes, exactly.

api/apps/api/test/access-control/project-acl-delete.e2e-spec.ts
kgajowy: @rubvalave in this very case I meant:

```
 await fixtures.GivenProjectExistsAndHasUsers();

 await fixtures.WhenRevokingAccessToViewerFromProjectAsOwner(...)
await fixtures.WhenRevokingAccessToOwnerFromProjectAsOwner(...)
//...

await fixtures.Then.....
```

api/apps/api/test/access-control/project-acl.fixtures.ts
kgajowy: as discussed:

```
  const { projectId, cleanup } = await GivenProjectExists(app, jwtToken);
  cleanups.push(cleanup)

//
const projectId = GivenProjectWasCreated()
const userA = GivenUserWasAddedToProject(projectId, Role)
const userB = GivenUserWasAddedToProject(projectId, Role)
const userC = GivenUserWasAddedToProject(projectId, Role)
```

kgajowy (approved):

hotzevzl (approved):
nice one :)

ikas (approved):
LGTM, great job :ok_hand:

PR #664, opened 2021-11-30
Sign-up confirmation base URL configurable through env var

ikas:
### Overview

This PR enables the sign up confirmation URL to be configurable through an environment variable (`SIGNUP_CONFIRMATION_TOKEN_PREFIX`). This variable should take the appropriate value according to the environment where the code is deployed.

### Feature relevant tickets

### DIFF
--- a/api/apps/api/config/default.json
+++ b/api/apps/api/config/default.json
@@ -58,12 +58,15 @@
     "apikey": "invalidSparkpostApikey",
     "origin": "https://api.eu.sparkpost.com:443"
   },
+  "application": {
+    "baseUrl": "http://localhost:3000"
+  },
   "passwordReset": {
-    "tokenPrefix": "http://localhost:3000/auth/reset-password?token=",
+    "tokenPrefix": "/auth/reset-password?token=",
     "expiration": 1800000
   },
   "signUpConfirmation": {
-    "tokenPrefix": "http://localhost:3000/auth/sign-up-confirmation?token="
+    "tokenPrefix": "/auth/sign-up-confirmation?token="
   }
 }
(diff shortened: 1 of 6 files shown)


hotzevzl (approved):
this looks good as is - feel free to merge, or if you wish to switch to a single envvar with string substitution feel free to go ahead and maybe even merge as the changes may be very small in any case, especially if this helps to go ahead with QCing the whole flow on test/demo instances

api/apps/api/config/default.json
```diff
@@ -58,12 +58,15 @@
...
+  },
   "passwordReset": {
-    "tokenPrefix": "http://localhost:3000/auth/reset-password?token=",
+    "tokenPrefix": "auth/reset-password?token=",
     "expiration": 1800000
   },
   "signUpConfirmation": {
-    "tokenPrefix": "http://localhost:3000/auth/sign-up-confirmation?token="
+    "tokenPrefix": "auth/sign-up-confirmation?token="
   }
```
hotzevzl, line 70: nitpicking on this - I would suggest to keep the host part of the URL free from trailing `/` and add it as leading to the path component (or fragments thereof), such as `"tokenPrefix": "/auth/reset-password?token=",` and so on - the current implementation won't cause any trouble, but IIRC we tried to stick to something like this after we had some trouble with external APIs being picky on double-slashes somewhere in the path.

ikas: Fixed on https://github.com/Vizzuality/marxan-cloud/pull/664/commits/6da3d2e4ae5631a234dcabc2d895beec37adc0e6

PR #665, opened 2021-12-01
feat: changes UserApiEntity from project to project-acl

rubvalave:
## Removes userProjects entity under projects module, adds Either monads for project-acl

- Removes UserProjectsApi entity from pure project module and moves under Project-ACL module.
- Changes how service in project-acl return data using monads.
- Adapts units tests for monad changes.


ikas (approved):
:rocket: :medal_sports:

kgajowy (commented):
Relative imports are important - we want to avoid importing from other modules unless this is necessary.
Also, when we would be importing from real module API (from barrel file like index.ts) instead reaching to internal implementations and polluting boundaries, then we could met some cyclic dependency during imports.

Good first step, we still need to remove `UsersProjectsApiEntity` from `projects` module

hotzevzl (approved):

PR #668, opened 2021-12-03
Initial setup for 1007 and projects-acl

rubvalave:
## Initial setup and PR for 1007: Add ACL to projects endpoints.

This PR would only include basic changes made to use the ACL inside the projects-controller.

Future PRs would be apply to this one for each endpoint, so it is easier to follow and review the ACL included.

### DIFF
--- a/api/apps/api/src/modules/access-control/projects-acl/project-acl.service.spec.ts
+++ b/api/apps/api/src/modules/access-control/projects-acl/project-acl.service.spec.ts
@@ -141,10 +141,10 @@ const getFixtures = async () => {
         }),
       ),
     ThenCannotCreateProject: async () => {
-      expect(await sut.canCreateProject(userId, projectId)).toEqual(false);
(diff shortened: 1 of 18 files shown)


api/apps/api/src/modules/access-control/projects-acl/project-acl.service.spec.ts
```diff
@@ -141,10 +141,10 @@ const getFixtures = async () => {
         }),
       ),
     ThenCannotCreateProject: async () => {
-      expect(await sut.canCreateProject(userId, projectId)).toEqual(false);
+      expect(await sut.canCreateProject(userId)).toEqual(false);
```
kgajowy, line 144: ```abstract canCreateProject(userId: string, projectId: string): Promise<Permit>;```
doesn't contract needs to be updated as well? (`redundant projectId`? )

kgajowy (approved):

PR #669, opened 2021-12-06
feat(cloning): export pieces processors

kgajowy:
# Scope

* moved some "shared" things to `lib`
* created queue processor with sample exporter
* leverages NestJS DI system to allow piece-handlers to register by themselves via decorators

Worth to see:
* `api/apps/geoprocessing/src/export/README.md` - how to add new piece support
* `docs/download-upload-clone/exporter-importer-flow.md` - spike & initial decision - of course this PR subject it to select solution by the team, the submitted one is just a proposition.


hotzevzl (approved):
looks great, thanks!

- just informational, as we don't really need to plan for this now - how do you envisage the pieces handlers would work with versioning (of clone output as a whole or of single pieces)?
- async jobs for pieces handling seems wise to me too
- speaking of dependencies, these would be most relevant at import time, but I think there's actually something to handle at export time too, namely the possibility (probably unlikely, but not zero) of changes being applied by users while a project and its pieces are being exported, which may make the inputs to some of the pieces invalid during the cloning process... what's your opinion on how to handle this?

api/libs/cloning/src/job-output.ts
```diff
@@ -0,0 +1,9 @@
+import { ClonePiece } from './domain';
+
+export interface JobOutput {
```
hotzevzl, line 3: do you anticipate `JobOutput` to potentially diverge from `JobInput`? as they stand, the former could extend the latter

kgajowy: @hotzevzl although they are structurally similar, they are serving different purposes - it would be like telling one animal of different kind is extending each other. As they are structurally similar, TS will allows using them as is (with just adding one property and `...`spreading the input.

So, I wouldn't connect them just to save a few lines, as they shall remain independent.

kgajowy:
> * just informational, as we don't really need to plan for this now - how do you envisage the pieces handlers would work with versioning (of clone output as a whole or of single pieces)?

Each Import/Export component could contain version information - I noted versioning possibilities in readme within this PR https://github.com/Vizzuality/marxan-cloud/pull/669/files#diff-9e80c56602770fac273e0c5406dcbe2c8713980b24455b9b8c69073a93fc14c4R11-R17

> changes being applied by users while a project and its pieces are being exported, which may make the inputs to some of the pieces invalid during the cloning process

as stated in [HLD](https://github.com/Vizzuality/marxan-cloud/blob/develop/docs/download-upload-clone/high-level-design.md) : 

```
if there is any job in progress for the project/scenarios, export shouldn't be allowed
import operation is asynchronous
if import of the project (or its underlying scenarios) is ongoing, no action on given project should be possible to take by the user
if import of the scenario is ongoing, no action on given scenario should be possible
```

as such actions are prohibited by design, therefore it is not necessary to "invalidate" them.

api/libs/cloning/src/job-output.ts
hotzevzl: Yes, that was the point of my question, not to save duplication but to check if we should consider them two faces of the same coin. If you consider this not to be the case, then we're ok.

kgajowy: @hotzevzl we could in this very case but it is a small internal change to the module which would need to be changed as soon as those two start to differ 🤷 Could be done if you see any benefit.

PR #670, opened 2021-12-06
feat(cloning): import domain

kgajowy:
# Scope

* import base domain
* request import application

# Scope for another PR / missing parts
* specs for "complete piece"
* adapters module
* connecting adapters with application module
* adding `complete piece` application (transactional repository with lock)


api/apps/api/src/modules/clone/import/domain/import/import.snapshot.ts
```diff
@@ -0,0 +1,14 @@
+import { ArchiveLocation, ResourceKind } from '@marxan/cloning/domain';
+import { ImportComponentSnapshot } from './import.component.snapshot';
+
+export type ImportComponents = Array<
+  ImportComponentSnapshot & { order: number }
+>;
+
+export interface ImportSnapshot {
+  id: string;
+  resourceId: string;
```
AaronPerezPerez, line 10: Should we use ResourceId class?

api/apps/api/src/modules/clone/import/domain/import/import.ts
```diff
@@ -0,0 +1,165 @@
...
+            component.id,
+            component.piece,
+            component.resourceId,
+          ),
+        );
+      }
+    }
+  }
+
+  private getNextComponents(
```
AaronPerezPerez, line 133: Maybe this could be simplified. Something like:
```typescript
  private getNextComponents(
    lastPiece: ImportComponent,
  ): { components: ImportComponent[]; finished: boolean } {
    const elements = this.pieces.filter((pc) => pc.order === lastPiece.order);

    const somePieceDoesNotHaveUri = elements.some((pc) => !Boolean(pc.uri));

    if (somePieceDoesNotHaveUri)
      return {
        components: [],
        finished: false,
      };

    const location = this.pieces.indexOf(lastPiece);
    const nextComponentLocation = location + 1;
    const nextComponent = this.pieces[nextComponentLocation];

    return {
      components: this.pieces.filter((pc) => pc.order === nextComponent?.order),
      finished: !nextComponent,
    };
  }
```
Haven't properly checked this, but something like this with a guard clause seems clearer and should work.

kgajowy: Thanks!

```
const elements = this.pieces.filter((pc) => pc.order === lastPiece.order);
    const hasIncompletePiecesForBatch = elements.some((pc) => !Boolean(pc.uri));

    if (hasIncompletePiecesForBatch) {
      return {
        components: [],
        finished: false,
      };
    }

    const location = this.pieces.indexOf(lastPiece);
    const nextComponentLocation = location + 1;
    const nextComponent = this.pieces[nextComponentLocation];

    if (nextComponent) {
      return {
        components: this.pieces.filter(
          (pc) => pc.order === nextComponent.order,
        ),
        finished: false,
      };
    }

    return {
      components: [],
      finished: true,
    };
  }
```

Ended with the above - tbh I feel that it is bit easier to understand the last piece with `if` clause instead shortening the code - but indeed the early return on not completed batch is good catch!

api/apps/api/src/modules/clone/import/domain/import/import.snapshot.ts
kgajowy: fixed in https://github.com/Vizzuality/marxan-cloud/pull/670/commits/de4030574ab0e538d72f8755b472f4b9dd08dc33 thanks for spotting!

api/apps/api/src/modules/clone/import/domain/import/import.ts
kgajowy: updated in https://github.com/Vizzuality/marxan-cloud/pull/670/commits/5340b5e8fc1aadad9566620d8abd0adc67687f7f

AaronPerezPerez: Nice! 💪🏽

PR #673, opened 2021-12-07
Scenario locks module implementation

ikas:
### Overview

This PR provides an initial implementation for the usage of scenario locks to make sure that only one user at a time can edit a scenario.

### Feature relevant tickets

[MARXAN-1085](https://vizzuality.atlassian.net/browse/MARXAN-1085)
[MARXAN-1086](https://vizzuality.atlassian.net/browse/MARXAN-1086)

### DIFF
--- a/api/apps/api/src/migrations/api/1638891599444-AddScenarioLocksTable.ts
+++ b/api/apps/api/src/migrations/api/1638891599444-AddScenarioLocksTable.ts
@@ -1,0 +1,41 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddScenarioLocksTable1638891599444 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TABLE "scenario_locks" (
+              scenario_id uuid NOT NULL,
+              user_id uuid NOT NULL,
+              grab_date timestamptz NOT NULL
+             );`,
+    );
+
+    await queryRunner.query(
+      `ALTER TABLE "scenario_locks" ADD CONSTRAINT scenario_locks_fk FOREIGN KEY (scenario_id) REFERENCES scenarios(id) ON DELETE CASCADE;`,
+    );
+
+    await queryRunner.query(
+      `ALTER TABLE "scenario_locks" ADD CONSTRAINT scenario_locks_fk_1 FOREIGN KEY (user_id) REFERENCES users(id) ON DELETE CASCADE;`,
+    );
+
+    await queryRunner.query(
+      `ALTER TABLE "scenario_locks" ADD CONSTRAINT scenario_locks_pk PRIMARY KEY (scenario_id, user_id);`,
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "scenario_locks" DROP CONSTRAINT scenario_locks_pk;`,
+    );
+
+    await queryRunner.query(
+      `ALTER TABLE "scenario_locks" DROP CONSTRAINT scenario_locks_fk_1;`,
+    );
+
+    await queryRunner.query(
+      `ALTER TABLE "scenario_locks" DROP CONSTRAINT scenario_locks_fk;`,
+    );
+
+    await queryRunner.query(`DROP TABLE "scenario_locks";`);
+  }
+}
--- a/api/apps/api/src/migrations/api/1638891599444-AddScenarioLocksTable.ts
+++ b/api/apps/api/src/migrations/api/1638891599444-AddScenarioLocksTable.ts
@@ -21,9 +21,17 @@ export class AddScenarioLocksTable1638891599444 implements MigrationInterface {
     await queryRunner.query(
       `ALTER TABLE "scenario_locks" ADD CONSTRAINT scenario_locks_pk PRIMARY KEY (scenario_id, user_id);`,
     );
+
+    await queryRunner.query(
+      `ALTER TABLE "scenario_locks" ADD CONSTRAINT scenario_locks_un UNIQUE (scenario_id);`,
+    );
   }
 
   public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "scenario_locks" DROP CONSTRAINT scenario_locks_un;`,
+    );
+
     await queryRunner.query(
       `ALTER TABLE "scenario_locks" DROP CONSTRAINT scenario_locks_pk;`,
     );
--- a/api/apps/api/src/migrations/api/1638891599444-AddScenarioLocksTable.ts
+++ b/api/apps/api/src/migrations/api/1638891599444-AddScenarioLocksTable.ts
@@ -6,7 +6,7 @@ export class AddScenarioLocksTable1638891599444 implements MigrationInterface {
       `CREATE TABLE "scenario_locks" (
               scenario_id uuid NOT NULL,
               user_id uuid NOT NULL,
-              grab_date timestamptz NOT NULL
+              created_at TIMESTAMP DEFAULT now()
              );`,
     );
 
--- a/api/apps/api/src/modules/scenarios/locks/lock.service.spec.ts
+++ b/api/apps/api/src/modules/scenarios/locks/lock.service.spec.ts
@@ -1,90 +1,69 @@
 import { Test } from '@nestjs/testing';
 import { getRepositoryToken } from '@nestjs/typeorm';
 import { EntityManager } from 'typeorm';
-import * as faker from 'faker';
+import { Either, left, right } from 'fp-ts/Either';
 
 import { FixtureType } from '@marxan/utils/tests/fixture-type';
 import {
+  AcquireFailure,
   lockedScenario,
   lockNotFoundError,
   LockService,
-} from '@marxan-api/modules/scenarios/locks/lock.service';
-import {
-  JobStatus,
-  Scenario,
-  ScenarioType,
-} from '@marxan-api/modules/scenarios/scenario.api.entity';
-import { ScenarioLockEntity } from '@marxan-api/modules/scenarios/locks/scenario.lock.entity';
-import { User } from '@marxan-api/modules/users/user.api.entity';
-import { left, right } from 'fp-ts/Either';
+  ReleaseFailure,
+} from './lock.service';
+import { ScenarioLockEntity } from './scenario.lock.entity';
 
 let fixtures: FixtureType<typeof getFixtures>;
 
-describe('Lock service test suite', () => {
-  beforeEach(async () => {
-    fixtures = await getFixtures();
-  });
+beforeEach(async () => {
+  fixtures = await getFixtures();
+});
 
-  it(`should be able to grab lock if lock is available`, async () => {
-    await fixtures.GivenNoLockHasBeenGrabbed();
-    await fixtures.ThenShouldBeAbleToGrabNewLockForGivenScenario();
-  });
+it(`should be able to grab lock if lock is available`, async () => {
+  await fixtures.GivenLockForScenarioIsNotFound();
+  const lock = await fixtures.WhenAcquiringALock();
+  await fixtures.ThenScenarioIsLocked(lock);
+});
 
-  it(`should not be able to grab lock if lock is not available`, async () => {
-    await fixtures.GivenALockHasBeenGrabbed();
-    await fixtures.ThenShouldNotBePossibleToGrabNewLockForGivenScenario();
-  });
+it(`should not be able to grab lock if lock is not available`, async () => {
+  await fixtures.GivenLockForScenarioIsFound();
+  const lock = await fixtures.WhenAcquiringALock();
+  await fixtures.ThenALockedScenarioErrorIsReturned(lock);
+});
 
-  it(`should be able to release the lock if lock exists`, async () => {
-    await fixtures.GivenALockHasBeenGrabbed();
-    await fixtures.ThenSameUserShouldBeAbleToReleaseLock();
-  });
+it(`should be able to release the lock if lock exists`, async () => {
+  await fixtures.GivenLockForScenarioIsFound();
+  const lock = await fixtures.WhenSameUserReleasesTheLock();
+  await fixtures.ThenScenarioIsNotLocked(lock);
+});
 
-  it(`should not be able to release the lock if lock exists for a different user`, async () => {
-    await fixtures.GivenNoLockHasBeenGrabbed();
-    await fixtures.ThenOtherUserShouldNotBeAbleToReleaseLock();
-  });
+it(`should not be able to release the lock if lock exists for a different user`, async () => {
+  await fixtures.GivenLockForScenarioIsNotFound();
+  const lock = await fixtures.WhenDifferentUserReleasesTheLock();
+  await fixtures.ThenScenarioIsStillLocked(lock);
+});
 
-  it(`should not be able to release the lock if lock does not exist`, async () => {
-    await fixtures.GivenNoLockHasBeenGrabbed();
-    await fixtures.ThenShouldNotBeAbleToReleaseLock();
-  });
+it(`should not be able to release the lock if lock does not exist`, async () => {
+  await fixtures.GivenLockForScenarioIsNotFound();
+  const lock = await fixtures.WhenSameUserReleasesTheLock();
+  await fixtures.ThenALockNotFoundErrorIsReturned(lock);
+});
 
-  it(`isLocked should return true if a lock exists`, async () => {
-    await fixtures.GivenALockExists();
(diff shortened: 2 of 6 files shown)


api/apps/api/src/modules/scenarios/locks/lock.service.ts
```diff
@@ -0,0 +1,58 @@
...
+  ) {}
+
+  async acquireLock(scenarioId: string, userId: string): Promise<void> {
+    return this.connection.transaction(async (entityManager) => {
+      const existingLock = await entityManager.find(ScenarioLockEntity, {
+        where: { scenarioId, userId },
+      });
+
+      if (existingLock.length > 0) {
+        throw new LockedScenarioError();
```
kgajowy, line 24: I think it would be nice to keep the convention of returning `Either` with possible errors (to at lest let consumer know what can go wrong within intention). 
Also Error instances themselves are bit trickier to handle for consumer.

api/apps/api/src/modules/scenarios/locks/lock.service.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  async isLocked(scenarioId: string): Promise<boolean> {
+    return this.connection.transaction(async (entityManager) => {
+      const locks = await entityManager.find(ScenarioLockEntity, {
+        where: { scenarioId },
+      });
+
+      return locks.length > 0;
+    });
```
kgajowy, line 56: Could be the transaction is not necessary - as the response passed down can be anyway invalid afterwards.

api/apps/api/src/modules/scenarios/locks/scenario.lock.entity.ts
```diff
@@ -0,0 +1,22 @@
...
+    type: `uuid`,
+    name: `scenario_id`,
+  })
+  scenarioId!: string;
+
+  @Column({
+    type: `date`,
+    name: `grab_date`,
+  })
+  grabDate!: Date;
```
kgajowy, line 21: [minor] just `createdAt` to keep convention

api/apps/api/src/migrations/api/1638891599444-AddScenarioLocksTable.ts
```diff
@@ -0,0 +1,41 @@
...
+      `ALTER TABLE "scenario_locks" ADD CONSTRAINT scenario_locks_fk FOREIGN KEY (scenario_id) REFERENCES scenarios(id) ON DELETE CASCADE;`,
+    );
+
+    await queryRunner.query(
+      `ALTER TABLE "scenario_locks" ADD CONSTRAINT scenario_locks_fk_1 FOREIGN KEY (user_id) REFERENCES users(id) ON DELETE CASCADE;`,
+    );
+
+    await queryRunner.query(
+      `ALTER TABLE "scenario_locks" ADD CONSTRAINT scenario_locks_pk PRIMARY KEY (scenario_id, user_id);`,
+    );
```
kgajowy, line 23: To keep it extra safe, we could add unique index on `scenarioId` (otherwise, formally, one scenario could have multiple locks)

api/apps/api/src/modules/scenarios/locks/lock.service.ts
```diff
@@ -0,0 +1,58 @@
...
+import { ScenarioLockEntity } from '@marxan-api/modules/scenarios/locks/scenario.lock.entity';
+import { LockedScenarioError } from './errors/locked.scenario.error';
+import { LockNotFoundError } from './errors/lock.not.found.error';
+
+@Injectable()
+export class LockService {
+  constructor(
+    @InjectRepository(ScenarioLockEntity)
+    private readonly locksRepo: Repository<ScenarioLockEntity>,
+    private connection: Connection,
```
kgajowy, line 14: You can acquire EntityManager directly from repo, reaching to connection may not be necessary.

kgajowy (commented):
The only thing I would love to see updated is returning errors instead throwing - others are rather minor suggestions.

api/apps/api/src/modules/scenarios/locks/lock.service.ts
ikas: @kgajowy https://github.com/Vizzuality/marxan-cloud/pull/673/commits/f8f415313c7d05aa8793f34fc7107515a4958e76

api/apps/api/src/modules/scenarios/locks/lock.service.ts
ikas: @kgajowy https://github.com/Vizzuality/marxan-cloud/pull/673/commits/f8f415313c7d05aa8793f34fc7107515a4958e76

api/apps/api/src/modules/scenarios/locks/scenario.lock.entity.ts
ikas: https://github.com/Vizzuality/marxan-cloud/pull/673/commits/4089893cc9412c2ea89e98dd68326c806486fa6d

api/apps/api/src/migrations/api/1638891599444-AddScenarioLocksTable.ts
ikas: https://github.com/Vizzuality/marxan-cloud/pull/673/commits/4089893cc9412c2ea89e98dd68326c806486fa6d

api/apps/api/src/modules/scenarios/locks/lock.service.ts
ikas: https://github.com/Vizzuality/marxan-cloud/pull/673/commits/4089893cc9412c2ea89e98dd68326c806486fa6d

api/apps/api/src/modules/scenarios/locks/scenario.lock.entity.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  @PrimaryColumn({
+    type: `uuid`,
+    name: `scenario_id`,
+  })
+  scenarioId!: string;
+
+  @Column({
+    type: `date`,
+    name: `grab_date`,
```
kgajowy, line 19: as this is brand new entity anyway, is there any reason to not change the column name as well? Am I missing anything?

api/apps/api/src/modules/scenarios/locks/lock.service.ts
```diff
@@ -0,0 +1,71 @@
...
+    });
+  }
+
+  async isLocked(scenarioId: string): Promise<boolean> {
+    const locks = await this.locksRepo.find({
+      where: { scenarioId },
+    });
+
+    return locks.length > 0;
+  }
```
kgajowy, line 70: [minor] just another thought - `count()` 🤷

api/apps/api/src/modules/scenarios/locks/scenario.lock.entity.ts
kgajowy: also `@CreateDateColumn({ name: 'created_at', type: 'timestamp without time zone' })`

api/apps/api/src/modules/scenarios/locks/lock.service.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+async function getFixtures() {
+  const fakeUser: User = {
+    id: 'fake-user',
+    email: faker.internet.email(),
+    passwordHash: faker.random.uuid(),
+    isActive: true,
+    isDeleted: false,
+    projects: [],
+    scenarios: [],
+  };
```
kgajowy, line 73: do we need whole User and reference it at all?

api/apps/api/src/modules/scenarios/locks/lock.service.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+let fixtures: FixtureType<typeof getFixtures>;
+
+describe('Lock service test suite', () => {
+  beforeEach(async () => {
+    fixtures = await getFixtures();
+  });
+
+  it(`should be able to grab lock if lock is available`, async () => {
+    await fixtures.GivenNoLockHasBeenGrabbed();
+    await fixtures.ThenShouldBeAbleToGrabNewLockForGivenScenario();
```
kgajowy, line 30: Missing actual `When` as an action. Same applies to checks below.

kgajowy:
```
GivenScenarioIsNotLocked()
WhenAcquiringALock()
ThenScenarioIsLocked()
```

api/apps/api/src/modules/scenarios/locks/lock.service.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+import { getRepositoryToken } from '@nestjs/typeorm';
+import { EntityManager } from 'typeorm';
+import * as faker from 'faker';
+
+import { FixtureType } from '@marxan/utils/tests/fixture-type';
+import {
+  lockedScenario,
+  lockNotFoundError,
+  LockService,
+} from '@marxan-api/modules/scenarios/locks/lock.service';
```
kgajowy, line 11: relative imports (and entity below as well)

api/apps/api/src/modules/scenarios/locks/lock.service.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+  Scenario,
+  ScenarioType,
+} from '@marxan-api/modules/scenarios/scenario.api.entity';
+import { ScenarioLockEntity } from '@marxan-api/modules/scenarios/locks/scenario.lock.entity';
+import { User } from '@marxan-api/modules/users/user.api.entity';
+import { left, right } from 'fp-ts/Either';
+
+let fixtures: FixtureType<typeof getFixtures>;
+
+describe('Lock service test suite', () => {
```
kgajowy, line 23: [minor[ whole `describe` is redundant and not giving any value in particular - only increases indentation level.

api/apps/api/src/modules/scenarios/locks/lock.service.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+          userId: fakeUser.id,
+          createdAt: new Date(),
+        },
+      ]);
+    },
+    ThenShouldBeAbleToGrabNewLockForGivenScenario: async () => {
+      const mockDate = new Date('2021-02-26T22:42:16.652Z');
+      const spy = jest
+        .spyOn(global, 'Date')
+        .mockImplementation(() => (mockDate as unknown) as string);
```
kgajowy, line 148: why do we even mock date? as far as I can tell, it isn't used anywhere (but may be should, what do we do about non-released, long-lasting locks?)

api/apps/api/src/modules/scenarios/locks/lock.service.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+      ).resolves.toStrictEqual(right(void 0));
+
+      expect(mockEntityManager.find).toHaveBeenCalledWith(ScenarioLockEntity, {
+        where: { scenarioId: fakeScenario.id, userId: fakeUser.id },
+      });
+
+      expect(mockEntityManager.save).toHaveBeenCalledWith({
+        scenarioId: fakeScenario.id,
+        userId: fakeUser.id,
+        createdAt: mockDate,
```
kgajowy, line 161: `createdAt: expect.any(Date)`

api/apps/api/src/modules/scenarios/locks/scenario.lock.entity.ts
ikas: no, you're right, I meant to change the column name too but I forgot, will make the change :+1:

api/apps/api/src/modules/scenarios/locks/lock.service.spec.ts
ikas: it will be used in the future (https://vizzuality.atlassian.net/browse/MARXAN-1091)

api/apps/api/src/modules/scenarios/locks/lock.service.spec.ts
ikas: nope, neither the scenario down below :ok_hand:

api/apps/api/src/modules/scenarios/locks/lock.service.spec.ts
kgajowy: @ikas if you set a watcher in specs, it shall run all tests affected by change - but you are right, that could be handy - so may be matter of preference (running/reading - I still bet on reading as it happens often 🤷 )

api/apps/api/src/modules/scenarios/locks/lock.service.spec.ts
kgajowy: 👍 for long-lasting locks, but do we need to mock global Date?

api/apps/api/src/modules/scenarios/locks/lock.service.ts
```diff
@@ -0,0 +1,67 @@
...
+        where: { scenarioId, userId },
+      });
+
+      if (existingLock.length < 1) {
+        return left(lockNotFoundError);
+      }
+
+      await entityManager.remove(existingLock[0]);
+
+      return right(void 0);
```
kgajowy, line 60: Wouldn't it be easier to just `delete` on given criteria? 🤔

api/apps/api/src/modules/scenarios/locks/scenario.lock.entity.ts
```diff
@@ -0,0 +1,19 @@
...
+  })
+  userId!: string;
+
+  @PrimaryColumn({
+    type: `uuid`,
+    name: `scenario_id`,
+  })
+  scenarioId!: string;
+
+  @CreateDateColumn({ name: 'created_at', type: 'timestamp' })
```
kgajowy, line 17: possibly could have `NOT NULL default now()`

kgajowy (approved):
(only 2 more polishing moves but not necessary)

api/apps/api/src/modules/scenarios/locks/lock.service.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+  await fixtures.GivenLockForScenarioIsFound();
+  const lock = await fixtures.WhenSameUserReleasesTheLock();
+  await fixtures.ThenScenarioIsNotLocked(lock);
+});
+
+it(`should not be able to release the lock if lock exists for a different user`, async () => {
+  await fixtures.GivenLockForScenarioIsNotFound();
+  const lock = await fixtures.WhenDifferentUserReleasesTheLock();
+  await fixtures.ThenScenarioIsStillLocked(lock);
+});
```
hotzevzl, line 44: I am not sure I get what we are testing here - reading the fixtures code, what happens makes sense (and indeed the test passes), but isn't `ThenScenarioIsStillLocked` misleading then? in this outcome, we don't really assert that the scenario is locked (and by extension that it is _still_ locked), if I am not missing something.

in any case, this is very minor - I may have misunderstood or it may just be that the description of one of the steps could be made more precise.

hotzevzl (approved):

api/apps/api/src/modules/scenarios/locks/lock.service.spec.ts
ikas: sorry, I forgot to comment here, but yeah, we don't need the global mock Date (I removed it in the last commit) :+1:

api/apps/api/src/modules/scenarios/locks/lock.service.ts
ikas: yep, that's true, on it :+1:

api/apps/api/src/modules/scenarios/locks/scenario.lock.entity.ts
ikas: Sorry Kamil, I didn't understand this comment: I added `NOT NULL default now()` in the migration that creates the column - do you mean we should add that here as well?

api/apps/api/src/modules/scenarios/locks/lock.service.spec.ts
ikas: ahah, yeah, the name is quite misleading, apologies for that :sweat_smile: I'll update the name, thanks for pointing out!!

api/apps/api/src/modules/scenarios/locks/scenario.lock.entity.ts
kgajowy: @ikas I believe so - in the end, we should aim at restoring the possibility to run migration generator - but sure this requires much more work (so adding it there would just make less effort later). Up2 you : )

ikas: you mean like this? https://github.com/Vizzuality/marxan-cloud/pull/673/commits/eba3a61e200d29e9c3bca93accda678ade925e3e

kgajowy: @ikas ✅ 👍

PR #674, opened 2021-12-07
feat(cloning): file repository

kgajowy:
# Scope

Introduces FileRepository to be used within `cloning` feature (and possibly others).
Provides "local" development implementation - but may as well become "target" one.

# Key changes
Environment variable isn't read from existing `AppConfig` for reasons:
* duplicating json configs
* FileRepository is a library
* avoids "global" config with everything inside - focuses on per-module needs
* leverages from built-in NestJS tooling

# Key notes / future advisory
* consider remove/unlink to be "standalone" - to avoid allowing removal of non-owned files (?) - to be discussed, who should be responsible and when this should happen (for example: add `remove` indeed to this module but each feature, like cloning, would have a wrapped storing all related files, may be even for given resource...)
* when deleting, be aware of path traversal issues (providing `../../../etc/passwd` like) and poison null byte (`\0`)
* `local/temporary` solution isn't error-safe. Consider updating it if it happens to become `current` solution

### DIFF
--- a/api/apps/api/test/files-repository/files-repository.e2e-spec.ts
+++ b/api/apps/api/test/files-repository/files-repository.e2e-spec.ts
@@ -1,0 +1,75 @@
+import { createReadStream } from 'fs';
+import { Readable } from 'stream';
+import { isRight, Right } from 'fp-ts/Either';
+import { Test } from '@nestjs/testing';
+
+import { FixtureType } from '@marxan/utils/tests/fixture-type';
+import {
+  FileRepository,
+  FileRepositoryModule,
+  TempStorageRepository,
+} from '@marxan/files-repository';
+
+let fixtures: FixtureType<typeof getFixtures>;
+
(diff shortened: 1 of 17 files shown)


api/apps/api/test/files-repository/files-repository.e2e-spec.ts
```diff
@@ -0,0 +1,75 @@
...
+  const app = await Test.createTestingModule({
+    imports: [FileRepositoryModule],
+  })
+    .overrideProvider(FileRepository)
+    .useClass(TempStorageRepository)
+    .compile();
+  const sut = app.get(FileRepository);
+
+  return {
+    cleanup: async () => app.close(),
```
kgajowy, line 47: Not removing the files - see PR description. Shouldn't be a big deal for development and CI.

AaronPerezPerez (approved):

PR #676, opened 2021-12-07
libbot: update ProtectedAreas module to use newest protected areas workflow [MARXAN-1101]

hotzevzl:
https://vizzuality.atlassian.net/browse/MARXAN-1101

### DIFF
--- a/data/bots/lib/libbot/geo-features.ts
+++ b/data/bots/lib/libbot/geo-features.ts
@@ -10,14 +10,14 @@ export class GeoFeatures {
     this.baseHttpClient = httpClient.baseHttpClient;
   }
 
-  async getIdOfFeaturesBySubstringMatchOnName(
+  async searchFeaturesWithinProject(
     projectId: string,
-    name: string,
+    nameFragment: string,
   ): Promise<string[]> {
     const opStart = Process.hrtime();
 
     const id = await this.baseHttpClient.get(
-      `/projects/${projectId}/features?q=${name}&fields=id`,
+      `/projects/${projectId}/features?q=${nameFragment}&fields=id`,
     )
       .then(getJsonApiDataFromResponse)
       .then((data: { id: string; type: string }[]) => {
@@ -26,7 +26,7 @@ export class GeoFeatures {
       .catch(logError);
 
     logInfo(
-      `Id of feature matching query string ${name} retrieved in ${
+      `Id of feature matching query string ${nameFragment} retrieved in ${
         tookMs(Process.hrtime(opStart))
       }ms.`,
     );
--- a/data/bots/lib/libbot/geo-features.ts
+++ b/data/bots/lib/libbot/geo-features.ts
@@ -1,5 +1,7 @@
+import Process from "https://deno.land/std@0.103.0/node/process.ts";
 import { BotHttpClient, getJsonApiDataFromResponse } from "./marxan-bot.ts";
-import { logError } from "./logger.ts";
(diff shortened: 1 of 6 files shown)


data/bots/lib/libbot/protected-areas.ts
```diff
@@ -4,13 +4,57 @@ import { IUCNCategory } from "./scenarios.ts";
 import { logDebug, logError, logInfo } from "./logger.ts";
 import { tookMs } from "./util/perf.ts";
 
+export enum ProtectedAreaKind {
+  Global = "global",
+  Project = "project",
+}
+
+interface ProtectedAreaSelection {
+  id: IUCNCategory;
```
kgajowy, line 13: formally it is `id: IUCNCategory | string;`

kgajowy (approved):

kgajowy (commented):
Also, shouldn't `GET` for available PA be updated?

data/bots/lib/libbot/protected-areas.ts
hotzevzl: good catch, thank you! fixed in f72eb072fa3fc18e8ed2672f218e97111acb46f3

hotzevzl:
> Also, shouldn't `GET` for available PA be updated?

It should, but I've been focusing on what we needed for a specific training bot, and all we need here is to hit the endpoint to get the list of IUCN categories of all the (WDPA) protected areas within the project's bbox, which works fine. I have created https://vizzuality.atlassian.net/browse/MARXAN-1105 for future needs (this will be needed for projects where we need to include custom protected areas).

kgajowy:
> 

Alrighty, just remember that we should rather remove the "old" endpoint and use the new one, so that we don't have deprecated code in base.

data/bots/lib/libbot/geo-features.ts
```diff
@@ -8,14 +10,33 @@ export class GeoFeatures {
...
   }
 
-  async getIdFromQueryStringInProject(
+  async getIdOfFeaturesBySubstringMatchOnName(
     projectId: string,
     name: string,
-  ) {
-    return await this.baseHttpClient.get(
+  ): Promise<string[]> {
+    const opStart = Process.hrtime();
```
kgajowy, line 17: we may want to consider adding pretty decorator for those (like https://github.com/norbornen/execution-time-decorator/blob/master/src/index.ts implementation or package directly) to avoid bloating code functions

kgajowy (approved):

data/bots/lib/libbot/geo-features.ts
hotzevzl: Yes, I have been postponing this, but inline hrtime diffs as a rule rather than as a one-off is ugly.

I have created a technical-debt task as I won't be able to take care of this while I try to close outstanding stuff before the break: https://vizzuality.atlassian.net/browse/MARXAN-1138

data/bots/lib/libbot/geo-features.ts
```diff
@@ -8,14 +10,33 @@ export class GeoFeatures {
     this.baseHttpClient = httpClient.baseHttpClient;
   }
 
-  async getIdFromQueryStringInProject(
+  async getIdOfFeaturesBySubstringMatchOnName(
```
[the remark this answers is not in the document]
hotzevzl: what was I thinking 😅 

I have amended this in 172121bd02888c0cfb32017b4537b48c43c8b040, slightly differently than your suggestion but in that sane general direction.

PR #678, opened 2021-12-09
CanCreateProject check in project endpoint

rubvalave:
## As above title says, pretty straightforward.

- Use of check inside the project-service correspondent method.
- Use of monad to handle errors.


ikas (approved):
:rocket: :1st_place_medal:

kgajowy (approved):

PR #683, opened 2021-12-10
Add is_custom calculated column to features

ikas:
### Overview

This PR adds a new column `is_custom` to the `features` table in the API DB. This column is calculated according to whether the row has set a `project_id` or not. This column is returned as the attribute `isCustom` when requesting features, and it will allow the FE to easily sort the returned features by whether they were user imported or not.

### Feature relevant tickets

[MARXAN-967](https://vizzuality.atlassian.net/browse/MARXAN-967)

### DIFF
--- a/api/apps/api/src/modules/geo-features/geo-feature.api.entity.ts
+++ b/api/apps/api/src/modules/geo-features/geo-feature.api.entity.ts
@@ -94,6 +94,10 @@ export class GeoFeature extends BaseEntity {
     referencedColumnName: 'id',
   })
   createdBy?: User;
+
+  @ApiPropertyOptional()
+  @Column('boolean', { name: 'is_custom' })
+  isCustom?: boolean;
 }
 
 export class JSONAPIGeoFeaturesData {
--- a/api/apps/api/test/geo-features.e2e-spec.ts
+++ b/api/apps/api/test/geo-features.e2e-spec.ts
@@ -4,6 +4,7 @@ import { PromiseType } from 'utility-types';
 import {
   GeoFeature,
   geoFeatureResource,
+  JSONAPIGeoFeaturesData,
 } from '@marxan-api/modules/geo-features/geo-feature.api.entity';
 import { tearDown } from './utils/tear-down';
 import { bootstrapApplication } from './utils/api-application';
@@ -78,6 +79,9 @@ describe('GeoFeaturesModule (e2e)', () => {
         expect(response.body.data[0].attributes.featureClassName).toEqual(
           geoFeaturesFilters.cheeta.featureClassName,
         );
+        expect(response.body.data[0].attributes.isCustom).toEqual(
+          !!response.body.data[0].attributes.projectId,
+        );
       });
 
       test.skip('should return a single result of geo-features whose alias property matches a given filter', async () => {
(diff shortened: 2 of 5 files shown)


api/apps/api/src/modules/geo-features/geo-feature.api.entity.ts
```diff
@@ -94,6 +94,10 @@ export class GeoFeature extends BaseEntity {
     referencedColumnName: 'id',
   })
   createdBy?: User;
+
+  @ApiPropertyOptional()
+  @Column('boolean', { name: 'is_custom' })
+  isCustom?: boolean;
```
hotzevzl, line 100: do we need to add this to `get GeoFeaturesService.serializerConfig()`?

honestly, I don't remember - I haven't added new fields and exposed them through the JSON:API payloads for quite some time now 😬

hotzevzl (approved):
thanks @ikas! Other than needing to add the property to the JSON:API config, this looks good to me.

api/apps/api/src/modules/geo-features/geo-feature.api.entity.ts
ikas: yeah, I think we do, and I think that's why E2E tests are failing! Thanks for pointing out, I'm force pushing an updated commit!

api/apps/api/test/geo-features.e2e-spec.ts
```diff
@@ -78,6 +79,9 @@ describe('GeoFeaturesModule (e2e)', () => {
         expect(response.body.data[0].attributes.featureClassName).toEqual(
           geoFeaturesFilters.cheeta.featureClassName,
         );
+        expect(response.body.data[0].attributes.isCustom).toEqual(
+          !!response.body.data[0].attributes.projectId,
+        );
```
kgajowy, line 84: I don't fully like the way we are doing kind of `white box testing` in e2e/integration specs 🤔 

may be we can extend `api/apps/api/test/upload-feature/upload-feature.e2e-spec.ts` (or reuse its fixtures?) instead relying on magic seeds?

(could be minor, but our future `us` would demand its more important 😅 )

ikas: I don't think we need a dedicated test suite for this: we are formalizing an assumption that features with no project associated are global, while features associated to a project are custom - so, we can take advantage of the existing tests to assert exactly these assumptions, I don't see the need for further tests or seeds :)

In the end, I ended up changing things a bit, but in essence what I'm doing is:
* testing that all features in the `api/apps/api/test/geo-features.e2e-spec.ts` suite have `isCustom` set to `false`;
* testing that features uploaded as part of the `api/apps/api/test/upload-feature/upload-feature.e2e-spec.ts` suite have `isCustom` set to `true`;

Let me know if you have any concerns with this approach, or if you prefer me to test this in a different way :)

kgajowy: @ikas 
> I don't think we need a dedicated test suite for this
all good, I mean exactly those you mentioned

>  I don't see the need for further tests or seeds :)
My issue with with using seeds that come from .sql whiles (`make seeds-db` etc) so tests are vulnerable to changing those.


👍

kgajowy (approved):

PR #684, opened 2021-12-10
CanView in project GetBLMValues endpoint 

rubvalave:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/blm/values/project-blm-repo.ts
+++ b/api/apps/api/src/modules/blm/values/project-blm-repo.ts
@@ -3,10 +3,14 @@ import { Either } from 'fp-ts/Either';
 export const unknownError = Symbol(`unknown error`);
 export const projectNotFound = Symbol(`project not found`);
 export const alreadyCreated = Symbol(`project already has defaults`);
+export const forbidden = Symbol(`access unauthorized`);
 
 export type CreateFailure = typeof unknownError | typeof alreadyCreated;
 export type SaveFailure = typeof unknownError | typeof projectNotFound;
-export type GetFailure = typeof unknownError | typeof projectNotFound;
+export type GetFailure =
+  | typeof unknownError
+  | typeof projectNotFound
+  | typeof forbidden;
 
 export interface ProjectBlm {
   id: string;
--- a/api/apps/api/src/modules/blm/values/project-blm-repo.ts
+++ b/api/apps/api/src/modules/blm/values/project-blm-repo.ts
@@ -3,15 +3,9 @@ import { Either } from 'fp-ts/Either';
 export const unknownError = Symbol(`unknown error`);
 export const projectNotFound = Symbol(`project not found`);
 export const alreadyCreated = Symbol(`project already has defaults`);
-export const forbidden = Symbol(`access unauthorized`);
-
 export type CreateFailure = typeof unknownError | typeof alreadyCreated;
 export type SaveFailure = typeof unknownError | typeof projectNotFound;
-export type GetFailure =
-  | typeof unknownError
-  | typeof projectNotFound
-  | typeof forbidden;
-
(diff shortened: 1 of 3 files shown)


api/apps/api/src/modules/blm/values/project-blm-repo.ts
```diff
@@ -3,10 +3,14 @@ import { Either } from 'fp-ts/Either';
 export const unknownError = Symbol(`unknown error`);
 export const projectNotFound = Symbol(`project not found`);
 export const alreadyCreated = Symbol(`project already has defaults`);
+export const forbidden = Symbol(`access unauthorized`);
```
kgajowy, line 6: why would blm-repo need to know anything about ACL? it isn't within its boundaries.

api/apps/api/src/modules/projects/projects.controller.ts
```diff
@@ -323,13 +324,19 @@ export class ProjectsController {
...
-      throw new NotFoundException(
-        `Could not find project BLM values for project with ID: ${id}`,
-      );
+      switch (result.left) {
+        case forbidden:
+          throw new ForbiddenException();
+        default:
+          throw new NotFoundException(
+            `Could not find project BLM values for project with ID: ${id}`,
+          );
```
kgajowy, line 338: without exhaustive check it can be overlooked and throw 404 for something totally unrelated - actually any error, so can be misleading.

```
const error = response.left;
    switch (error) {
      case metadataNotFound:
      case outputMetadataNotFound:
        throw new NotFoundException(`Marxan was not yet executed.`);
      case outputZipNotYetAvailable:
      case marxanFailed:
      case inputZipNotYetAvailable:
        throw new NotFoundException(
          `Marxan has not yet finished or finished with error.`,
        );
      default:
        const _exhaustiveCheck: never = error;
        throw _exhaustiveCheck;
    }
```
(this is, if any new option is added, TS/IDE will force you to add there something)

api/apps/api/src/modules/blm/values/project-blm-repo.ts
kgajowy: in this case, it may belong to `projects`

api/apps/api/src/modules/projects/projects.controller.ts
rubvalave: I just added the same thing to handle ACL errors properly to this PR and apart from that improved how the errors are handled here I think so it could only throw an 404 for a `projectNotFound` error. All others will remain the same though.

kgajowy: @rubvalave still missing `_exhaustiveCheck` to ensure we can react once some new error is added to possible outcomes.

AaronPerezPerez (approved):

api/apps/api/src/modules/projects/projects.controller.ts
```diff
@@ -323,13 +325,21 @@ export class ProjectsController {
...
+      switch (result.left) {
+        case projectNotFound:
+          throw new NotFoundException(
+            `Could not find project BLM values for project with ID: ${id}`,
+          );
+        case forbiddenError:
+          throw new ForbiddenException();
+        default:
+          throw new InternalServerErrorException();
+      }
```
hotzevzl, line 342: nothing to do here and now, but I think we'll end up with a lot of almost identical boilerplate in most controller handlers, eventually, as we phase in access control throughout - it may be a good idea to think about a utility function that handles at least the default cases we expect everywhere (or nearly) such as not found, forbidden and "all the rest"...

hotzevzl (approved):

api/apps/api/src/modules/projects/projects.controller.ts
rubvalave: I think it is sorted out now

PR #694, opened 2021-12-15
Improvements to get project users endpoint

ikas:
### Overview

This PR implements improvements to the `GET api/v1/roles/projects/{id}/users` endpoint, as described in [MARXAN-1121](https://vizzuality.atlassian.net/browse/MARXAN-1121).

### Feature relevant tickets

[MARXAN-1121](https://vizzuality.atlassian.net/browse/MARXAN-1121)

### DIFF
--- a/api/apps/api/src/modules/access-control/projects-acl/project-acl.controller.ts
+++ b/api/apps/api/src/modules/access-control/projects-acl/project-acl.controller.ts
@@ -12,6 +12,7 @@ import {
   HttpCode,
   ForbiddenException,
   HttpStatus,
+  Query,
 } from '@nestjs/common';
 import { JwtAuthGuard } from '@marxan-api/guards/jwt-auth.guard';
 import { ProjectAclService } from './project-acl.service';
@@ -46,10 +47,12 @@ export class ProjectAclController {
   async findUsersInProject(
     @Param('projectId', ParseUUIDPipe) projectId: string,
     @Req() req: RequestWithAuthenticatedUser,
+    @Query('q') nameSearch?: string,
   ): Promise<UsersInProjectResult> {
     const result = await this.projectAclService.findUsersInProject(
       projectId,
       req.user.id,
+      nameSearch,
     );
 
     if (isLeft(result)) {
--- a/api/apps/api/src/modules/access-control/projects-acl/project-acl.controller.ts
+++ b/api/apps/api/src/modules/access-control/projects-acl/project-acl.controller.ts
@@ -23,8 +23,11 @@ import {
   ApiOperation,
   ApiTags,
 } from '@nestjs/swagger';
-import { UserRoleInProjectDto } from './dto/user-role-project.dto';
 import { isLeft } from 'fp-ts/lib/These';
+import {
+  UserRoleInProjectDto,
+  UsersInProjectResult,
+} from './dto/user-role-project.dto';
 
 @UseGuards(JwtAuthGuard)
 @ApiBearerAuth()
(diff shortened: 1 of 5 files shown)


tiagojsag (approved):

api/apps/api/src/modules/access-control/projects-acl/project-acl.controller.ts
```diff
@@ -43,17 +47,19 @@ export class ProjectAclController {
   async findUsersInProject(
     @Param('projectId', ParseUUIDPipe) projectId: string,
     @Req() req: RequestWithAuthenticatedUser,
-  ): Promise<UserRoleInProjectDto[] | boolean> {
+    @Query('q') nameSearch?: string,
+  ): Promise<UsersInProjectResult> {
```
hotzevzl, line 51: this comment is not actually about a change in this PR, as I hadn't noticed this earlier, but we are not returning a _real_ JSON:API response payload here - I assume frontends are fine with this since the implementation so far wasn't going through the JSON:API serializer in any case, but I suggest to try avoiding using different response formats across the API (except for things such as GeoJSON or other non-JSON-document-kind-of content types)

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -148,22 +148,32 @@ export class ProjectAclService implements ProjectAccessControl {
...
         'userId.id',
         'userId.avatarDataUrl',
-      ])
-      .getMany();
+      ]);
+
+    if (nameSearch) {
+      query.andWhere('userId.displayName ILIKE :name', {
+        name: `%${nameSearch}%`,
+      });
```
hotzevzl, line 173: I see no Bobby Tables issue with this 🤔

hotzevzl (commented):
looks good, with the main question mark being not serialising the response as JSON:API

api/apps/api/src/modules/access-control/projects-acl/project-acl.controller.ts
rubvalave: I started reading about the JSON:API payloads and examples, so there would be a way, prior to returning the actual payload, where some serializer would structure the data in the way JSON:API specifies, right? Or am I understanding this incorrectly?

rubvalave: I say so because if it is used elsewhere, I can try to use or adapt it here so it follows the same principles.

hotzevzl: > so there would be a way, prior to returning the actual payload, where some serializer would structure the data in the way JSON:API specifies, right?

Yes, that's right - all the CRUD endpoints (or most of them, if I got distracted and didn't notice) should use JSON:API as response format (and more or less so also for GET query params).

Since I won't be around to help you with this over the coming days, and since this endpoint was not using a JSON:API response format so far in any case, I suggest to merge this PR as is, and in January I can onboard you to the JSON:API strategy we've used elsewhere so that you can apply the same in other endpoints when you get to these.

In the meanwhile if you want to have a look around, take for example `api/apps/api/src/modules/users/users.service.ts`

In most of these classes, the `get serializerConfig()` getter returns the configuration for the shape of JSON:API responses that we then feed to the serializer (`jsonapi-serializer`) through the `AppBaseService.serialize()` method.

To close the loop, have a look at `UsersController.findAll()`, for example, where the fetching of data is followed by its serialization, whose results we finally send in the response.

hotzevzl (approved):

PR #701, opened 2021-12-16
throw an error if API response is not 201 [MARXAN-1142]

hotzevzl:
Maybe this is too restrictive? But given the use case, and that we
should be expecting exactly a 201, I would go for 201 only nevertheless.

https://vizzuality.atlassian.net/browse/MARXAN-1142

### DIFF
--- a/data/bots/core-demos/demo-okavango/core.ts
+++ b/data/bots/core-demos/demo-okavango/core.ts
@@ -36,7 +36,7 @@ export const runBot = async (settings: MarxanBotConfig) => {
     name: "Okavango " + crypto.randomUUID(),
     description: "",
     planningAreaId: planningAreaId,
-    planningUnitGridShape: 'from_shapefile',
+    planningUnitGridShape: "from_shapefile",
   });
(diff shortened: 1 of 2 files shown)


kgajowy:
> Maybe this is too restrictive? But given the use case, and that we should be expecting exactly a 201, I would go for 201 only nevertheless.
> 
> https://vizzuality.atlassian.net/browse/MARXAN-1142

it may be `fragile` - how about 
```
const isSuccess = (statusCode: number) => statusCode >= 200 && statusCode < 400
```
or similar?

kgajowy (commented):
https://github.com/Vizzuality/marxan-cloud/pull/701#issuecomment-996508501

hotzevzl:
> [#701 (comment)](https://github.com/Vizzuality/marxan-cloud/pull/701#issuecomment-996508501)

fixed in d1a680c7fdedccc6adb2d61f61ad89e9f88a76ab

kgajowy (approved):

PR #702, opened 2021-12-17
Improve native execution

tiagojsag:
This PR should help a bit with running things fully natively. I also added a few more env vars (which I suspect would be needed in the future anyway) and some basic docs here and there. Functionally this should have little to no impact in anyone's current setup... maybe... probably... sometimes


Makefile
```diff
@@ -1,18 +1,23 @@
...
+endif
+
 CIENV := $(if $(filter $(environment), ci), -f docker-compose-test-e2e.ci.yml , -f docker-compose-test-e2e.local.yml)
 API_DB_INSTANCE := $(if $(environment), test-e2e-postgresql-api, postgresql-api)
 GEO_DB_INSTANCE := $(if $(environment), test-e2e-postgresql-geo-api, postgresql-geo-api)
 REDIS_INSTANCE := $(if $(environment), test-e2e-redis, redis)
-_API_POSTGRES_USER := $(if $(filter $(environment), ci),${API_POSTGRES_USER},$(shell grep -e API_POSTGRES_USER ${ENVFILE} | sed 's/^.*=//'))
-_API_POSTGRES_DB := $(if $(filter $(environment), ci),${API_POSTGRES_DB},$(shell grep -e API_POSTGRES_DB ${ENVFILE} | sed 's/^.*=//'))
-_GEO_POSTGRES_USER := $(if $(filter $(environment), ci),${GEO_POSTGRES_USER},$(shell grep -e GEO_POSTGRES_USER ${ENVFILE} | sed 's/^.*=//'))
-_GEO_POSTGRES_DB := $(if $(filter $(environment), ci),${GEO_POSTGRES_DB},$(shell grep -e GEO_POSTGRES_DB ${ENVFILE} | sed 's/^.*=//'))
```
hotzevzl, line 15: it's been a looong time (and I barely remember details from last week, if I didn't document them), but IIRC the reason for using a prefix on these make vars was to avoid clashes with env vars (from which they *may* be set, or not, depending on the environment), so *I think* this change will break CI or running things in Docker locally.

Makefile
```diff
@@ -1,18 +1,23 @@
...
+#
 # If using the same variables in recipes that need to use a dotenv file other
 # than .env, remember to check that no values from .env are being used
 # inadvertently.
-ENVFILE := $(if $(environment), .env-test-e2e, .env)
+ENVFILE := $(if $(environment), .env-$(environment), .env)
+ifneq (,$(wildcard $(ENVFILE)))
+    include $(ENVFILE)
+    export
+endif
```
hotzevzl, line 15: I like this, but I think that piggybacking on the existing `environment` make params may break CI and/or running some things (tests) in Docker locally. It should be ok to refactor the existing use of this param to make it possible to use it both to read an environment-specific .env file and to differentiate the test runs between CI and local Docker Compose setups.

api/apps/api/config/custom-environment-variables.json
```diff
@@ -4,12 +4,22 @@
   },
   "postgresApi": {
     "url": "API_POSTGRES_URL",
+    "host": "API_POSTGRES_HOST",
+    "username": "API_POSTGRES_USER",
+    "password": "API_POSTGRES_PASSWORD",
+    "database": "API_POSTGRES_DB",
+    "port": "API_POSTGRES_PORT",
```
hotzevzl, line 11: `API_POSTGRES_URL` already includes host, user, password, db and port - can we rather compose this single envvar when running natively so that we avoid having duplication of pg connection configs?

api/apps/api/src/migrations/api/1608149578000-EnablePostgis.ts
```diff
@@ -0,0 +1,25 @@
+import { Logger } from '@nestjs/common';
+import { MigrationInterface, QueryRunner } from 'typeorm';
+import { PostgreSQLUtils } from '@marxan-api/utils/postgresql.utils';
+
+export class EnablePostgis1608149578000 implements MigrationInterface {
+  async up(queryRunner: QueryRunner): Promise<any> {
+    if (await PostgreSQLUtils.version13Plus()) {
+      await queryRunner.query(`
+CREATE EXTENSION IF NOT EXISTS postgis;
```
hotzevzl, line 9: This will enable PostGIS in the apidb, but we don't need it there.

This should not be needed when running the services natively either... are you sure this is needed?

Makefile
hotzevzl: shame on me for not clearly documenting/commenting why the dangling underscore, of course

also - `_something` may instinctively trigger "unused var/placeholder var" instinct, so the choice may not be the best. I have a vague recollection of not being able to decide on a better naming schema to differentiate the two layers of variables and ending up using a quick change to get the task done.

Makefile
tiagojsag: we briefly discussed this during a call, but for reference:
Right now, the 'environment' argument is used more like a feature flag of sorts, and not so much as a way to define different environments. This change pushes to actually restore it to being basically an identifier between different sets of configurations. If we need to further tune specific behavior, we can add env vars to said environments, to control behavior. However, I don't think that will be necessary, at least for now

api/apps/api/src/migrations/api/1608149578000-EnablePostgis.ts
tiagojsag: There are some columns of type geometry, so this is needed.

kgajowy: Which ones? pretty sure api.geo shouldn't have any (otherwise it shouldn't even boot 🤔 )

Makefile
```diff
@@ -1,18 +1,23 @@
 .PHONY: start
 
 # Read values as needed from .env
+#
+# Optionally pass 'environment=<something>' to load env vars from '.env-something' instead
+# Useful to separate i.e. dev from testing databases
```
kgajowy, line 6: [very minor] usually it is `.env.something`, like `.env.local` in most of the tooling that support it

Makefile
```diff
@@ -186,18 +191,50 @@ restore-volumes-data:
...
+	sed -e "s/\$$user/00000000-0000-0000-0000-000000000000/g" api/apps/api/test/fixtures/test-wdpa-data.sql | psql -U "${GEO_POSTGRES_USER}" -h "${GEO_POSTGRES_HOST}" ${GEO_POSTGRES_DB};
+	psql -U "${API_POSTGRES_USER}" -h "${API_POSTGRES_HOST}" ${API_POSTGRES_DB} < api/apps/api/test/fixtures/test-features.sql
+	@for i in api/apps/api/test/fixtures/features/*.sql; do \
+		table_name=`basename -s .sql "$$i"`; \
+		featureid=`psql -X -A -t -U "${API_POSTGRES_USER}" -h "${API_POSTGRES_HOST}" ${API_POSTGRES_DB} -c "select id from features where feature_class_name = '$$table_name'"`; \
+		echo "appending data for $${table_name} with id $${featureid}"; \
+		sed -e "s/\$$feature_id/$$featureid/g" api/apps/api/test/fixtures/features/$${table_name}.sql | psql -U "${GEO_POSTGRES_USER}" -h "${GEO_POSTGRES_HOST}" ${GEO_POSTGRES_DB}; \
+		done;
+
+native-seed-api-with-test-data: native-db-migrate native-seed-api-init-data | native-seed-geoapi-init-data
```
kgajowy, line 238: How those seeds differ from previous? are those related to creating users/projects, while `init-data` pushes wdpa/features and so on?

README.md
```diff
@@ -62,29 +69,39 @@ details.
...
+     * `API_POSTGRES_HOST` (string, required): host of the database server to be
+       used for the PostgreSQL connection (API)
+     * `API_POSTGRES_PORT` (number, required): port of the database server to be
+       used for the PostgreSQL connection (API)
+     * `API_POSTGRES_USER` (string, required): username to be used for the
+       PostgreSQL connection (API)
+     * `API_POSTGRES_PASSWORD` (string, required): password to be used for the
+       PostgreSQL connection (API)
+     * `API_POSTGRES_DB` (string, required): name of the database to be used for
+       the PostgreSQL connection (API)
```
kgajowy, line 83: [minor] it is always hard to maintain the .md files - for this very particular case, there is a common approach to provide `.env.example` tho for starters which document (and provides some defaults) the variables.

api/apps/api/src/utils/config.utils.spec.ts
```diff
@@ -8,6 +8,14 @@ process.env.NETWORK_CORS_ORIGINS = extraCorsOrigin;
 import { AppConfig } from './config.utils';
 
 describe('AppConfig', () => {
+  beforeAll(() => {
+    if (process.env.NODE_CONFIG_DIR !== 'apps/api/config') {
+      throw Error(
+        `Running the test suite with NODE_CONFIG_DIR=${process.env.NODE_CONFIG_DIR}, which may cause this test to fail. Please use NODE_CONFIG_DIR=apps/api/config.`,
+      );
+    }
```
kgajowy, line 16: due to how `AppConfig` is implemented, even just importing it (which is pretty common in an indirect way, due to a lot of usage everywhere) makes it load the config from envs. Could be a good idea to move it to global setup for unit tests.

In the long run, I would suggest to replace the mechanism with https://docs.nestjs.com/techniques/configuration

Makefile
tiagojsag: these are copy-paste of the equivalent, non "native-" prefixed tasks above, but just run locally instead of inside a docker container.

api/apps/api/src/migrations/api/1608149578000-EnablePostgis.ts
tiagojsag: image

Makefile
kgajowy: @tiagojsag sure thing, I just used the opportunity to possibly describe them for future people, not to question its existence/modification

tiagojsag: Understanding what they do is a fight for another day

api/apps/api/src/migrations/api/1608149578000-EnablePostgis.ts
kgajowy: `extent` was dropped , as well as `tr_GetBbox` I believe 🤔

tiagojsag: ok, but that means they still exist at some point. Perhaps there is an improvement opportunity here, where this can be refactored further down the road.

api/apps/api/config/custom-environment-variables.json
tiagojsag: upon closer inspection on reverting my changes, I actually noticed that API_POSTGRES_URL is a sort of internal variable, as devs are always expected to provide the individual host/user/password/db values. So, in DRY spirit, I'm removing the URL var instead

api/apps/api/src/migrations/api/1608149578000-EnablePostgis.ts
hotzevzl: good point. I have created https://vizzuality.atlassian.net/browse/MARXAN-1165 to take care of this.

Makefile
```diff
@@ -1,21 +1,25 @@
...
     export
 endif
 
-CIENV := $(if $(filter $(environment), ci), -f docker-compose-test-e2e.ci.yml , -f docker-compose-test-e2e.local.yml)
-API_DB_INSTANCE := $(if $(environment), test-e2e-postgresql-api, postgresql-api)
-GEO_DB_INSTANCE := $(if $(environment), test-e2e-postgresql-geo-api, postgresql-geo-api)
-REDIS_INSTANCE := $(if $(environment), test-e2e-redis, redis)
+API_DB_INSTANCE := $(if $(filter $(IS_DOCKER_E2E_TESTS_RUNTIME), true), test-e2e-postgresql-api, postgresql-api)
+GEO_DB_INSTANCE := $(if $(filter $(IS_DOCKER_E2E_TESTS_RUNTIME), true), test-e2e-postgresql-geo-api, postgresql-geo-api)
+REDIS_INSTANCE := $(if $(filter $(IS_DOCKER_E2E_TESTS_RUNTIME), true), test-e2e-redis, redis)
```
hotzevzl, line 20: I like the way you added the ability to manage arbitrary environments and then adapted the previous hardcoded special treatment of two environments (ci and non-ci) here, but I am not clear (no wonder, it's make-land and it's late Friday afternoon) how this new bool var is supposed to be working: I don't have it defined locally, so IIUC the _else_ branch of these conditional assignments should be used, but then e2e tests still work fine, so I am losing the plot.

Irrespective of my puzzlement, I would suggest to add a comment (in the Makefile itself, maybe, for this all to be self-contained? or in the main `README.md` to go along all/most of the other env vars we may use in `.env` files) about how to use this var and when/where to set it.

api/apps/api/src/utils/config.utils.spec.ts
hotzevzl: agree, the Nest-y way would be handy as part of a future tech debt cleanup (and alignment with the Nest way wherever there's no need to go ad-hoc).

hotzevzl (commented):
@tiagojsag thanks for this (and your previous work on making native execution possible).

Besides the comments inline in the code review, there's something that we should address - I don't see how any of the changes in this PR may affect that, so it may be due to changes in the earlier PR (hence no strict request for changes here), but basically what is now happening is that when I run e2e tests locally through the docker compose setup (rather than simply running them within a running docker container for normal API use), the "normal" apidb and geodb volumes get wiped (via `clean-slate`, I guess).

I can try to figure out which change introduced this regression, but that will have to be for next week.

PR #704, opened 2021-12-17
feat(cloning): add export configuration

kgajowy:
As per https://vizzuality.atlassian.net/browse/MARXAN-1109 

This PR adds : 
*  "dummy" pieces handlers
* fix on forever-pending job (`submitted` event instead `finished`)

Unpacked export:
image

### DIFF
--- a/api/apps/geoprocessing/src/export/pieces-exporters/project-config.ts
+++ b/api/apps/geoprocessing/src/export/pieces-exporters/project-config.ts
@@ -32,6 +32,8 @@ export class ProjectConfig extends PieceProcessor {
       throw new Error(`Exporting scenario is not yet supported.`);
     }
 
+    await delay();
+
     const scenarios: { name: string }[] = await this.entityManager.query(
       `
        SELECT name FROM scenarios where project_id = $1
@@ -68,3 +70,8 @@ export class ProjectConfig extends PieceProcessor {
     };
   }
 }
+
+const delay = () =>
+  new Promise((resolve) => {
+    setTimeout(resolve, 1000);
+  });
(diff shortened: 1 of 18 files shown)


api/apps/geoprocessing/src/export/pieces-exporters/project-config.ts
```diff
@@ -0,0 +1,77 @@
...
+    await delay();
+
+    const scenarios: { name: string }[] = await this.entityManager.query(
+      `
+       SELECT name FROM scenarios where project_id = $1
+    `,
+      [input.resourceId],
+    );
+
+    const metadata = JSON.stringify({
```
tiagojsag, line 44: is this meant to mock the project metadata coming from the db?

tiagojsag (commented):
I have a couple of question, that are probably not blockers, must just me being new here. feel free to merge if you think it's ok as it is

api/apps/geoprocessing/src/export/pieces-exporters/project-config.ts
```diff
@@ -0,0 +1,77 @@
...
+  isSupported(piece: ClonePiece): boolean {
+    return piece === ClonePiece.ExportConfig;
+  }
+
+  async run(input: JobInput): Promise<JobOutput> {
+    if (input.resourceKind === ResourceKind.Scenario) {
+      throw new Error(`Exporting scenario is not yet supported.`);
+    }
+
+    await delay();
```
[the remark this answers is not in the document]
kgajowy: https://github.com/Vizzuality/marxan-cloud/pull/706/files#diff-21d5c8cc188cdfa256ed298ce1d63d6f66da1968b0ac9ae0e30902f84ca1912cR48-R49

its described in tasks mentioned in takeover plan. 

You are right it is odd - and in final implementation it shouldn't be there at all.

api/apps/geoprocessing/src/export/pieces-exporters/project-config.ts
kgajowy: it doesn't mock, it should actually be a valid implementation of what's needed for config, according to initial assumptions/HLD.

PR #707, opened 2021-12-20
Improve native execution - MVP

tiagojsag:
Subset of https://github.com/Vizzuality/marxan-cloud/pull/702 that should allow basic native execution, without the extensive refactor that the makefile could use


kgajowy (commented):
✅ LGTM to me but please wait for @hotzevzl 's approve (or at least, revisit later)

PR #712, opened 2021-12-21
docs(code-guide): error handling

kgajowy:
Please start with `README.md` before code - hopefully it will navigate you smoothly.

Cheers!

### DIFF
--- a/api/libs/code-guide/src/error-handling/01-entry-point/base.module.ts
+++ b/api/libs/code-guide/src/error-handling/01-entry-point/base.module.ts
@@ -1,0 +1,12 @@
+import { Module } from '@nestjs/common';
+import { LegacyService } from './legacy-service';
+import { ThirdPartyService } from './third-party-service';
+import { Facade } from './facade';
+import { SomeController } from './controller';
+
+@Module({
+  providers: [LegacyService, ThirdPartyService, Facade],
+  controllers: [SomeController],
+  exports: [Facade],
+})
+export class BaseModule {}
--- a/api/libs/code-guide/src/error-handling/01-entry-point/controller.ts
+++ b/api/libs/code-guide/src/error-handling/01-entry-point/controller.ts
@@ -1,0 +1,18 @@
+import { BadRequestException, Controller, Post } from '@nestjs/common';
+import { Facade } from './facade';
+
+@Controller()
+export class SomeController {
+  constructor(private readonly facade: Facade) {}
+
+  @Post()
+  async doThings() {
+    const result = await this.facade.doComposedAction();
+
+    if (result) {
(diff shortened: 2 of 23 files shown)


AaronPerezPerez (approved):

tiagojsag (commented):
Thanks for this, I found it very clear and useful. 

One concern I have is that this sort of approach effectively adds (instead of replacing) another layer of complexity. As you mention, we still need to account for Errors, because deps will use those, and because node natively uses those. So, in practice, we are adding a simpler way of handling errors, but required to support both. In light of this, do you still feel like the advantages of this approach outweigh the cons of having to support both scenarios? Wouldn't it be better to use "the devil we know" and have a more robust Error handling mechanism?

kgajowy:
> Thanks for this, I found it very clear and useful.
> 
> One concern I have is that this sort of approach effectively adds (instead of replacing) another layer of complexity. As you mention, we still need to account for Errors, because deps will use those, and because node natively uses those. So, in practice, we are adding a simpler way of handling errors, but required to support both. In light of this, do you still feel like the advantages of this approach outweigh the cons of having to support both scenarios? Wouldn't it be better to use "the devil we know" and have a more robust Error handling mechanism?

As state in doc, it doesn't *always* make sense - especially in services internal to module.

 > we still need to account for Error
doc stated we may not (if it wasn't expected)

> another layer of complexity
I wouldn't call it another `layer of complexity` - it may be just (or as much as) a way to understand possible consequences of failure and what they mean - which address the issue of not being able to predict potential failures for consumers as well as not even considering handling them at all.

> do you still feel like the advantages of this approach outweigh the cons
In many cases - yes. As above, it isn't always the right and the only way. I would at least recommend using such for public API of given module and where it makes sense to (for example, timeout/error to external payment service, maybe insert to db when some unique constraints are in place and so on).

So the rule of thumb would be - if a consumer (other developers) of particular method/module would be grateful for understanding what can go wrong - use it. Same goes for important internals that different errors may cause different compensate actions (or require some meaningful message to consumer).

> “Indeed, the ratio of time spent reading versus writing is well over 10 to 1. We are constantly reading old code as part of the effort to write new code. ...[Therefore,] making it easy to read makes it easier to write.”
― Robert C. Martin

I fully agree with that statement deep inside my heart - writing more code but making it more readable is worth it.

tiagojsag (approved):

hotzevzl (approved):
great writeup and step-by-step examples, thank you! definitely worth to be posted on your own engineering blog 🙏🏼 

I found this to be a nice complement: https://medium.com/inato/expressive-error-handling-in-typescript-and-benefits-for-domain-driven-design-70726e061c86

PR #714, opened 2021-12-22
MARXAN1079 - Adds findUsersInScenario method and endpoint

rubvalave:
## Implement endpoint to list users with roles assigned to a given scenario


api/apps/api/src/modules/access-control/access-control.module.ts
```diff
@@ -4,13 +4,14 @@ import { TypeOrmModule } from '@nestjs/typeorm';
...
-import { ScenarioAccessControl } from './scenarios-acl/scenario-access-control';
-import { ScenarioAclService } from './scenarios-acl/scenario-acl.service';
+import { ProjectAccessControl } from '@marxan-api/modules/access-control/projects-acl/project-access-control';
+import { ProjectAclService } from '@marxan-api/modules/access-control/projects-acl/project-acl.service';
+import { ProjectAclController } from '@marxan-api/modules/access-control/projects-acl/project-acl.controller';
+import { ProjectAclModule } from '@marxan-api/modules/access-control/projects-acl/project-acl.module';
+import { ScenarioAclModule } from '@marxan-api/modules/access-control/scenarios-acl/scenario-acl.module';
+import { ScenarioAccessControl } from '@marxan-api/modules/access-control/scenarios-acl/scenario-access-control';
+import { ScenarioAclService } from '@marxan-api/modules/access-control/scenarios-acl/scenario-acl.service';
+import { ScenarioAclController } from '@marxan-api/modules/access-control/scenarios-acl/scenario-acl.controller';
```
kgajowy, line 14: hmm, why? `project-acl` and `scenarios-`acl` are the submodules of `access-control` 🤔

api/apps/api/test/access-control/scenario-acl/scenario-acl-get.e2e-spec.ts
```diff
@@ -0,0 +1,59 @@
...
+afterEach(async () => {
+  await fixtures?.cleanup();
+});
+
+test(`getting scenario users when user has no role in the scenario`, async () => {
+  const scenarioId = await fixtures.GivenScenarioWasCreated();
+  const response = await fixtures.WhenGettingScenarioUsersAsNotInScenario(
+    scenarioId,
+  );
+  fixtures.ThenForbiddenIsReturned(response);
```
kgajowy, line 18: Consider creating reusable step
```
fixtures.WhenGettingScenarioUsers(scenarioId, userId)
```

api/apps/api/src/modules/access-control/access-control.module.ts
rubvalave: A couple of lines above it is using `@marxan-api` imports and I thought it should be used as well here.

However, giving it a deeper thought, I think the two lines above are the entitites and, at least one of them, used to be placed under the projects submodule before and maybe that's why it uses this.

Should I change all of them in this case to relative, right? (I'm asking but I will probably do it right away)

api/apps/api/test/access-control/scenario-acl/scenario-acl-get.e2e-spec.ts
rubvalave: The one that is reused in other tests of the other endpoints is `WhenGettingScenarioUsersAsOwner` as it is needed to check that a scenario has been updated with new users and also to check that the users you deleted are effectively gone.

I didn't want to use something generic like `WhenGettingScenarioUsers` because this action is defined by the user doing the action and a general one can take away some essential information that needs to be coupled together with the action itself.

api/apps/api/src/modules/access-control/access-control.module.ts
kgajowy: @rubvalave you can also consider using index.ts files for each submodule as well to think through what exactly given module should expose and why

api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.service.ts
```diff
@@ -104,4 +106,37 @@ export class ScenarioAclService implements ScenarioAccessControl {
     });
     return otherOwnersInScenario >= 1;
   }
+
+  async findUsersInScenario(
+    scenarioId: string,
+    userId: string,
+    nameSearch?: string,
+  ): Promise<Either<Permit, UserRoleInScenarioDto[]>> {
```
tiagojsag, line 114: the Permit/false thing I mentioned on other PRs

tiagojsag (changes requested):

tiagojsag (approved):

PR #715, opened 2021-12-22
Marxan1080 -  update scenario users endpoint and method

rubvalave:
## Implement endpoint to grant permissions to a user on a given scenario


api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.service.ts
```diff
@@ -104,4 +107,45 @@ export class ScenarioAclService implements ScenarioAccessControl {
...
+      if (!existingUserInProject) {
+        await this.roles.save({
+          scenarioId,
+          userId,
+          roleName,
+        });
+      } else {
+        await this.roles.update({ scenarioId, userId }, { roleName });
+      }
+    } catch (err) {
```
kgajowy, line 144: It won't work like that. Only the `apiQueryRunner` instance will be transactional, the `this.roles` repo is still "normal" mode.

Use https://docs.nestjs.com/techniques/database (EntityManager) - we also have a few examples of its usage within codebase.,

api/apps/api/test/access-control/scenario-acl/scenario-acl.fixtures.ts
```diff
@@ -281,6 +303,7 @@ export const getFixtures = async () => {
     ThenAllUsersinScenarioAfterEveryTypeOfUserHasBeenAddedAreReturned: (
       response: request.Response,
     ) => {
+      console.log(response);
```
kgajowy, line 306:
```suggestion
```

api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.controller.ts
```diff
@@ -0,0 +1,58 @@
...
+  async updateUserInProject(
+    @Body() dto: UserRoleInScenarioDto,
+    @Param('scenarioId', ParseUUIDPipe) scenarioId: string,
+    @Req() req: RequestWithAuthenticatedUser,
+  ): Promise<void | boolean> {
+    const result = await this.scenarioAclService.updateUserInProject(
+      scenarioId,
+      dto,
+      req.user.id,
+    );
```
kgajowy, line 52: it isn't fully clear to me - dto hides some things (also why would it be used by service, as it is REST thing?) and name states `in project` but there is `scenarioId` given.

```
this.acl.grant(scenarioId, role, userId)
```

api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.service.ts
rubvalave: I used the same method as in [`project-acl.service`](https://github.com/Vizzuality/marxan-cloud/blob/develop/api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts) so I guess both should be changed right?

kgajowy: @rubvalave in this case, yes

api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.controller.ts
rubvalave: We need both loggedUserId (to check ownership) and userId + roleName on the other side to change the given user into the role selected, that's why I send the dto directly to the service. Wouldn't it be more confusing if I got the things out of the DTO?

rubvalave: Maybe changing the name of it into something like `UserAndRoleToChange` or whatever will be better option here.

tiagojsag (changes requested):
- the linter is complaining about code style, as well as unused imports on files you edited. Please make sure your development environment is configured to notify you of these types of issues.
- Your tests don't really validate the business logic, they just validate the response code. I checked out your branch, commented lines 117-158 on the service, and ran the tests, which all passed.

api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.controller.ts
```diff
@@ -0,0 +1,58 @@
...
+} from '@nestjs/swagger';
+import { isLeft } from 'fp-ts/lib/These';
+import {
+  UserRoleInScenarioDto,
+} from '@marxan-api/modules/access-control/scenarios-acl/dto/user-role-scenario.dto';
+
+@UseGuards(JwtAuthGuard)
+@ApiBearerAuth()
+@ApiTags('Scenarios-Users Roles')
+@Controller(`${apiGlobalPrefixes.v1}/roles/scenarios`)
```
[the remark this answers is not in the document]
rubvalave: But that is the root url for the endpoints under this module on Swagger, the url in issue description is in line 36. `  @Patch(':scenarioId/users')
`

Also tests will validate the logic (added new users with roles and changed roles) when PR 714 is merged as the only way to check if users are or not in a project was with the endpoint added in there (that's why the comment I added for this PR on the tests).

Maybe it was not a good idea to cut into pieces the work to be done in the controller/service because it ends up with cases like this where something needs a previous piece of work to run "effectively"

api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.service.ts
```diff
@@ -104,4 +108,54 @@ export class ScenarioAclService implements ScenarioAccessControl {
...
+        await apiQueryRunner.manager.save(userRoleToSave);
+      } else {
+        await apiQueryRunner.manager.update(
+          UsersScenariosApiEntity,
+          { scenarioId, userId },
+          { roleName },
+        );
+      }
+    } catch (err) {
+      await apiQueryRunner.rollbackTransaction();
```
tiagojsag, line 155: so if one of the queries above fails, an exception is thrown, but you still return right(), which seems off to me. Shouldn't we inform the user that something failed?

api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.service.ts
```diff
@@ -104,4 +108,54 @@ export class ScenarioAclService implements ScenarioAccessControl {
     });
     return otherOwnersInScenario >= 1;
   }
+
+  async updateUserInScenario(
+    scenarioId: string,
+    userAndRoleToChange: UserRoleInScenarioDto,
+    loggedUserId: string,
+  ): Promise<Either<Permit, void>> {
```
tiagojsag, line 116: left has Permit type, but you only use it with false, so maybe some stricter typing may be used

tiagojsag (changes requested):

api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.service.ts
rubvalave: After fighting with debugging in VSCode for the try/catch/finally I think now it handles better this part, both by informing the user if an error occurs and returning `right()` in the right place.

PR #716, opened 2021-12-22
Marxan1017 - revoke access to scenarios endpoint and method

rubvalave:
## Implement endpoint to revoke permissions of a user in a given scenario


api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.controller.ts
```diff
@@ -0,0 +1,54 @@
...
+  })
+  async deleteUserFromProject(
+    @Param('scenarioId', ParseUUIDPipe) scenarioId: string,
+    @Param('userId', ParseUUIDPipe) userId: string,
+    @Req() req: RequestWithAuthenticatedUser,
+  ): Promise<void | boolean> {
+    const result = await this.scenarioAclService.revokeAccess(
+      scenarioId,
+      userId,
+      req.user.id,
```
kgajowy, line 47: user id twice may be misleading, consider slight rename to diff from requester/subject

api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.service.ts
```diff
@@ -104,4 +105,21 @@ export class ScenarioAclService implements ScenarioAccessControl {
...
+    scenarioId: string,
+    userId: string,
+    loggedUserId: string,
+  ): Promise<Either<Permit, void>> {
+    if (!(await this.isOwner(loggedUserId, scenarioId))) {
+      return left(false);
+    }
+
+    if (!(await this.hasOtherOwner(userId, scenarioId))) {
+      return left(false);
```
tiagojsag, line 119: IMO it would be important to differentiate between the two scenarios when "Forbidden" may be returned. Specifically, I think we should return a meaningful message to the user in case they are the sole owner of a project.

tiagojsag (changes requested):

api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.service.ts
```diff
@@ -104,4 +105,21 @@ export class ScenarioAclService implements ScenarioAccessControl {
     });
     return otherOwnersInScenario >= 1;
   }
+
+  async revokeAccess(
+    scenarioId: string,
+    userId: string,
+    loggedUserId: string,
+  ): Promise<Either<Permit, void>> {
```
tiagojsag, line 113: Permit has both true or false, but your left() calls only use false

tiagojsag (changes requested):

api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.service.ts
rubvalave: Differentiated now (and changed slightly how it is handled).

Note -> The method used needs a refactor into a common utils function because the same structure is used (and will be used) plenty of times for different entities of the ACL module.

PR #718, opened 2021-12-23
Home hero/intro section

SARodrigues:
### Overview

This is the intro/hero for the new homepage. 

Key changes:  
- **Hexagon images + background animation**   
  Given some peculiarities (different opacity/blurring, zoom levels) between the different hexagons, the animations are achieved using [Animation Controls](https://www.framer.com/docs/animation/#component-animation-controls). The declarative nature of the project's settings led to quite a few `useAnimation` initialisations, though that shouldn't be an issue. The sequences are achieved without having to re-render the component ever, which should help with performance (I hope).  

  In order to allow for changes/tweaks to be made easily, each image's settings per "slide" are declared in a `constants.ts`, down to the positioning of each hexagon, how much it'll scale, how much it'll blur, etc. 

  Images are originally squares, turned into hexagons using a [clipPath](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/clipPath). The borders around the images are SVGs animated separately. 
- **Optimised/tweaked images**  
  The background images were slightly tweaked. Given that they may (and will likely not) fill the width of the screen, a CSS background has to be set and animated along with the background image, to take up the rest of the space. Due to the images' backgrounds being noisy and CSS colors are obviously solid, this was causing a visible "break/bar" in the hero background, when using a high screen brightness.

  The edges of the images were smoothed to prevent such a visible transition between image and background, then the images were ran through [ImageOptim](https://imageoptim.com/mac) to get their file size back down (and they ended up smaller than the originals, fortunately).   
- **Basic transparent background navbar** (background color changes during the animation)  
  The navigation is positioned with `absolute`, so that we wouldn't require layout changes. We may want to
  change this in the future, in case we add more pages with such heros/intros which overlap the navbar  

- **Mobile experience**  
  The design was made responsive and should adjust very nicely to a whole array of mobile screen sizes. The background will still be animated on mobile, but due to space constraints the hexagonal images won't be displayed. This should, however, help with mobile performance.  

### Designs

[MARXAN-1134](https://vizzuality.atlassian.net/browse/MARXAN-1134)

### Feature relevant tickets

[MARXAN-1134](https://vizzuality.atlassian.net/browse/MARXAN-1134)


SARodrigues:
Now that I think about it, I think the hexagon images should be cropped into squares and their resolution reduced. It'll bring their file size down quite a lot, without any visual degradation. 

Also, maybe _"slides"_ would have been a better word than _"scenes"_, and _"HomeHero"_ should be renamed to _"HomeIntro"_.

anamontiaga (approved):

anamontiaga (approved):
It has fit perfect.
I will merged.
Lets fixed naming issue and image size issue on 1132.

PR #719, opened 2021-12-23
Bug/marxan 1152 change unique project role

rubvalave:
## Fix for users roles in project-acl

- Fix unique owner being able of changing their role and effectively locking out the project.
- Fix `project_contributor` and `project_viewer` roles not seeing their projects in the general view.
- Add test to check that the unique owner cannot change their role and locking out the project.


anamontiaga (approved):

tiagojsag:
Insta-merge </3 see also https://github.com/Vizzuality/marxan-cloud/pull/716/files#r775549716

PR #728, opened 2022-01-03
Limiting uploads file size (frontend) + friendlier file size error me…

SARodrigues:
This PR limits (updates the limits) of the uploads size. 

### Overview

- An `util` file was added to help with units conversion, currently only `bytesToMegabytes`  
- A `constants` file was added, where all uploader limits are set together, making it easier to change them in the future   
- Frontend error messages relative to the max upload size being exceeded, currently returned by `react-dropzone`, were tweaked.  
    Unfortunately `react-dropzone` returns something like _"File is larger than 1000000 bytes"_, which is not very intuitive both because the user may not be familiar with what a `byte` is, but also because our recommended file size is displayed in `Megabytes`. 
    Now the error message will be _"File is larger than 10 MB"_. 
- Specific error messages returned by the backend will now be displayed in the "toast", instead of just a _"(...) could not be uploaded"_   

Updated limits:  
- Planning units: 1MB  
- Planning areas: 1MB   
- Planning areas with grid: 10MB  
- Protected areas: 10MB  
- Features: 20MB  
- Cost Surface: 20MB  

### Testing instructions

Verify that the respective uploaders display the correct limits, and that when attempting to upload a file which exceeds the limits, an error message is displayed saying _"File is larger than X MB."_  

Increase the size limits in `constants/file-uploader-size-limits.js` to bypass the file size frontend validation and trigger a backend error, and verify that the resulting toast error message displays the backend error message. 

Files with arbitrary sizes to test can be generated with: https://pinetools.com/random-file-generator

### Feature relevant tickets

[MARXAN-1065](https://vizzuality.atlassian.net/browse/MARXAN-1065)
[MARXAN-1170](https://vizzuality.atlassian.net/browse/MARXAN-1170)


app/layout/projects/new/form/planning-area-grid-uploader/component.tsx
```diff
@@ -95,7 +91,14 @@ export const PlanningAreaGridUploader: React.FC<PlanningAreaGridUploaderProps> =
 
   const onDropRejected = (rejectedFiles) => {
     const r = rejectedFiles[0];
-    const { errors } = r;
+
+    // `file-too-large` backend error message is not friendly.
```
SARodrigues, line 95: Not really a fan of repeating this bit of code in all uploaders, but given the customisation of the message (custom max size) I'm not sure how to go about it. It feels too specific to create an util for it, but could be done. What do you think?

app/constants/file-uploader-size-limits.js
```diff
@@ -0,0 +1,6 @@
+export const PLANNING_UNIT_UPLOADER_MAX_SIZE = 1e6; // 1MB
+export const PLANNING_AREA_UPLOADER_MAX_SIZE = 1e6; // 1MB
+export const PLANNING_AREA_GRID_UPLOADER_MAX_SIZE = 1e7; // 10MB
+export const PROTECTED_AREA_UPLOADER_MAX_SIZE = 1e7; // 10MB
+export const FEATURES_UPLOADER_MAX_SIZE = 2e7; // 20MB
+export const COST_SURFACE_UPLOADER_MAX_SIZE = 2e7; // 20MB
```
hotzevzl, line 6: @SARodrigues I suggest to use MiBs rather than MBs here. I know it won't matter in most circumstances, and the spec of defaults in M-1032 wasn't fully clear about the specific measurement unit, but we do actually default to multiple of MiBs in the API implementation (https://github.com/Vizzuality/marxan-cloud/blob/develop/api/apps/api/src/modules/uploads/upload-limits.ts) - and I'd say we should actually limit the configurability of these limits to MiBs in the API too, if sticking to this works for you. It should simplify how we handle these limits, as I definitely don't expect anything sub-MiB (or fractions of MiB) to ever be needed nor make sense.

app/layout/projects/new/form/planning-area-grid-uploader/component.tsx
anamontiaga: I think the best thing would be to have an upload component for everything, but I don't know how agnostic we would get to do it since the upload logic is not the same for all cases. 

We can look together if it is feasible or a priority to do so.

anamontiaga (approved):

app/constants/file-uploader-size-limits.js
SARodrigues: @hotzevzl Updated to use MiBs! I added a note to the respective frontend file with a link to the backend counterpart, to prevent any confusion in the future. :)

PR #731, opened 2022-01-04
Marxan-1168-acl-transactions-not-commiting-properly

rubvalave:


api/apps/api/test/access-control/project-acl/project-acl-patch.e2e-spec.ts
```diff
@@ -92,3 +92,57 @@ test(`adds a not allowed user role to project`, async () => {
...
+
+test(`adds non-existant userId`, async () => {
+  const projectId = await fixtures.GivenProjectWasCreated();
+  const nonExistantUserIdResponse = await fixtures.WhenAddingNonExistantUserId(
+    projectId,
+  );
+  fixtures.ThenTransactionFailedIsReturned(nonExistantUserIdResponse);
+});
+
+test(`adds and deletes users alternatively`, async () => {
```
tiagojsag, line 112: there's a lot going on in this test. I know why it exists, but me-in-6-months won't remember. Typically, this sort of regression tests has an issue id added as comment somewhere, so we can have a bit of background on why it exists, and what it covers. Since we don't have an issue for this particular problem, I would suggest you add some comments to explain why this test exists, and what problem it covers/prevents, so us-in-6-months (or someone who's new to the project) can understand why such a complex test exists.

tiagojsag (changes requested):
functionally looks ok. however there are a bunch of typos that could have easily been avoided. please make sure you add a spellchecker to your IDE. there's also a couple more small details to address, everything minor

api/apps/api/test/access-control/project-acl/project-acl-patch.e2e-spec.ts
hotzevzl: Tiago covered all the substantive recommendations here - I'd just suggest to also slightly improve the test description: firstly, I think we are trying to convey "alternately", not "alternatively" here; I'd also argue it's more of a granting and revoking back and forth (adding and deleting *from the project* is also correct, strictly speaking, but my first thought before going through the test content was that users were being created and deleted outright), but there's only so much we should try to convey in a short test description, so I would not stress about this too much.

I'd concur that a brief comment explaining the *why* (as comments should do) would help, as the function names tell us mostly *what* we're doing and expecting, not so much why we're doing all of this and what we want to have confidence in.

Besides these notes, well done, great work on these tests!

hotzevzl (approved):

tiagojsag (approved):

PR #732, opened 2022-01-05
Fix issues with BlmValuesCalculator

aciddaute:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/projects/blm/change-blm-range.handler.ts
+++ b/api/apps/api/src/modules/projects/blm/change-blm-range.handler.ts
@@ -10,14 +10,13 @@ import {
   invalidRange,
   updateFailure,
 } from './change-blm-range.command';
-import { SetProjectBlm } from './set-project-blm';
 import { PlanningUnitAreaFetcher } from './planning-unit-area-fetcher';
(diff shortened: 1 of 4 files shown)


tiagojsag (approved):

hotzevzl (approved):
thanks @aciddaute - all looks good.

Did you manage to reproduce the issue reported by Ana with negative values appearing in the set of BLM values? I would not block this PR if not, in any case: as is it should be already addressing at least the most obvious issues with the calculation of the BLM values set and merging this will help FE colleagues proceed with their implementation - we could take care of any leftover glitches, if any, in a separate PR.

hotzevzl:
> thanks @aciddaute - all looks good.
> 
> Did you manage to reproduce the issue reported by Ana with negative values appearing in the set of BLM values? I would not block this PR if not, in any case: as is it should be already addressing at least the most obvious issues with the calculation of the BLM values set and merging this will help FE colleagues proceed with their implementation - we could take care of any leftover glitches, if any, in a separate PR.

for reference, this was happening when the upper bound of the range was close enough to the min (e.g `[0,1]`) for the bug in the calculation formula to cause some values to be negative.

PR #733, opened 2022-01-05
Blm calibration partial and final results score property replaced by cost

aciddaute:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/blm/values/scenario-calibration-repo.ts
+++ b/api/apps/api/src/modules/blm/values/scenario-calibration-repo.ts
@@ -2,7 +2,7 @@ export interface CalibrationRunResult {
   id: string;
   scenarioId: string;
   blmValue: number;
-  score: number;
+  cost: number;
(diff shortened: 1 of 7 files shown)


tiagojsag (approved):

tiagojsag (commented):
seems like we have a test blind spot, but that's an issue for another day

hotzevzl (approved):

hotzevzl:
> seems like we have a test blind spot, but that's an issue for another day

@tiagojsag here there was firstly a blind spot in my brain apparently, as I ranted about using `cost` in the HLD document but then specified a wrong payload with `score` instead, which is what was then used in the implementation.

Nevertheless, I think we're not checking that the result has a specific shape (we're checking its length and other aspects).

aciddaute:
I'm adding an e2e test that verifies the shape of the results retrieved via API :+1:

PR #736, opened 2022-01-07
Test/blm calibration results endpoint

aciddaute:
This PR adds an e2e test for ensuring that API contract is met for blm calibration results endpoint

### DIFF
--- a/api/apps/api/test/scenarios/get-scenario-blm-calibration-results.e2e-spec.ts
+++ b/api/apps/api/test/scenarios/get-scenario-blm-calibration-results.e2e-spec.ts
@@ -1,0 +1,22 @@
+import { FixtureType } from '@marxan/utils/tests/fixture-type';
+import { getFixtures } from './get-scenario-blm-calibration-results.fixtures';
+
+let fixtures: FixtureType<typeof getFixtures>;
+
+describe('get-scenario-blm-calibration-results', () => {
+  beforeEach(async () => {
+    fixtures = await getFixtures();
+  }, 100000);
+
+  afterEach(async () => {
+    await fixtures?.cleanup();
+  });
+
+  it(`retrieves blm calibration results via API`, async () => {
+    await fixtures.GivenScenarioWasCreated();
+
+    await fixtures.WhenBlmCalibrationIsLaunched();
+
+    await fixtures.ThenCalibrationResultsShouldBeAvailableViaAPI();
+  }, 100000);
+});
--- a/api/apps/api/test/scenarios/get-scenario-blm-calibration-results.e2e-spec.ts
+++ b/api/apps/api/test/scenarios/get-scenario-blm-calibration-results.e2e-spec.ts
@@ -17,6 +17,6 @@ describe('get-scenario-blm-calibration-results', () => {
 
     await fixtures.WhenBlmCalibrationIsLaunched();
 
-    await fixtures.ThenCalibrationResultsShouldBeAvailableViaAPI();
+    await fixtures.ThenCalibrationResultsShouldBeAvailable();
   }, 100000);
 });
--- a/api/apps/api/test/scenarios/get-scenario-blm-calibration-results.fixtures.ts
+++ b/api/apps/api/test/scenarios/get-scenario-blm-calibration-results.fixtures.ts
@@ -1,0 +1,81 @@
+import { ScenarioType } from '@marxan-api/modules/scenarios/scenario.api.entity';
+import { HttpStatus } from '@nestjs/common';
+import * as request from 'supertest';
+import { BlmCalibrationRunResultDto } from '../../src/modules/scenarios/dto/scenario-blm-calibration-results.dto';
+import { GivenProjectExists } from '../steps/given-project';
+import { GivenUserIsLoggedIn } from '../steps/given-user-is-logged-in';
+import { bootstrapApplication } from '../utils/api-application';
+import { OrganizationsTestUtils } from '../utils/organizations.test.utils';
+import { ProjectsTestUtils } from '../utils/projects.test.utils';
+import { ScenariosTestUtils } from '../utils/scenarios.test.utils';
+
+export const getFixtures = async () => {
+  const app = await bootstrapApplication();
+  const token = await GivenUserIsLoggedIn(app);
+
+  const { projectId, organizationId } = await GivenProjectExists(
+    app,
+    token,
+    {
+      countryCode: 'AGO',
+      name: `Project name ${Date.now()}`,
+    },
+    {
+      name: `Org name ${Date.now()}`,
+    },
+  );
+
+  await ProjectsTestUtils.generateBlmValues(app, projectId);
+  let scenarioId: string;
+  const blmRange = [0, 100];
+
+  return {
+    cleanup: async () => {
+      await ProjectsTestUtils.deleteProject(app, token, projectId);
+      await ScenariosTestUtils.deleteScenario(app, token, scenarioId);
+      await OrganizationsTestUtils.deleteOrganization(
+        app,
+        token,
+        organizationId,
+      );
+      await app.close();
+    },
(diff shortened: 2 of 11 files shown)


api/apps/api/test/scenarios/get-scenario-blm-calibration-results.e2e-spec.ts
```diff
@@ -0,0 +1,22 @@
...
+  afterEach(async () => {
+    await fixtures?.cleanup();
+  });
+
+  it(`retrieves blm calibration results via API`, async () => {
+    await fixtures.GivenScenarioWasCreated();
+
+    await fixtures.WhenBlmCalibrationIsLaunched();
+
+    await fixtures.ThenCalibrationResultsShouldBeAvailableViaAPI();
```
hotzevzl, line 20: only cosmetic: I'd say the `ViaAPI` part seems redundant here

api/apps/api/test/utils/api-application.ts
```diff
@@ -30,6 +32,8 @@ export const bootstrapApplication = async (
     .useValue(fakeProjectChecker)
     .overrideProvider(FileRepository)
     .useClass(TempStorageRepository)
+    .overrideProvider(ScenarioCalibrationRepo)
+    .useClass(FakeScenarioCalibrationRepo)
```
hotzevzl, line 36: using a fake test double here helps us to work around the variability of time required to run an actual calibration, however, aren't we ending up confirming every time we run this test suite that the data we set in the fake repo is indeed correct in shape?

I would suggest to either accept the liability of using the real repo, wait for the calibration to finish and so on, or at least extend the result generation backwards to the point where we read data from a Marxan output file, in which case we'd at least test that the way we parse the output files and generate result payloads from there is valid and doesn't break in the future.

Probably the "using the full calibration workflow" option may be the easiest here, but I don't have strong opinions. If doing so, I'd say a long-enough sleep could be ok as a first step, but in general we may want to implement a proper await via polling of status, similarly to what we do in bots or the frontend does, but this would be definitely overkill for the moment, even though it's only a few dozens lines of code.

api/apps/api/test/scenarios/get-scenario-blm-calibration-results.fixtures.ts
```diff
@@ -0,0 +1,81 @@
...
+        app.getHttpServer(),
+      )
+        .get(`/api/v1/scenarios/${scenarioId}/calibration`)
+        .set('Authorization', `Bearer ${token}`)
+        .expect(200);
+
+      const results = response.body;
+
+      expect(results.length).toBeGreaterThan(0);
+
```
hotzevzl, line 70: another thing we may want to check here is that the blm values follow a monotonic sequence.

This would actually be a test that checks the BLM value calculator results are correct (given the `[min, max]` bounds, the six generated values are on a monotonic sequence), so maybe actually best to ignore me on this one 😅

Likewise, in practice that `result.blmValue` is within the bounds of `blmRange` would be something to include in tests for the BLM values set calculator, but they can stay here too if they are not testing for this elsewhere.

hotzevzl (commented):
thanks @aciddaute!

minor/cosmetic suggestions, plus a more substantive one (see inline) about the usefulness of using a test double for the repo - I may be missing something though!

api/apps/api/test/scenarios/get-scenario-blm-calibration-results.e2e-spec.ts
aciddaute: Yes, I'll remove that part :+1:

api/apps/api/test/utils/api-application.ts
aciddaute: We tried to implement the test in the way that you point out (waiting until the blm calibration process finishes) but we realized that bullmq queues implementation is mocked, meaning that geoprocessing and api services cannot communicate between them 🤔

aciddaute: Since blm calibration process starts after adding a job to the corresponding queue, we didn't manage to run the calibration in the e2e test

api/apps/api/test/scenarios/get-scenario-blm-calibration-results.fixtures.ts
aciddaute: I think that tests in `api/apps/api/src/modules/projects/blm/domain/blm-values-calculator.spec.ts` ([link](https://github.com/Vizzuality/marxan-cloud/blob/9ecda96ef097a80a8ceb812826cbe60d1a048b26/api/apps/api/src/modules/projects/blm/domain/blm-values-calculator.spec.ts)) already ensures that resulting blm values are between the given range and follow a monotonic sequence

api/apps/api/test/utils/api-application.ts
aciddaute: Maybe we could 'invoke' in some way geoprocessing service in order to manually start the calibration process. I'll try it :+1:

hotzevzl: ah yes, you are right.

actually - looking at the relevant code more carefully now (sorry, I should have done that earlier), I think that given that what we send in the response is pretty much what we get out of the repository, I would say that maybe all we need is to keep your current test implementation, and just add explicit typing to the fake data we use for the fake repository, what do you think? I'll suggest the relevant change inline separately.

I feel that although we could plug back into the calibration process, this would require more effort than useful for the value that we're trying to achieve here.

api/apps/api/test/utils/scenario-calibration-repo.test.utils.ts
```diff
@@ -0,0 +1,60 @@
+import {
+  CalibrationRunResult,
+  ScenarioCalibrationRepo,
+} from '../../src/modules/blm/values/scenario-calibration-repo';
+
+export class FakeScenarioCalibrationRepo implements ScenarioCalibrationRepo {
+  fakeCalibrationResults = [
```
hotzevzl, line 7: this is the change I mentioned in my earlier comment a few moments ago:

```suggestion
  fakeCalibrationResults: CalibrationRunResult[] = [
```
(this would not be enough - we'd need to either hardcode or generate a UUID for `scenarioId` in the mock data below - I'd say it would be functionally identical, so up to you -- if you want to go this route, in any case)

api/apps/api/test/scenarios/get-scenario-blm-calibration-results.fixtures.ts
hotzevzl: yes, that's all we need, indeed 👍🏼

api/apps/api/test/utils/api-application.ts
aciddaute: It makes sense :+1:

api/apps/api/test/utils/scenario-calibration-repo.test.utils.ts
hotzevzl: yes, this works, thanks!

hotzevzl (approved):

api/apps/geoprocessing/src/marxan-sandboxed-runner/adapters-blm/blm-partial-results.geo.entity.ts
```diff
@@ -35,16 +35,16 @@ export class BlmPartialResultEntity {
   blmValue!: number;
 
   /**
-   * Score of the run
+   * Cost output parameter
    */
   @Column({
-    name: `score`,
+    name: `cost`,
     type: 'float',
```
kgajowy, line 42: REST Api results should rather be DB-agnostic. If needed, such mapping should be done on controller level (serializer) or could even be done at `sql` level using `as`

tiagojsag: not clear what in here is postgres specific.

PR #738, opened 2022-01-07
avoid index size limit when uploading custom planning unit grids from shapefile [MARXAN-1031]

hotzevzl:
https://vizzuality.atlassian.net/browse/MARXAN-1031

Please see shapefile attached to the Jira task for validating the functionality of this PR.

In practice, any PU grid shapefile with non-regularly-shaped (e.g. square, hexagon) PUs should allow to reproduce the issue without this PR, and should be processed correctly (if within other criteria such as maximum file size, valid geometries, etc.) with this PR.

### DIFF
--- a/api/apps/api/src/utils/postgresql.utils.ts
+++ b/api/apps/api/src/utils/postgresql.utils.ts
@@ -3,7 +3,7 @@ import { getManager } from 'typeorm';
 /**
  * Utility functions related to lower-level interaction with PostgreSQL servers.
  *
- * @debt This should be moved to a self-standing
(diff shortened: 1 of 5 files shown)


hotzevzl:
@aagm on second thought (having refreshed my postgresql index "wisdom" over the weekend), I suppose I could vastly simplify all this with a `hash` type index... what do you think?

In practice all we need is an equality check, which is what the `hash` index should give us, though I am not clear about the implications given that this was and is going to be an index over 3 columns.

hotzevzl:
merging after cross-checking with Agnieszka and as discussed with Alicia - thank you for the input!

PR #740, opened 2022-01-10
Adding scenario status filters on the frontend

SARodrigues:
### Overview

This PR adds Scenario (`run`) job status filtering.  
Filtering via API is not yet available, so the filtering is done on the frontend.   

**Notes:**  
- Filtering is currently done in the respective hook, so that we have it available seamlessly throughout the app   
  - The commented out modal code was kept virtually unchanged  
- The filters (only the `status` one) are not sent to the backend, as it'd return no results  
- Each scenario's run status is currently pulled from `api/v1/projects/{$projectId}/scenarios/status`, `run` job  
- The original property `status` returned by the backend is not being used, nor overwritten as one would be inclined to do. Couple reasons for this:  
    - Prevent conflicts when the backend implements that property  
    - The `ScenariosList` deals with all sorts of job statuses (not just the `run` one), this way it should be clear that it's not the same one sent by the backend, plus that way there was no need to remain a bunch of variables in that component  
- I have noticed that the backend can return 1 of 4 different statuses: `created`, `running`, `done`, `failure`. I've added filters for all of them. (`created` are scenarios that have been created but were never `run`)  
- Filters/Search/Sorting will be cleared when switching probjects  

**Other notes:**  
- I took the liberty to fix some typos in the code (mostly error messages, they wouldn't ever been seen by an user)  
- I also took the liberty of reorganising/refactoring the Scenarios#show component a little, because I have noticed that:      
   - Two consecutive API calls were being made to the same endpoint, with with `search`/`filters` and the other without (excluding `projectId`).  This would cause double the rendering as well.  
   - Both sets of returning data were a bit confusing, and not needed. If I had to take an educated guess, this would be from an initial stage of the project; possibly when searching was done in the frontend (which is no longer the case)  
    It's not perfect, but I mainly wanted to address the double-calls / double-rendering  

### Testing instructions

1. Create an empty project, with  no scenarios. Verify that:    
    - The pretty screen to create a new scenario is show  
    - The toolbar is not shown (search, filters)  
    - The create scenario button at the very bottom is not shown  
2. Create a few scenarios, ensure that:  
    - Search still works correctly  
    - Filters work correctly  
        - It may be needed to mock `job` statuses returned by the API, as it'll be hard to catch scenarios with a `running` and a `failure` statuses.   
     - Search + Filters together work correctly  
     - When there are no results, the appropriate message is shown  
     - The _"Clear all"_ button in the filters model clears all filters   
     - That when navigating to a new project, filters/search will be cleared  
       (it becomes a problem specially when switching to a project with no scenarios)

### Feature relevant tickets

[MARXAN-685](https://vizzuality.atlassian.net/browse/MARXAN-685)

### Screenshots  
Screenshot 2022-01-10 at 15 11 53

Screenshot 2022-01-10 at 15 11 39

### DIFF
--- a/app/components/scenarios/item/component.tsx
+++ b/app/components/scenarios/item/component.tsx
@@ -63,6 +63,7 @@ export interface ItemProps {
   progress?: number;
   lastUpdate: string;
   jobs?: Record<string, any>[];
+  runStatus: 'created' | 'running' | 'done' | 'failure',
   lastUpdateDistance: string;
   className?: string;
   onEdit: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void;
@@ -79,7 +80,8 @@ export const Item: React.FC<ItemProps> = ({
   progress,
   lastUpdateDistance,
   className,
-  jobs = [],
+  jobs,
+  runStatus,
   onEdit,
   onView,
   onCancelRun,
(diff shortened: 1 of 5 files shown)


anamontiaga:
When navigating to a new project, previous filters remains.

PR #742, opened 2022-01-10
Feat marxan 838 block when export is pending

AaronPerezPerez:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


api/apps/api/src/modules/scenarios/project-checker/project-checker.service-real.ts
```diff
@@ -8,19 +8,46 @@ import { Project } from '@marxan-api/modules/projects/project.api.entity';
...
+  doesntExist,
+  DoesntExist,
+  hasPendingExport,
+  HasPendingExport,
+  ProjectChecker,
+} from '@marxan-api/modules/scenarios/project-checker/project-checker.service';
 
 @Injectable()
-export class ProjectChecker {
+export class ProjectCheckerReal implements ProjectChecker {
```
hotzevzl, line 20: surely minor - I'd rather keep class names descriptive, and the `Real` suffix here is a bit of a last resort... and I know things can get a bit ugly when an abstract class is such not so much because of actual abstract-ness but to provide a way to use the same interface for test fake classes, but what about just

```suggestion
export class ProjectChecker implements ProjectValidator {
```
or
```suggestion
export class ProjectChecker implements ProjectCheckerRules {
```

or something like this? It's all quite arbitrary/opinionated, but I'd save the most descriptive and plain name for the "real" class.

api/apps/api/src/modules/scenarios/project-checker/project-checker.service-real.ts
```diff
@@ -8,19 +8,46 @@ import { Project } from '@marxan-api/modules/projects/project.api.entity';
...
   constructor(
     private readonly apiEvents: ApiEventsService,
     @InjectRepository(Project)
     private readonly repository: Repository<Project>,
     private readonly planningAreas: PlanningAreasService,
   ) {}
 
+  async hasProjectPendingExports(
```
hotzevzl, line 28: minor, I suggest to keep this to an actual question word order such as

```suggestion
  async projectHasPendingExports(
```
or even
```suggestion
  async hasPendingExports(
```

api/apps/api/src/modules/scenarios/project-checker/project-checker.service-real.ts
```diff
@@ -8,19 +8,46 @@ import { Project } from '@marxan-api/modules/projects/project.api.entity';
...
+        topic: projectId,
+        kind: In([
+          API_EVENT_KINDS.project__export__finished__v1__alpha,
+          API_EVENT_KINDS.project__export__failed__v1__alpha,
+          API_EVENT_KINDS.project__export__submitted__v1__alpha,
+        ]),
+      })
+      .catch(this.createNotFoundHandler());
+
+    const pendingExportExist =
```
hotzevzl, line 42:
```suggestion
    const pendingExportExists =
```

api/apps/api/src/modules/scenarios/scenarios.service.ts
```diff
@@ -548,4 +553,17 @@ export class ScenariosService {
...
+    // TODO: Check where do we should use the guard
+    const editIsBlocked = await this.projectChecker.hasProjectPendingExports(
+      scenario.projectId,
+    );
+
+    if (isLeft(editIsBlocked))
+      throw new BadRequestException(
+        `Scenario ${scenarioId} editing is blocked because of pending export`,
+      );
+  }
```
hotzevzl, line 568: this is fine for the time being - I think at some point we'll need to move it to its own module (maybe even a project library) if we want to reuse the same logic across all the spots in the codebase where we may want to reject starting a change if there is an export pending on the parent project.

api/apps/api/test/project/update-project.fixtures.ts
```diff
@@ -40,21 +47,39 @@ export const getFixtures = async () => {
         .patch(`/api/v1/projects/${projectId}`)
         .set('Authorization', `Bearer ${token}`)
         .send({
-          name: 'Test updated',
+          name: updatedName,
         }),
+    WhenProjectIsUpdatedWithARunningExport: async () => {
```
hotzevzl, line 52:
```suggestion
    WhenProjectIsUpdatedWhileAnExportIsPending: async () => {
```

api/apps/api/test/project/update-project.fixtures.ts
```diff
@@ -40,21 +47,39 @@ export const getFixtures = async () => {
...
         }),
     ThenWhenReadingProjectItHasNewData: async () => {
       const projectData = await request(app.getHttpServer())
         .get(`/api/v1/projects/${projectId}`)
         .set('Authorization', `Bearer ${token}`);
 
-      expect(projectData.body.data.attributes.name).toBe('Test updated');
+      expect(projectData.body.data.attributes.name).toBe(updatedName);
+    },
+    ThenWhenReadingProjectItHasTheOriginalData: async () => {
```
hotzevzl, line 77: for clarity, I suggest to split this into another action (e.g. `WhenGettingProjectData`) and an assertion (e.g. `ThenProjectIsUnchanged`.

At the very least, for clarity, I'd suggest:

```suggestion
    ThenProjectIsUnchanged: async () => {
```

api/apps/api/test/scenarios/start-scenario-calibration.e2e-spec.ts
```diff
@@ -30,6 +30,14 @@ describe('start-scenario-calibration', () => {
     await fixtures.ThenWhenReadingProjectCalibrationItHasTheNewRange();
   });
 
+  it(`throws an exception if an export is running`, async () => {
+    await fixtures.GivenScenarioWasCreated();
+
+    await fixtures
+      .ThenShouldFailWhenStartingAnScenarioCalibrationWithA()
```
hotzevzl, line 37: minor and from earlier changes, but I suggest to take this chance to fix the name of this fixture method

```suggestion
      .ThenShouldFailWhenStartingAScenarioCalibrationWithA()
```

api/apps/api/test/scenarios/update-scenario.e2e-spec.ts
```diff
@@ -0,0 +1,122 @@
...
+describe('update scenario', () => {
+  beforeEach(async () => {
+    fixtures = await getFixtures();
+  });
+
+  afterEach(async () => {
+    await fixtures?.cleanup();
+  });
+
+  it(`should update an scenario with new data`, async () => {
```
hotzevzl, line 25:
```suggestion
  it(`should update a scenario with new data`, async () => {
```

api/apps/api/test/scenarios/update-scenario.e2e-spec.ts
```diff
@@ -0,0 +1,122 @@
...
+    },
+    GivenScenarioWasCreated: async () => {
+      const result = await ScenariosTestUtils.createScenario(app, token, {
+        name: originalName,
+        type: ScenarioType.marxan,
+        projectId,
+      });
+      scenarioId = result.data.id;
+    },
+    WhenUpdatingNameAndDescriptionFromAnScenario: () => {
```
hotzevzl, line 87:
```suggestion
    WhenUpdatingNameAndDescriptionOfAScenario: () => {
```

api/apps/api/test/scenarios/update-scenario.e2e-spec.ts
```diff
@@ -0,0 +1,122 @@
...
+          projectChecker.addPendingExportForProject(projectId);
+          await request(app.getHttpServer())
+            .patch(`/api/v1/scenarios/${scenarioId}`)
+            .send({ name: updatedName, description: updatedDescription })
+            .set('Authorization', `Bearer ${token}`)
+            .expect(HttpStatus.BAD_REQUEST);
+        },
+      };
+    },
+    ThenWhenReadingTheScenarioDataItIsUpdated: async () => {
```
hotzevzl, line 106: as in previous comments, this could be split into another action (`WhenGettingScenarioData()`) and an assertion (`ThenScenarioDataWasUpdatedSuccessfully()`). At the very least, I'd recommend `ThenScenarioWasUpdatedSuccessfully()`, avoiding the `ThenWhen` part

api/apps/api/test/scenarios/update-scenario.e2e-spec.ts
```diff
@@ -0,0 +1,122 @@
...
+      const scenarioData = await request(app.getHttpServer())
+        .get(`/api/v1/scenarios/${scenarioId}`)
+        .set('Authorization', `Bearer ${token}`);
+      expect(scenarioData.body.data.attributes.name).toEqual(updatedName);
+      expect(scenarioData.body.data.attributes.description).toEqual(
+        updatedDescription,
+      );
+    },
+    ThenWhenReadingTheScenarioDataItIsNotUpdated: async () => {
```
hotzevzl, line 115: as above. Or at least something like `ThenSenarioIsUnchanged()`.

hotzevzl (commented):
thank you, and apologies for leaving this hanging for so long.

Mostly comments/recommendations for readability - see in line, but tl;dr

- `ProjectCheckerReal()` seems awkward as a name, let's try to keep the most descriptive name for the "actual" class
- some readability recommendations for tests

api/apps/api/src/modules/scenarios/project-checker/project-checker.service-real.ts
AaronPerezPerez: This issue is always a pain 😂 
I've always preferred to have the most straighforward naming for the interface/abstract class, as it's the one that will be appearing more times in the codebase. `ProjectCheckerReal` can be improved but, as an example with repositories:
Abstract class: `ProjectRepository`
In memory implementation: `MemoryProjectRepository`
TypeORM implementation: `TypeORMProjectRepository` 
In this example (for me) seems the mos correct way of naming, as I see a little bit confusing to have `ProjectChecker` implementing `ProjectValidator`.
Returning to the original naming, I think the best way to handle this is having the abstract class as `ProjectChecker` and looking for a better name for the real implementation that represents that it is that, the actual implementation.

api/apps/api/src/modules/scenarios/project-checker/project-checker.service-real.ts
AaronPerezPerez: Much clearer!

api/apps/api/src/modules/scenarios/scenarios.service.ts
AaronPerezPerez: Mmm maybe some kind of provider that we could inject in the services that need it could be enough.
Something like:
```typescript 
export class ProjectEditGuardRealImplementationWithABetterName extends ProjectEditGuard {}
......
// And then we could use it like this in just one line:
await this.projectEditGuard.ensureEdittingIsAllowed(scenarioId)
```
If you like this (I actually do, didn't think about this before) we can implement it quickly!

api/apps/api/src/modules/scenarios/project-checker/project-checker.service-real.ts
hotzevzl: agree that it's always a pain - though sometimes it can also be a (code) smell 😅 

thing is, in the kind of use we're making here and elsewhere, it's not like (in pseudocode) Dog implements abstract Animal.

I'd suggest `MarxanProjectChecker` for the "real" one maybe?

api/apps/api/src/modules/scenarios/scenarios.service.ts
hotzevzl: yep, I agree with this approach.

I'd recommend to keep it as is in this PR and move it to a provider in a separate task (but up to you)

AaronPerezPerez: I think that it would be ok to abstract it in this PR to kept it clean 🤔

api/apps/api/src/modules/scenarios/project-checker/project-checker.service-real.ts
AaronPerezPerez: Mmm ok, that'd be good 👌🏽

api/apps/api/src/modules/projects/edit-guard/marxan-edit-guard.service.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  beforeEach(async () => {
+    fixtures = await getFixtures();
+  });
+
+  it(`throws an exception if the given project has an ongoing export`, async () => {
+    const projectId = fixtures.GivenProjectWasCreated();
+
+    await fixtures.WhenExportIsRequested(projectId);
+
+    await fixtures.ThenAnExceptionIsThrown(projectId);
```
hotzevzl, line 21: usual suggestion for clarity on what we are trying to assert...

```suggestion
    await fixtures.ThenAnExceptionIsThrownIfCheckingWhetherProjectCanBeEdited(projectId);
```
In practice I'd argue this should be broken down again in another action (e.g. `WhenCheckingWhetherTheProjectCanBeEdited()`) + assertion (`ThenAnExceptionIsThrown()`), but I'd leave the decision up to you.

api/apps/api/src/modules/projects/edit-guard/marxan-edit-guard.service.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  });
+
+  it(`does nothing if the given project does not have an ongoing export`, async () => {
+    const projectId = fixtures.GivenProjectWasCreated();
+
+    await fixtures.ThenNothingHappens(projectId);
```
hotzevzl, line 27: as above, I'd suggest to try to clarify the naming of this, if not fully breaking it down (which would make even more sense here where we're outright missing a `when` step), e.g.
```suggestion
    await fixtures.NoExceptionIsThrownIfCheckingWhetherProjectCanBeEdited(projectId);
```

as it is now, although the code is testing the right behaviour, our description reads "given a project was created, then nothing happens", which at to someone reading it at a glance may sound confusing

hotzevzl (approved):
great, I like this :)

and sorry for being a pain with "naming things" 😬

api/apps/api/src/modules/projects/edit-guard/marxan-edit-guard.service.spec.ts
AaronPerezPerez: I had the same doubt, I think is more correct splitting it in two steps 👌🏽

PR #743, opened 2022-01-10
MARXAN-832 project resource pieces adapter

aciddaute:
This PR adds a resource pieces adapter for projects

### Feature relevant tickets

[Jira](https://vizzuality.atlassian.net/browse/MARXAN-832)

### DIFF
--- a/api/apps/api/src/modules/clone/export/adapters/resource-pieces.adapter.ts
+++ b/api/apps/api/src/modules/clone/export/adapters/resource-pieces.adapter.ts
@@ -25,7 +25,9 @@ export class ResourcePiecesAdapter implements ResourcePieces {
     );
 
     if (!resourcePiecesProvider) {
-      throw new Error(`ResourcePieces adapter not found for ${kind}`);
+      throw new Error(
+        `ResourcePieces adapter not found for ${kind} resource kind. This is probably caused by a missing provider in export-adapters module`,
+      );
     }
 
     return resourcePiecesProvider.discoveredClass.instance as ResourcePieces;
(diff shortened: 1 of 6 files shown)


api/apps/api/src/modules/clone/export/adapters/resource-pieces.adapter.ts
```diff
@@ -1,45 +1,42 @@
...
+    const resourcePiecesProviders = await this.discovery.providersWithMetaAtKey<ResourceKind>(
+      ResourcePiecesProviderMetadata,
+    );
+
+    const resourcePiecesProvider = resourcePiecesProviders.find(
+      (provider) => provider.meta === kind,
+    );
+
+    if (!resourcePiecesProvider) {
+      throw new Error(`ResourcePieces adapter not found for ${kind}`);
```
hotzevzl, line 28: just a note - when we get to throw this error it should normally be because of a bug (or missing implementation), not something we fail to do because of user input/data available/etc. I wonder if we should make this more explicit somehow (in the Marxan bots code I used something like `No feature found with given name: this is probably a bug in this bot` as an error message.
In this case I'm not overly concerned because this error message would likely not make it through to the user if they use the web app, but it may be available for us devs (depending on how it's handled - likely turned into a 500 by the global exception filter). Just food for thought 😄

hotzevzl (approved):

tiagojsag (approved):

api/apps/api/src/modules/clone/export/adapters/resource-pieces.adapter.ts
aciddaute: I've changed the message to `ResourcePieces adapter not found for ${kind} resource kind. This is probably caused by a missing provider in export-adapters module` 👍

PR #744, opened 2022-01-10
Feat marxan 1177 include scenario id on start calibration endpoint response

aciddaute:
### Feature relevant tickets

[Jira](https://vizzuality.atlassian.net/browse/MARXAN-1177)


hotzevzl:
@aciddaute I have discussed the request for this change with Ana and I think this may not be needed as there may actually be an underlying bug or other issue that we're trying to work around. I am triaging the underlying issue, will update this PR when I have more details.

PR #745, opened 2022-01-10
enable inspectability in docker the simple way

hotzevzl:
This should be the simplest way (simplest while running services in containers, that is) to enable debugging API and geoprocessing services via DAP.

Maybe including a `.vscode/launch.json` is not such a great idea as this may clash with local configs. I need to look into better options.

With this PR, to start both API and Geoprocessing with inspector on, use `make debug-api` instead of `make start-api`.

### DIFF
--- a/api/apps/geoprocessing/entrypoint.sh
+++ b/api/apps/geoprocessing/entrypoint.sh
@@ -5,10 +5,10 @@ case "$1" in
     develop)
         if [ $ENABLE_DEBUG_MODE = "true" ]; then
             echo "Running Development Server with inspector"
-            exec yarn api:start:debug
+            exec yarn geoprocessing:start:debug
         else
             echo "Running Development Server"
-            exec yarn api:start:dev
+            exec yarn geoprocessing:start:dev
         fi
         ;;
     test-unit)


hotzevzl:
@aagm @rubvalave please feel free to test and review this PR: my main reason for leaving it in draft status is that I wasn't sure what to do about the `launch.json` file - I'm open to suggestion, but I think I would include this with a different name (for example, within a documentation folder dedicated to debugging in containers) so that those who are interested can copy-paste the relevant launch configuration but there's no risk of clashes with any existing launch config files.

(also, `.vscode` is currently in gitignore so leaving the file where I initially placed it, by force-adding it to the repo, would not make much sense in any case)

.vscode/launch.json
```diff
@@ -0,0 +1,36 @@
+{
```
tiagojsag, line 1: please not custom IDE crap on repos. Add this to a wiki or something.

aagm: First Andrea, Thx very much for taking the time to investigate and putting it in an easy enough way to set it up locally. Said it, I am with Tiago here, adding it to a MD file in the docs should be enough to set it up locally.

hotzevzl: agree. I moved this to a folder dedicated for IDE crap that any kind soul may want to share - I'd prefer not to move it out of the repo altogether so that we can keep as much as possible of generally useful docs in a single place (rather than stuff for our internal operations which should very well be separate, and indeed is - not that we have much, but still).

tiagojsag: If everyone uses a different IDE, and we keep all those configs in here, we're gonna have a bad time. This is the same reason why we don't keep .DStore files in repos - keeping the projects as dependency-free as we can. IDEs are not a dependency of the project. As an extreme example: say those files require some local path to work? what will be the approach then?

Also: I feel like this PR should focus a lot more on modifying existing dev stuff, rather than adding yet another package.json command + docker compose file + etc.

hotzevzl: IDE configs may be a courtesy to those who start working on the project and can not only spin up the project easily but also some use specific settings that may work in their IDEs with no or little changes.

I am also happy to drop the suggested vscode debug launch config altogether and let new devs ask for it, if they need it, of for them to figure it out - it won't take more than a few minutes.

Including some example IDE configs wouldn't make any of them a project _dependency_, however, the same way we do provide a convenient Docker Compose setup for those who want to use it, but nothing in the codebase makes Linux namespaces, containers, Docker and whatnot a dependency - as long as TCP sockets and node/pg/redis are set up, one can even run all this stuff on punch cards, as far as I am concerned 😁 And also use Emacs or whatever editor. I do actually use emacs and nvim occasionally but I would not ship config snippets because these editors' configs tend to be fiddled with much more than what a typical vscode user seems to do.

Of course the more we include, the more can become outdated. It would be a balance between some convenience and not adding irrelevant stuff. I don't mind either way.

As for modifying existing dev stuff, what I could probably suggest for the task at hand is for me to modify the current tasks in `package.json` to enable to pass through `NODE_OPTIONS` from `.env` (or preferably a subset to avoid paving the way for arbitrary fiddling and other implications). I find that having a separate config for the debug setup makes the setup more explicit/self-explanatory, but I'm also open to other suggestions.

tiagojsag: I don't oppose having some convenience resources within the project (properly identified as such) if that is helpful to others, but I still have concerns if it's the best solution we can come up with.

On the existing dev stuff: I wonder if there's a simpler way for us pass a different entrypoint.sh argument through docker-compose to the api containers. Having yet another docker-compose file feels excessively noisy, but maybe that's what it takes - I am not a docker-compose expert, so I'm really just asking you if you know :)

Another thing I noticed is that the package.json already has a couple of tasks related to debugging, so it feels like we are falling down the "competing standards" xkcd meme you know I love. Since everyone but me seems to be running this on docker, I suspect those are orphaned, so maybe we can remove them. I think that it may also be a good idea to have a "package.json reference" documentation somewhere, where we add a relatively detailed description of what each package.json command does and when it should be used. Us-in-6-months can take care of the existing ones, but maybe this PR could already tackle the commands it adds, so at least we have a starting point, and not another broken window

PR #746, opened 2022-01-10
MARXAN-1178-Projects listing not working for viewers and contributors users

rubvalave:
Small bug fix by not using proper WHERE ANY clause.

### DIFF
--- a/api/apps/api/test/projects.e2e-spec.ts
+++ b/api/apps/api/test/projects.e2e-spec.ts
@@ -118,7 +118,7 @@ describe('ProjectsModule (e2e)', () => {
       );
     });
 
-    test('A user should be able to get a list of projects as owner', async () => {
+    test('A user with owner role on some projects should be able to get a list of the projects they have a role in', async () => {
       const response = await request(app.getHttpServer())
         .get('/api/v1/projects')
         .set('Authorization', `Bearer ${jwtToken}`)
@@ -130,7 +130,7 @@ describe('ProjectsModule (e2e)', () => {
       expect(jsonAPIResponse.data).toHaveLength(4);
     });
 
-    test('A user should be able to get a list of projects as contributor', async () => {
+    test('A user with contributor role on some projects should be able to get a list of the projects they have a role in', async () => {
       const response = await request(app.getHttpServer())
         .get('/api/v1/projects')
         .set('Authorization', `Bearer ${contributorToken}`)
@@ -142,7 +142,7 @@ describe('ProjectsModule (e2e)', () => {
       expect(jsonAPIResponse.data).toHaveLength(2);
     });
 
-    test('A user should be able to get a list of projects as viewer', async () => {
+    test('A user with viewer role on some projects should be able to get a list of the projects they have a role in', async () => {
       const response = await request(app.getHttpServer())
         .get('/api/v1/projects')
         .set('Authorization', `Bearer ${viewerToken}`)
@@ -154,7 +154,7 @@ describe('ProjectsModule (e2e)', () => {
       expect(jsonAPIResponse.data).toHaveLength(2);
     });
 
-    test('A user should be able to get a list of projects with q param', async () => {
+    test('A user with owner role should be able to get a list of the projects with q param where they have a role in', async () => {
       const response = await request(app.getHttpServer())
         .get('/api/v1/projects?q=User')
         .set('Authorization', `Bearer ${jwtToken}`)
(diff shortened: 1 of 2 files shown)


api/apps/api/test/projects.e2e-spec.ts
```diff
@@ -114,7 +118,7 @@ describe('ProjectsModule (e2e)', () => {
       );
     });
 
-    test('A user should be able to get a list of projects', async () => {
+    test('A user should be able to get a list of projects as owner', async () => {
```
hotzevzl, line 121: very minor - I think the test descriptions (this and for the new ones you added) don't convey clearly what we're checking here, and combined with the fact (which is here just incidental, and made worse by the original decision - mine - to create these users, projects and roles via a magic seed that is not part of the test run itself) that the user with owner role has access to 4 projects and the users with contributor and viewer roles have access to fewer projects, one needs to read more code to understand what's going on - without checking, I thought at first that we were checking that a user with owner role could see more projects than users with "lower" roles

I would suggest to say (getting a bit too long, but I'd trade a bit of verbosity for accuracy):
```suggestion
    test('A user with owner role on some projects should be able to get a list of the projects they have a role in', async () => {
```

hotzevzl (approved):

api/apps/api/test/projects.e2e-spec.ts
rubvalave: Maybe I was on the wrong here as I used the already seeded projects/users instead of creating new ones, but I wanted to avoid having a long test with the steps of creating and assigning users and projects just to be able to check them.

I added your suggestions, but If you think we would benefit of having tests with appropiate steps of creating project + creating users + assigning different roles to check the fetching of projects is done correctly just let me know.

hotzevzl: Thanks Rubén.

Avoiding to rely on the SQL seed should improve our tests, among other things by making what Tiago recommended in his review of this PR simpler and more explicit, without depending, with,in tests on some "magic" data that is created elsewhere.

For context, the SQL seed really was meant to be used very early on to have some basic data as we were starting: I think the time has come for me to propose a deletion of that seed as we have several different ways of populating data depending on use case, between CI, demo projects and tests. I am fully guilty of setting a sloppy example by also using SQL seed data in tests earlier on. Do as I say, don't do as I do, basically 😬.

However, I wouldn't want to hold up this PR for too long: I'd say if you can address Tiago's comments, the creation of users/projects/scenarios within tests can be done in a separate PR, and I'd be happy to help with anything that could make this easier (such as updating our utility functions, etc.).

tiagojsag: +9000 for removing the seeds file, and instead use a per-test data creation + cleanup
But that may be a separate task

api/apps/api/test/projects.e2e-spec.ts
```diff
@@ -123,9 +128,60 @@ describe('ProjectsModule (e2e)', () => {
       const jsonAPIResponse: ProjectResultPlural = response.body;
 
       expect(jsonAPIResponse.data[0].type).toBe('projects');
+      expect(jsonAPIResponse.data).toHaveLength(4);
+
+      const projectsArray: JSONAPIProjectData[] = jsonAPIResponse.data
```
tiagojsag, line 133:
```suggestion
      const projectNames: string[] = jsonAPIResponse.data.map(p => p.attributes.name);
      expect(projectNames).toEqualWithNoOrder(['a', 'b', ....])
```

api/apps/api/test/projects.e2e-spec.ts
```diff
@@ -123,9 +128,65 @@ describe('ProjectsModule (e2e)', () => {
...
+      const response = await request(app.getHttpServer())
+        .get('/api/v1/projects')
+        .set('Authorization', `Bearer ${viewerToken}`)
+        .expect(200);
+
+      const jsonAPIResponse: ProjectResultPlural = response.body;
+
+      expect(jsonAPIResponse.data[0].type).toBe('projects');
+      expect(jsonAPIResponse.data).toHaveLength(2);
+      const projectsArray: JSONAPIProjectData[] = jsonAPIResponse.data.sort(
```
tiagojsag, line 178:
```suggestion
      const projectNames: string[] = jsonAPIResponse.data.map(e => e.attributes.name).sort()
      expect(projectNames).toEqual(['foo', 'bar'])
```

tiagojsag (changes requested):
@rubvalave seems like you need to do one of those self-CR here

PR #749, opened 2022-01-12
export repository with support for transactions [MARXAN-871] [MARXAN-1174]

AaronPerezPerez:
### Overview

Implements the Export Repository for TypeORM with a find and save method. Also, this PR adds tests for the repository, and finally we also added a transaction method to do transactional tasks.

## Related issues
[MARXAN-1147 Export Repository should support transactions](https://vizzuality.atlassian.net/browse/MARXAN-1147)
[MARXAN-871 Export Repository](https://vizzuality.atlassian.net/browse/MARXAN-871)


hotzevzl (changes requested):
@aciddaute @AaronPerezPerez @angelhigueraacid good job!

The implementation looks good to me, I only recommend to adjust the test fixtures to stick to clear step separation (should be a fairly minor change).

api/apps/api/test/integration/typeorm-export.repository.integration.e2e-spec.ts
```diff
@@ -179,5 +194,15 @@ const getFixtures = async () => {
...
     },
+    ThenAllExportComponentsShouldBeFinished: ({
+      exportData,
+    }: {
+      exportData: Export | undefined;
+    }) => {
+      expect(exportData).toBeDefined();
+      exportData!.toSnapshot().exportPieces.map((piece) => {
+        expect(piece.finished).toEqual(true);
+      });
```
kgajowy, line 205: assertions in mapping look really strange ; )

`[{finished: true}, {finished: false}].every(i => i.finished)`

aciddaute: I'll fix it with a `forEach` :+1:

api/apps/api/src/modules/clone/export/adapters/entities/exports.api.entity.ts
```diff
@@ -0,0 +1,62 @@
...
+  static fromAggregate(exportAggregate: Export): ExportEntity {
+    const snapshot = exportAggregate.toSnapshot();
+
+    const exportEntity = new ExportEntity();
+
+    exportEntity.id = snapshot.id;
+    exportEntity.resourceId = snapshot.resourceId;
+    exportEntity.resourceKind = snapshot.resourceKind;
+    exportEntity.archiveLocation = snapshot.archiveLocation;
+    exportEntity.components = snapshot.exportPieces.map((piece) => {
```
kgajowy, line 43: also consider `class-transformer` to avoid too much typing

api/apps/api/src/modules/clone/export/adapters/typeorm-export.repository.ts
```diff
@@ -1,16 +1,63 @@
...
+      resourceKind: exportEntity.resourceKind,
+      archiveLocation: exportEntity.archiveLocation,
+      exportPieces: exportEntity.components.map((component) => ({
+        id: new ComponentId(component.id),
+        finished: component.finished,
+        piece: component.piece,
+        resourceId: component.resourceId,
+        uris: component.uris.map(
+          (element) => new ComponentLocation(element.uri, element.relativePath),
+        ),
```
kgajowy, line 42: Looks like mapping knowledge is now spread to both repo and entity. Consider putting it in one place.

api/apps/api/test/integration/typeorm-export.repository.integration.e2e-spec.ts
```diff
@@ -0,0 +1,208 @@
...
+  it('should update an export when a nested piece is completed', async () => {
+    await fixtures.WhenAnExportIsSavedSuccessfully();
+    await fixtures.WhenAComponentIsCompleted();
+    const exportData = await fixtures.WhenReadingTheSavedExportFromRepository();
+
+    await fixtures.ThenExportDataShouldBeOk({
+      exportData,
+      componentsAreCompleted: true,
+    });
+  });
```
kgajowy, line 45: this may be a bit confusing. Test says `should update an export when a nested piece is complete` so I would rather expect more process/product naming and 'events' what actually happen.

```
GivenExportWasRequested();
WhenLastPieceWasCompleted();
ThenExportIsSaved();
ThenExportFinalizedEventIsSent();  // don't remember the name, sorry
```

api/apps/api/test/integration/typeorm-export.repository.integration.e2e-spec.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  it('should save entities working with transaction', async () => {
+    await fixtures.WhenAnExportIsSavedSuccessfullyWithinATransaction();
+    await fixtures.WhenAllComponentsAreCompletedWithTransactions();
+
+    const exportData = await fixtures.WhenReadingTheSavedExportFromRepository();
+
+    fixtures.ThenAllExportComponentsShouldBeFinished({
+      exportData,
+    });
```
kgajowy, line 55: `withTransactions` seems to be irrelevant technical detail there 🤔

api/apps/api/test/integration/typeorm-export.repository.integration.e2e-spec.ts
kgajowy: @aciddaute but why `forEach`, if you can do that with `every()` ?

aciddaute: Oh sorry, I didn't see that, I'll use `every()` ty Kamil 🙌

api/apps/api/test/integration/typeorm-export.repository.integration.e2e-spec.ts
AaronPerezPerez: As this is an integration test for our custom repository, this test should not know about events. Also, the second step is just one completed piece, not every one / the last one

api/apps/api/test/integration/typeorm-export.repository.integration.e2e-spec.ts
AaronPerezPerez: As we are testing the repository, we wanted to have another integrtation test for that specific method of our repository

kgajowy: @AaronPerezPerez hm, steps deceived me. How about then testing it truly as an integration?

```
export = GivenTheExport()
WhenSaving(fixtures.completePieceA(export);
WhenSaving(fixtures.completePieceB(export);
ThenBothPiecesAreSaved(); // just use entity => export like in normal application and verify on "domain" object.
```

api/apps/api/src/modules/clone/export/adapters/export-adapters.module.ts
```diff
@@ -6,15 +6,25 @@ import { ExportRepository } from '../application/export-repository.port';
 import { ResourcePieces } from '../application/resource-pieces.port';
 import { ArchiveCreator } from '../application/archive-creator.port';
 import { ResourcePiecesAdapter } from './resource-pieces.adapter';
-import { InMemoryExportRepo } from './in-memory-export.repository';
 import { NodeArchiveCreator } from './node-archive-creator';
+import { TypeormExportRepository } from '@marxan-api/modules/clone/export/adapters/typeorm-export.repository';
+import { ExportEntity } from '@marxan-api/modules/clone/export/adapters/entities/exports.api.entity';
+import { ExportComponentEntity } from '@marxan-api/modules/clone/export/adapters/entities/export-components.api.entity';
+import { ComponentLocationEntity } from '@marxan-api/modules/clone/export/adapters/entities/component-locations.api.entity';
```
kgajowy, line 13: relative imports (:

api/apps/api/test/integration/typeorm-export.repository.integration.e2e-spec.ts
```diff
@@ -0,0 +1,212 @@
...
+import { ClonePiece, ComponentId, ResourceKind } from '@marxan/cloning/domain';
+import { FixtureType } from '@marxan/utils/tests/fixture-type';
+import { Connection } from 'typeorm';
+import { v4 } from 'uuid';
+import { ResourceId } from '../../src/modules/clone/export';
+import { ExportEntity } from '../../src/modules/clone/export/adapters/entities/exports.api.entity';
+import { ComponentLocation } from '../../src/modules/clone/export/application/complete-piece.command';
+import { ExportRepository } from '../../src/modules/clone/export/application/export-repository.port';
+import { Export, ExportId } from '../../src/modules/clone/export/domain';
+import { bootstrapApplication } from '../utils/api-application';
```
kgajowy, line 12: @paths there on the other hand

api/apps/api/test/integration/typeorm-export.repository.integration.e2e-spec.ts
```diff
@@ -0,0 +1,212 @@
...
+      const connection = app.get<Connection>(Connection);
+      const exportRepo = connection.getRepository(ExportEntity);
+      const exportComponentRepo = connection.getRepository(
+        ExportComponentEntity,
+      );
+      const locationRepo = connection.getRepository(ComponentLocationEntity);
+
+      await locationRepo.delete({});
+      await exportComponentRepo.delete({});
+      await exportRepo.delete({});
```
kgajowy, line 80: could be a good secret test of cascade delete (running only `await exportRepo.delete({});`)

api/apps/api/test/integration/typeorm-export.repository.integration.e2e-spec.ts
aciddaute: But if we want to test the `transaction` method behaviour we should complete several pieces concurrently 🤔 . The steps of the current test are:

1. `GivenExportWithMultipleComponentsWasRequested`: Requests an export with 10 pieces
2. `WhenAllComponentsAreCompletedConcurrentlyWithTransactions`: Completes those pieces concurrently using `transaction` method
3. `WhenReadingTheSavedExportFromRepository`: Reads the export from the repo
4. `ThenAllExportComponentsShouldBeFinished`: Checks that all pieces are marked as finished

Note that the code above is outdated.

api/apps/api/src/modules/clone/export/adapters/export-adapters.module.ts
tiagojsag: Do we have a documented rule of when to use absolute/relative imports? I've seen multiple comments on PRs around this, so I'm wondering if everyone is aware of exactly how we should approach this - I am not, for one.

api/apps/api/src/modules/clone/export/adapters/export-adapters.module.ts
```diff
@@ -6,15 +6,25 @@ import { ExportRepository } from '../application/export-repository.port';
...
+    TypeOrmModule.forFeature([
+      ExportEntity,
+      ExportComponentEntity,
+      ComponentLocationEntity,
+    ]),
+  ],
   providers: [
     {
       provide: ExportRepository,
-      useClass: InMemoryExportRepo, // TODO TypeormExportRepository
```
tiagojsag, line 17: This seems like the last usage of the InMemory repo besides tests, so we should probably get rid of it.
A related question: why is ExportRepository an abstract class instead of an interface?

aciddaute: In order to be able to use `ExportRepository` as the `provide` property of the provider. We can't use a Typescript interface as a value for `provide` property. We could also declare it as an interface and use a Symbol for the `provide` property along as we use the `@Inject` decorator for DI. `TypeormExportRepository` and `InMemoryExportRepo` implements the abstract class, so we should be safe 🤔

hotzevzl (approved):

api/apps/api/src/modules/clone/export/adapters/export-adapters.module.ts
hotzevzl: we don't, and I got lost in the evolution of this through time.

My preference would be:
- within a Nest module, relative paths
- across modules within the same service, "root-relative" paths
- across services or from libs, `@<whatever>` "root-relative" paths

though IDEs should make the difference between the first two cases mostly cosmetic - personally I like "root-relative" paths as I find that it removes the small cognitive overload that (at least for me) relative paths introduce by having to remember or figure out "where am I" in the paths hierarchy to then travel up and down the tree mentally. But even this, I normally deal with ok by just ctrl-clicking through the relevant imports if I can't figure them out at a glance.

tiagojsag (approved):

PR #751, opened 2022-01-13
hack to investigate MARXAN-987

hotzevzl:
See [MARXAN-987](https://vizzuality.atlassian.net/browse/MARXAN-987) for context. This branch is not meant to be merged as is, but it may hopefully be helpful to further dig beyond what I can understand.

A previous fix (see recent commits in `layout/scenarios/edit/analysis/adjust-planning-units/buttons/uploading/component.tsx` did work around other sides of this issue, but also prevented GeoJSON files with >1 Polygon-type features to be processed correctly, as only the first feature was being taken into account.

The only way we could make GeoJSON with Multipolygon-type features (such as the one that is provided in MARXAN-987 to help reproduce the original issue with the failed `@math.gl/webmercator` assert, `high_human_pressure.zip`) was to keep the previous changes (0d1f9cfd86aaf4a5209da58594d645db1ab7d93f) *but also* setting `type: "Multipolygon"` for the features as reshaped in that commit. This would still only process the first feature (which in some cases may be the only one, and therefore enough, but not in general), but there's also a lot of ad-hoc processing of GeoJSON data which is likely brittle with edge cases and so on.

### DIFF
--- a/app/layout/scenarios/edit/analysis/adjust-planning-units/buttons/uploading/component.tsx
+++ b/app/layout/scenarios/edit/analysis/adjust-planning-units/buttons/uploading/component.tsx
@@ -111,16 +111,44 @@ export const AnalysisAdjustUploading: React.FC<AnalysisAdjustUploadingProps> = (
 
         const validGeoJSON = {
           ...g,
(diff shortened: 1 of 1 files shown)


hotzevzl:
closing this as not needed beyond triage.

PR #752, opened 2022-01-13
fix(front): update features list after uploading new one[MARXAN-1189]

anamontiaga:
### Overview

This PR contains the fix of an error related to updating all the features in the modal to add new features. It was detected that when uploading a new feature, the change was not reflected in the list unless the user closed the modal and reopened it.

### Testing instructions

- [ ] Add a new feature by uploading. (See attached file on JIRA task). 
- [ ] Classify it as bioregional.
- [ ] Check if features list is automatically updated on upload modal.

### Feature relevant tickets

[MARXAN-1189](https://vizzuality.atlassian.net/browse/MARXAN-1189)

### DIFF
--- a/app/layout/scenarios/edit/features/add/list/component.tsx
+++ b/app/layout/scenarios/edit/features/add/list/component.tsx
@@ -1,4 +1,6 @@
-import React, { useCallback } from 'react';
+import React, { useCallback, useEffect } from 'react';
+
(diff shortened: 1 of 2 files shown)


SARodrigues (approved):
Works! Does take a little to show up

I wonder if there's a way we can display a "loading" icon or something in the future, to let the user know, while it doesn't show up in the list

PR #754, opened 2022-01-13
[WIP][MARXAN-1122]adds-ACL-scenarios-controller

rubvalave:
-Adds ACL to every endpoint in the scenarios-controller.
-Fixes some missing ImplementsACL decorator not present in projects-listing-controller.
-Various additions to ACL-utils.


api/apps/api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -213,9 +234,20 @@ export class ScenariosController {
   @Get(':id/planning-units/tiles/:z/:x/:y.mvt')
   async proxyProtectedAreaTile(
     @Req() request: Request,
+    @Req() req: RequestWithAuthenticatedUser,
```
tiagojsag, line 237: Do we really need the request twice here? The arg you added extends Request, so is there any reason why it can't simply replace the existing Request?

api/apps/api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -213,9 +234,20 @@ export class ScenariosController {
...
   ) {
-    return this.proxyService.proxyTileRequest(request, response);
+    /* Due to the usage of proxyService in other modules
+    the ACL control for this endpoint is placed in the controller */
+    if (
+      !(await this.scenarioAclService.canViewScenario(req.user.id, scenarioId))
+    ) {
+      throw new ForbiddenException();
+    }
+    const result = await this.proxyService.proxyTileRequest(request, response);
```
tiagojsag, line 248:
```suggestion
    return this.proxyService.proxyTileRequest(request, response);
```

api/apps/api/src/modules/scenarios/scenarios.service.ts
```diff
@@ -319,49 +476,86 @@ export class ScenariosService {
   async getOneSolution(
     scenarioId: string,
     runId: string,
+    userId: string,
     _fetchSpecification: FetchSpecification,
-  ) {
+  ): Promise<Either<typeof forbiddenError, ScenariosOutputResultsApiEntity>> {
     await this.assertScenario(scenarioId);
-    // TODO permissions guard
-    // TODO runId is part of scenarioId
```
tiagojsag, line 326: Has this been handled?

api/apps/api/src/modules/scenarios/scenarios.service.ts
```diff
@@ -408,67 +606,133 @@ export class ScenariosService {
...
+  async createSpecification(
+    scenarioId: string,
+    userId: string,
+    dto: CreateGeoFeatureSetDTO,
+  ): Promise<Either<typeof forbiddenError | typeof internalError, any>> {
     const scenario = await this.assertScenario(scenarioId);
+    if (!(await this.scenarioAclService.canEditScenario(userId, scenarioId))) {
+      return left(forbiddenError);
+    }
     return await this.specificationService.submit(
```
tiagojsag, line 671: right() missing here?

api/apps/api/src/modules/scenarios/scenarios.service.ts
```diff
@@ -408,67 +606,133 @@ export class ScenariosService {
...
+    Either<
+      LastUpdatedSpecificationError | typeof forbiddenError,
+      CreateGeoFeatureSetDTO
+    >
+  > {
     const scenario = await this.assertScenario(scenarioId);
+    if (!(await this.scenarioAclService.canViewScenario(userId, scenarioId))) {
+      return left(forbiddenError);
+    }
     return await this.specificationService.getLastUpdatedFor(
```
tiagojsag, line 691: and here?

tiagojsag (changes requested):
- Seems like you have a few rigth() calls missing. I highlighted a couple, but there seem to be more
- lots of tests missing - those missing right() calls should have been caught by the tests you didn't add
- return await is more often than not an anti pattern

hotzevzl (commented):
thanks @rubvalave. I am not too convinced with the new `findAllPaginatedAcl()` method, as this would leave the base class' own `findAllPaginated()` open to accidental use without access control checks.

One possible way to handle this would be what I sketched quickly in https://github.com/Vizzuality/marxan-cloud/tree/feat/MARXAN-1122-scenario-acl-application%2Bandrea-suggestions.

That's still not ideal because the base class then becomes coupled with ACL concerns, which I'd say could be ok because it's the marxan backend's own wrapper for `BaseService`, but one could also argue otherwise.

Another possible way would be to make `BaseService.findAllPaginated()` a private method for internal use, renaming it for example to `BaseService._findAllPaginated()` as with other methods that we wouldn't want consumers to rely on, and implement `findAllPaginated()` as per your implementation, but with the leaner name (without the `Acl` part).

Either way, there could be arguments for and against, so I'm just bringing forward some suggestions.

hotzevzl:
> thanks @rubvalave. I am not too convinced with the new `findAllPaginatedAcl()` method, as this would leave the base class' own `findAllPaginated()` open to accidental use without access control checks.
> 
> One possible way to handle this would be what I sketched quickly in https://github.com/Vizzuality/marxan-cloud/tree/feat/MARXAN-1122-scenario-acl-application%2Bandrea-suggestions.

so quickly in fact that it's obviously incomplete and doesn't even transpile, but just to give a rough idea 😇

api/apps/api/src/modules/scenarios/scenarios.controller.ts
hotzevzl: my fault (mindlessly accepting IDE suggestions) - `RequestWithAuthenticatedUser` is a remnant of past projects, and I brought it over importing the wrong `Request`, so fixing `RequestWithAuthenticatedUser` to extend `Request` from Express is the correct way to handle this.

api/apps/api/src/modules/scenarios/scenarios.service.ts
rubvalave: Permissions yes, but I overlooked that I removed the latter part so I added it again.

api/apps/api/test/scenario-input-files/scenario-cost-surface/scenario-cost-surface.e2e-spec.ts
```diff
@@ -1,19 +1,61 @@
...
-describe(`When scenario has PUs with cost and lock status`, () => {
-  beforeAll(async () => {
-    await world.GivenScenarioWithPuAndLocks();
+describe(`As owner, when scenario has PUs with cost and lock status`, () => {
+  it(`returns relevant data for pu.dat`, async () => {
+    const result = await fixtures.WhenGettingMarxanDataAsOwner();
+    const [headers, ...costAndStatus] = result.split('\n');
+
+    expect(headers).toEqual('id\tcost\tstatus');
+    expect(costAndStatus).toMatchInlineSnapshot(`
```
AaronPerezPerez, line 18: Wouldn't it be better to use .toEqual() ?
It seems clearer IMHO
```suggestion
    expect(costAndStatus).toEqual([
        "0	200	0",
        "1	400	2",
        "2	600	3",
        "3	800	0",
      ])
```

api/apps/api/test/scenario-input-files/scenario-cost-surface/scenario-cost-surface.e2e-spec.ts
```diff
@@ -54,6 +96,62 @@ describe(`When scenario has PUs with cost and lock status`, () => {
...
 
+describe(`As viewer, when scenario has PUs with cost and lock status`, () => {
+  it(`returns relevant data for pu.dat`, async () => {
+    await fixtures.GivenViewerWasAddedToScenario();
+    const result = await fixtures.WhenGettingMarxanDataAsViewer();
+    const [headers, ...costAndStatus] = result.split('\n');
+
+    expect(headers).toEqual('id\tcost\tstatus');
+    expect(costAndStatus).toMatchInlineSnapshot(`
+      Array [
```
AaronPerezPerez, line 107: Same as before

api/apps/api/test/scenario-protected-areas/fixtures.ts
```diff
@@ -124,6 +167,17 @@ export const getFixtures = async () => {
         selected: true,
       });
     },
+    ThenItDoesNotContainsSelectedCustomArea: async (
+      response: any[],
```
AaronPerezPerez, line 171: Can we add proper typing to this param?

api/apps/api/test/scenario-protected-areas/fixtures.ts
```diff
@@ -165,6 +262,17 @@ export const getFixtures = async () => {
         selected: true,
       });
     },
+    ThenItDoesNotContainsSelectedGlobalArea: async (
+      response: any[],
```
AaronPerezPerez, line 266: Same as before

api/apps/api/test/scenario-solutions/get-all-solutions.fixtures.ts
```diff
@@ -0,0 +1,118 @@
...
+    WhenGettingSolutionsAsContributor: async () =>
+      request(app.getHttpServer())
+        .get(`/api/v1/scenarios/${scenarioId}/marxan/solutions`)
+        .set('Authorization', `Bearer ${contributorToken}`),
+    WhenGettingSolutionsAsViewer: async () =>
+      request(app.getHttpServer())
+        .get(`/api/v1/scenarios/${scenarioId}/marxan/solutions`)
+        .set('Authorization', `Bearer ${viewerToken}`),
+    ThenSolutionsShouldBeResolved: async (response: request.Response) => {
+      expect(response.body.meta).toMatchInlineSnapshot(`
```
AaronPerezPerez, line 84: Again, I prefer using `.toEqual()`, but above all I prefer that we use the same method (any of them) everywhere
```suggestion
      expect(response.body.meta).toEqual({
        "page": 1,
        "size": 25,
        "totalItems": 1,
        "totalPages": 1,
      })
```

api/apps/api/test/scenario-solutions/get-all-solutions.fixtures.ts
```diff
@@ -0,0 +1,118 @@
...
+      Object {
+        "page": 1,
+        "size": 25,
+        "totalItems": 1,
+        "totalPages": 1,
+      }
+    `);
+      expect(response.body.data.length).toEqual(1);
+      expect(response.body.data[0].attributes).toMatchInlineSnapshot(
```
AaronPerezPerez, line 93: Same as before

AaronPerezPerez (changes requested):

api/apps/api/test/scenario-input-files/scenario-cost-surface/scenario-cost-surface.e2e-spec.ts
rubvalave: To be honest, for most of those that use `MatchInLineSnapshot`, I just added variants without changing much the original tests done by Kamil. I guess this is a matter of preference, and yeah maybe it will be better if we use the same method for every test.

tiagojsag: I second @AaronPerezPerez recommendation.

tiagojsag (approved):
TL;DR, so I approve, but if you are up for changing those `toMatchInlineSnapshot` calls, great

hotzevzl (approved):

AaronPerezPerez (approved):

api/apps/api/test/scenario-input-files/scenario-cost-surface/scenario-cost-surface.e2e-spec.ts
hotzevzl: @rubvalave though I'm fine with either, I think the _matter of preference_ bit is mostly about workflow, not so much code style per se.

Snapshots (and the inline variant used throughout these tests) can be useful if wanting to reap some free labour from Jest, as they can be updated via `jest --updateSnapshot ----testNamePattern <pattern>` rather than having to manually edit the expected values for `toEqual()`/`toStrictEqual()` (the latter being what we likely want in most cases, tin-foil hat on).

I'd say that for small objects there is no practical difference. When checking asserting equality of complex objects such as a, say GeoJSON document, then snapshots can help substantively.

PR #755, opened 2022-01-13
[MARXAN-1191] Fixing sliders' issues (cut off labels, floating point errors)

SARodrigues:
### Overview

This PR addresses a couple of issues with the (percentage) Sliders:  

1. Sliders' labels being cut off when the slider is set at either `0%` or `100%`  
    Unfortunately this is due to how the scrolling of the the component's contents (when the contents don't fit vertically) works. This requires an `overflow-hidden` in the parent components, which is effectively why these labels get hidden.

    We could rework this, but this would require us setting correct paddings in all the sections manually (instead of letting the `Pill` component standardise them. 

    Alternatively, we can make the sliders shorter width-wise just enough to let the labels fit, which was the step taken here. I noticed that in the _Features_ section the sliders are inset so this doesn't stray from the designs much. 

2. Sliders' percentages have floating point precision errors.  
    We know the percentage will always be an integer, so we can just round them. 

### Testing instructions

Navigate to the _"Protected Areas"_ tab of a project, section `2/2`, and verify that:  
- [ ] When setting sliders to either `0%` or `100%`, the label above the thumb doesn't get cut off 
- [ ] Trying multiple values (40-60 range is usually where I get the bug triggered), click on the label and verify that the number is always an integer  

### Feature relevant tickets

[MARXAN-1187](https://vizzuality.atlassian.net/browse/MARXAN-1187)
[MARXAN-1191](https://vizzuality.atlassian.net/browse/MARXAN-1191)

### Screenshots  
Screenshot 2022-01-13 at 16 33 56
Screenshot 2022-01-13 at 16 33 50
Screenshot 2022-01-13 at 16 35 39

### DIFF
--- a/app/components/forms/slider/value/component.tsx
+++ b/app/components/forms/slider/value/component.tsx
@@ -56,7 +56,8 @@ export const Value: React.FC<ValueProps> = ({
   const [isEditing, setIsEditing] = useState<boolean>(false);
 
   // Percentages come in a range from 0.0 to 1, so we have to multiply by 100 to display them
-  const displayValue = (formatOptions.style === 'percent') ? value * 100 : value;
+  // Due to floating point precision issues, we need to trunc percentages' display values
+  const displayValue = (formatOptions.style === 'percent') ? Math.round(value * 100) : value;
 
   // CANCELLING AND SAVING EDITS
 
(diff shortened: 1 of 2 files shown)


mbarrenechea (approved):
I think it's the best option right now

PR #757, opened 2022-01-17
Home banner: improve animations

mbarrenechea:
### Overview

Use transform properties instead of position rules

https://css-tricks.com/tale-of-animation-performance/


anamontiaga (approved):

SARodrigues:
> https://css-tricks.com/tale-of-animation-performance/

Very interesting, I had no idea!

PR #758, opened 2022-01-17
Adjust planning unit: support multipolygon

mbarrenechea:
### Overview

Add support for multipolygons inside the "adjust planning unit upload file"


anamontiaga:
@mbarrenechea a delay has been introduced in BE and now we have a change in status (from 'running' to 'done'). Even so, the fetch of the planning units is not carried out again after the done status appears. Not sure what is the best way to solve it.

Also I noticed when uploading smaller files layer updates.

mbarrenechea:
> @mbarrenechea a delay has been introduced in BE and now we have a change in status (from 'running' to 'done'). Even so, the fetch of the planning units is not carried out again after the done status appears. Not sure what is the best way to solve it.
> 
> Also I noticed when uploading smaller files layer updates.

Checking it

PR #759, opened 2022-01-17
[MARXAN-1195] Fixing adjusting planning units triggering a dispatch/render loop

SARodrigues:
### Overview

This PR mainly addresses an issue with the [ScenariosSidebarAnalysisSections](https://github.com/Vizzuality/marxan-cloud/blob/develop/app/layout/scenarios/edit/analysis/adjust-planning-units/component.tsx) component. It uses an `useEffect` hook to update all `puIncludedValue`, `puExcludedValue`, `puTmpIncludedValue`, `puTmpExcludedValue`, with `PUData` (which comes from the API) which ends up running on every render.  

However, a dispatch for any of those _will_ trigger a re-render; this is particularly problematic when the user chooses to _"Select planning units"_, the respective `AnalysisAdjustClicking` component (along with any click) will be setting `puTmpIncludedValue`, `puTmpExcludedValue` causing a loop. 

&nbsp;

The main solution implemented was to check whether the user is using that method, and prevent setting `puTmpIncludedValue`, `puTmpExcludedValue`, which will already be set, which in turn stops the loop.

In order to prevent similar issues down the line, I took the liberty to compare `PUData`'s `include`/`exclude` to the existing arrays in the store, so that we don't update the store unnecessarily (triggering re-renders throughout the app), (using lodash's [xor](https://lodash.com/docs/4.17.15#xor) to check whether the arrays are equivalent, even if the order of their elements is different).  

&nbsp;  

Additionally, I noticed that the _"Clear"_ button doesn't work as one would expect. It does do its expected job of clearing PU included/excluded values, when the user isn't in the middle of selecting planning units, but it won't clear the temporary ones "in progress", so I made sure to clear those as well, so that the map is truly reset to consistent values as those coming from the API. 

&nbsp; 

### Testing instructions

This bug requires testing the code locally. 

Add a `console.log('RENDER');` to either (well, ideally both): 
```
/app/layout/scenarios/edit/analysis/adjust-planning-units/component.tsx
/app/layout/scenarios/edit/analysis/component.tsx
```

Then: 
- Open a Project   
- Click _"Edit"_ on a Scenario  
- Click the _"Analysis"_ tab  
- Click the _"Adjust planning units"_  section  
- Adjust planning units using all three methods, for both _"Include areas"_ and _"Exclude areas"_  
- Verify that:  
    - [x] Adjusting the planning units works correctly on all scenarios/methods, both when including and excluding areas   
    - [x] There is no `RENDER` loop in the console  
        There'll be multiple renders, but it should never enter a loop. It'll be obvious if it happens. 

### Feature relevant tickets

- [MARXAN-1195](https://vizzuality.atlassian.net/browse/MARXAN-1195)
- [MARXAN-1067](https://vizzuality.atlassian.net/browse/MARXAN-1067) (Possibly the ultimate fix for this one as well)

### DIFF
--- a/app/layout/scenarios/edit/analysis/adjust-planning-units/buttons/clicking/component.tsx
+++ b/app/layout/scenarios/edit/analysis/adjust-planning-units/buttons/clicking/component.tsx
@@ -94,7 +94,6 @@ export const AnalysisAdjustClicking: React.FC<AnalysisAdjustClickingProps> = ({
     }, {
       onSuccess: ({ data: { meta } }) => {
         dispatch(setJob(new Date(meta.isoDate).getTime()));
-        setSubmitting(false);
         onSelected(null);
         dispatch(setClicking(false));
 
@@ -110,7 +109,6 @@ export const AnalysisAdjustClicking: React.FC<AnalysisAdjustClickingProps> = ({
         });
       },
       onError: () => {
-        setSubmitting(false);
         addToast('adjust-planning-units-error', (
           <>
             <h2 className="font-medium">Error!</h2>
@@ -121,6 +119,9 @@ export const AnalysisAdjustClicking: React.FC<AnalysisAdjustClickingProps> = ({
         ), {
           level: 'error',
         });
+      },
+      onSettled: () => {
+        setSubmitting(false);
       },
     });
   }, [sid,
--- a/app/layout/scenarios/edit/analysis/adjust-planning-units/component.tsx
+++ b/app/layout/scenarios/edit/analysis/adjust-planning-units/component.tsx
@@ -7,6 +7,7 @@ import { useRouter } from 'next/router';
 import { getScenarioEditSlice } from 'store/slices/scenarios/edit';
 
 import { motion } from 'framer-motion';
+import { xor } from 'lodash';
 
 import { useScenarioPU, useSaveScenarioPU } from 'hooks/scenarios';
 import { useToasts } from 'hooks/toast';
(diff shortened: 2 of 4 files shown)


app/layout/scenarios/edit/analysis/adjust-planning-units/buttons/clicking/component.tsx
```diff
@@ -122,6 +120,9 @@ export const AnalysisAdjustClicking: React.FC<AnalysisAdjustClickingProps> = ({
           level: 'error',
         });
       },
+      onSettled: () => {
```
SARodrigues, line 123: About these... Not nitpicking, but when debugging I went on a different direction. Figured I might as well leave these changes in. I can removed them if needed; moved them into a separate commit which can be easily reverted.

app/layout/scenarios/edit/analysis/adjust-planning-units/component.tsx
```diff
@@ -71,6 +114,10 @@ export const ScenariosSidebarAnalysisSections: React.FC<ScenariosSidebarAnalysis
     const { includedDefault, excludedDefault } = PUData;
     setClearing(true);
 
+    // Clear all temp PU values
+    dispatch(setTmpPuIncludedValue(includedDefault));
```
SARodrigues, line 118: This is what makes the `Clear` button also clear ongoing changes from the map, as to not confuse the user.

anamontiaga (approved):
Tested locally. No loops. Three adjust PU methods for lock-in/lock-out working ok

PR #760, opened 2022-01-17
Choose scenario type modal text updates

SARodrigues:
### Overview

This PR updates the text on the _"Choose scenario type"_ modal.  

I also took the liberty of moving the text into a `constants.ts` file, to match what we have going on in recent components.   

### Testing instructions

- Open a project  
- Click _"Create scenario"_  
    - [ ] Verify that the multiple scenarios text matches the ones in the task  

### Feature relevant tickets

[MARXAN-1037](https://vizzuality.atlassian.net/browse/MARXAN-1037)

### Screenshot   

Screenshot 2022-01-17 at 17 10 34

### DIFF
--- a/app/layout/projects/show/scenarios/types/component.tsx
+++ b/app/layout/projects/show/scenarios/types/component.tsx
@@ -7,48 +7,7 @@ import cx from 'classnames';
 
 import Icon from 'components/icon';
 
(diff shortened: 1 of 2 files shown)


SARodrigues:
This PR only really changes the text as requested by the client. I noticed that the text in the columns isn't all at the same level, when compared to the designs, but I'll make a comment on the JIRA task; it should be a separate issue in case we do have to make that change (it'll take longer than a simple text update).

PR #772, opened 2022-01-20
Feat/marxan 1145 handle failed processed pieces

aciddaute:
### Feature relevant tickets

[Mark export as failed if any of the pieces failed to process](https://vizzuality.atlassian.net/browse/MARXAN-1145)


api/apps/api/src/modules/clone/infra/export/mark-export-as-failed.handler.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    if (!kind) {
+      // TODO update with Scenario once supported.
+      return;
+    }
+
+    await this.apiEvents.createIfNotExists({
+      kind,
+      topic: resourceId.value,
+      externalId: kind + exportId,
```
hotzevzl, line 32: can we stick to a fixed pattern for these `externalId`s for this specific functionality? with or without hyphen between `kind` and `exportId`... I'd generate these `externalId`s through a function so that we can apply the same template consistently (as long as we always use the function to generate the value of this property 😬)

api/apps/api/src/modules/planning-units/planning-units-completion.service.ts
```diff
@@ -74,6 +74,6 @@ export class PlanningUnitsCompletionService
 
   private async completed(event: EventData<PlanningUnitsJob>) {
     const jobData = await event.data;
-    await this.eventBus.publish(new PlanningUnitSet(jobData.projectId));
+    this.eventBus.publish(new PlanningUnitSet(jobData.projectId));
```
hotzevzl, line 77: why the removed `await`s here and in a couple more occurrences below, including in a test?

api/apps/api/src/modules/clone/export/application/complete-piece.handler.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+  await fixtures.WhenAPieceIsCompleted(exportInstance.id, secondPiece.id);
+
+  fixtures.ThenExportAllComponentsFinishedEventIsEmitted(exportInstance.id);
+});
+
+it('should not publish any event if export instance is not found', async () => {
+  const exportId = new ExportId(v4());
+  await fixtures.GivenNoneExportWasRequested(exportId);
+  await fixtures.WhenAPieceOfAnUnexistingExportIsCompleted(exportId);
+  fixtures.ThenNoneEventIsEmitted();
```
hotzevzl, line 57:
```suggestion
  fixtures.ThenNoEventIsEmitted();
```

hotzevzl (approved):

api/apps/api/src/modules/planning-units/planning-units-completion.service.ts
aciddaute: `publish` method of `EventBus` is synchronous [link](https://github.com/nestjs/cqrs/blob/744ef1b409168f6f1fa146b670860a8d9ae89a80/src/event-bus.ts#L55)

AaronPerezPerez: We realized that `.publish()` is sync

api/apps/api/src/modules/clone/infra/export/mark-export-as-failed.handler.ts
aciddaute: We've added `ApiEventsService.composeExternalId(id: string, kind: API_EVENT_KINDS): string` method for ensuring consistency :+1:

PR #775, opened 2022-01-24
storage for cloning/import/export: brief and high-level design [MARXAN-816]

hotzevzl:
https://vizzuality.atlassian.net/browse/MARXAN-816

### DIFF
--- a/docs/download-upload-clone/storage/brief.md
+++ b/docs/download-upload-clone/storage/brief.md
@@ -38,6 +38,12 @@ large files (e.g. > 4GiB) on personal computers.
 This would be available for download for a limited timeframe, typically
 configured by the platform admins of each MarxanCloud instance.
 
+- removing individual pieces
+
+This may be needed if pieces are eventually included in a single downloadable
+artifact. If ongoing engineering considerations lead to making the individual
+pieces available for download, this step may not be necessary.
+
 - allowing users to download the archive for a given project
 
 Users with suitable permissions on a project (see [relevant entries in the
@@ -56,3 +62,13 @@ and its scenarios, without having to download any archive artifacts, in some
 cases they may choose to first download a copy of the final archive: for
 example, for offline safeguarding/archiving, or to upload the archive to a
 different MarxanCloud instance.
+
+- checking the integrity of an uploaded archive
+
+Although some users may wish to edit project archives offline in order to tweak
+some of the data/metadata, MarxanCloud instances will prevent this by default in
+order to avoid data issues that may lead to uploading projects in an
+inconsistent state.
+
+- garbage-collecting archives that are not yet expired if their source project
+  is deleted
--- a/docs/download-upload-clone/storage/brief.md
+++ b/docs/download-upload-clone/storage/brief.md
@@ -24,20 +24,31 @@ outlined in the [high-level design document](./high-level-design.md).
 
 - storing one or more archive pieces as they are prepared and finalized
 
+- streaming archive pieces from their cloud storage location when cloning the
+  exported project/scenarios
+
+- checking the integrity of all the pieces of a project and its scenarios as
+  they are being used to create a clone of the original project/scenarios
+
+- removing from cloud storage all the archive pieces once their validity expires
+
+- garbage-collecting archive pieces that are not yet expired, if their source
+  project or scenario is deleted
+
+Moreover, _only when users request to be able to download a single zip file_
+with all the contents of a project and any scenarios the following steps will
+also apply:
+
 - preparing a final archive file that includes all the finalized pieces
 
 - storing the final archive
 
-- removing individual pieces
-
 - allowing users to download the archive for a given project
 
-- removing the archive and any related files once its validity expires
+- removing the archive once its validity expires
 
-- allowing users to upload an archive to a MarxanCloud instance to get the
+- allowing users to upload a zip archive to a MarxanCloud instance to get the
   archived project "rehydrated" as a new project on the target instance
 
-- checking the integrity of an uploaded archive
-
-- garbage-collecting archives that are not yet expired if their source project
-  is deleted
+- garbage-collecting archive zip artifacts that are not yet expired, if their
+  source project or scenario is deleted
--- a/docs/download-upload-clone/storage/brief.md
+++ b/docs/download-upload-clone/storage/brief.md
@@ -1,0 +1,58 @@
+# Storage for MarxanCloud project archives
+
+This document outlines the key points of the storage strategy for MarxanCloud
+project archives: these are the downloadable artifacts that contain all the
+relevant data and metadata for a MarxanCloud project and its scenarios, which
+can subsequently be uploaded to a MarxanCloud instance in order to set up a
+copy of the downloaded project.
+
+## Project and scenario export and import flow - storage
+
+The preparation of a MarxanCloud project archive (see main [brief
+document](../brief.md)) is articulated through the preparation of several
+"pieces", each storing a coherent part of the data and metadata that make up a
+project or scenario.
+
+Whereas some metadata pieces may take up only a few hundred bytes, core project
+geo data may result in large pieces (in the GiB order of magnitude). Handling
+such files efficiently is therefore a key requirement during preparation,
+storage, download, upload and processing while creating a MarxanCloud project
+from an uploaded archive.
+
+In broad strokes, we need to account for:
+
+- storing one or more archive pieces as they are prepared and finalized
+
+For anything else than small metadata files, these pieces should be streamed to
+their eventual storage destination.
+
+- preparing a final archive file that includes all the finalized pieces
+
+This would typically be a zip (ZIP64) file. It may also be advisable to keep all
+the archive files alone and distribute them as a downloadable set of individual
+pieces instead, in order to avoid typical user challenges while handling very
+large files (e.g. > 4GiB) on personal computers.
+
+- storing the final archive
+
+This would be available for download for a limited timeframe, typically
+configured by the platform admins of each MarxanCloud instance.
+
+- allowing users to download the archive for a given project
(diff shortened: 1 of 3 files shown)


AaronPerezPerez (approved):

docs/download-upload-clone/storage/brief.md
```diff
@@ -0,0 +1,43 @@
...
+
+- preparing a final archive file that includes all the finalized pieces
+
+- storing the final archive
+
+- removing individual pieces
+
+- allowing users to download the archive for a given project
+
+- removing the archive and any related files once its validity expires
```
tiagojsag, line 35: I assume this would happen "long" after the export file generation ended. At this stage, shouldn't there be just one file - the final, zipped file?

hotzevzl: I left this open as I am not sure about this - I'd be actually open (maybe even keen?) to *not* archive all the pieces into a single, potentially humongous, zip file. The obvious benefit of a single zip is the user experience for non-tech users - get one file, handle one file...

On the other hand, especially in the case of very large zip files, we incur the extra operational "cost" (more pain than cost) of extracting our pieces from the zip file. Yes, we can uncompress and save individual files as we stream, but if anything goes wrong during this operation for a very large zip file, it's a pain. And then we hold locally, temporarily, the extracted pieces, which we could otherwise stream directly, already uncompressed/unpacked, from blob storage...

So yes, my preference so far would strongly lean on leaving the pieces in blob storage, and maybe even not bothering creating a single zip file for the end user.

docs/download-upload-clone/storage/high-level-design.md
```diff
@@ -0,0 +1,163 @@
...
+
+For anything else than small metadata files, these pieces should be streamed to
+their eventual storage destination.
+
+- preparing a final archive file that includes all the finalized pieces
+
+This would typically be a zip (ZIP64) file. It may also be advisable to keep all
+the archive files alone and distribute them as a downloadable set of individual
+pieces instead, in order to avoid typical user challenges while handling very
+large files (e.g. > 4GiB) on personal computers.
```
tiagojsag, line 117: I highly doubt that anyone working in this space has their first Fat32 limitation issue using MarxanCloud. NTFS has been the default for 20+ years now. I think it's time we move on.

docs/download-upload-clone/storage/high-level-design.md
```diff
@@ -0,0 +1,163 @@
...
+
+
+- storing one or more archive pieces as they are prepared and finalized
+
+For anything else than small metadata files, these pieces should be streamed to
+their eventual storage destination.
+
+- preparing a final archive file that includes all the finalized pieces
+
+This would typically be a zip (ZIP64) file. It may also be advisable to keep all
```
tiagojsag, line 114: I would ditch making the individual pieces available, at least for now. This is not only more work for us, but also potentially confusing for a user - what will you do with a single scenario piece, for example? I think this also slightly contradicts, in spirit, the checksum part, were we go out of our way to ensure the pieces are not manually modified.

docs/download-upload-clone/storage/high-level-design.md
```diff
@@ -0,0 +1,163 @@
...
+then uploaded by themselves or others as a new project, either directly from
+cloud storage or by first downloading it to their local computers), file size
+savings may be important only to some extent: typically when users want to store
+archives locally/offline; when sharing archives directly through their cloud
+storage locations, file size savings may be practically irrelevant as blob
+storage is cheap as well as - at the expected scale - data ingress and egress.
+
+Preference should be given, instead, to file formats (such as GIS file formats)
+that could be streamed while supporting native compression or other strategies
+to limit file sizes to a minimum.
```
tiagojsag, line 34: The file format/compression solution is not clear to me. I did some googling on the subject, and there seems to be a way to create a zip file without actually using compression, and I assume that's what you're going for here, but it's an assumption, so perhaps you could make it clearer as to how exactly this would work.

Another option/suggestion I have is to evaluate if there is any low compression rate, low cost compression algorithm we can use to get some compression for cheap. Optionally, we may even give the user the choice of using compression or not.

docs/download-upload-clone/storage/high-level-design.md
hotzevzl: I guess having worked at a IT helpdesk for some time made me irrationally over-afraid of what could go wrong on users desktops 😁 

I have no strong opinion about serving a single archive file - I used to consider this a given for data exports, but I'm not too sure anymore about its value, besides making it harder for users to accidentally delete pieces, if we distribute them alone. The 4GiB limitation is a bad example - indeed that specific legacy limitation would not be an issue in general (unless users copy stuff to a weirdly formatted usb thumb drive)

docs/download-upload-clone/storage/high-level-design.md
hotzevzl: so, I whereas I would slightly lean on giving users a single zip file, I still think that keeping the single pieces available on blob storage until their expiration (or deletion if we support explicitly deleting an export) would be a good idea: if each individual piece is streamed to blob storage while being prepared, then - unless the user wants to download the archive locally, in which case we can prepare a zip for them - we could leave the pieces there, and if the user wants to duplicate a project, we wouldn't have to deal with any archive ourselves, just stream in the pieces, use to populate tables/etc, and this could be done piece by piece, with dependencies honoured where some piece needs to get in before others can.

why do you think this would be more work for us?

docs/download-upload-clone/storage/high-level-design.md
hotzevzl: sure, I can clarify this in the text

the gist is, for project exports that don't need to be served to users (so only used for cloning without the intermediate artifact downloaded to a user's computer), compressing files shouldn't change much in the big scheme of things: yes, storing GeoJSON uncompressed is a huge waste of space, but disk and bandwidth are cheap, especially if there's no actual billable egress from a data centre but everything goes through a local VPC (again, for plain cloning, not when users download an archive artifact), and we can simply get the raw data and start processing it as it comes in.

I'd keep these two considerations separate - on one hand, combining all the pieces in a single file, on the other, whether to compress either single pieces, or a user-downloadable artifact, or both. I'll try to articulate this better in the brief.

tiagojsag:
@hotzevzl I feel like these separate discussions are best addressed as a single thread, because they overlap. Based on what I had in mind, and what you replied, I think this is the best solution:

- when exporting we do so using blob storage, with the exportId in the path (as you suggested)
- by default, we export the pieces as-is. We use compression for the formats that have it built-in, but otherwise don't compress data (for example, for json exports of sql data).
- Depending on user input, we may optionally grab the above blob "folder" and pack it into a zip file. Here we have two options:
  - We can pack data without compression - zip supports this, and I assume it will be very fast
  - We can evaluate different compression algorithms to see if we can find a quick one that will get relevant gains in file size without being too expensive to run.

So basically we ask the user if they want to just export, or export+download - and we do the export to blobs in both cases, and trigger an additional "packing into a single file" functionality in the second one.

In terms of data volume for the users, we could do one of two things:
- yolo it, and assume users have really big disks
- extract the data into multiple blobs, evaluate the size of that, and tell the user "the uncompressed volume is x GB. are you sure you want to download?" before packing the data into a single file. We could optimize this by adding a threshold below which the message would not be displayed.

Personally, I think "yolo" is the way to go. Downloading this sort of data is something that I think only technical users will use, and those don't scare easily when they see a multiple GB download - they are probably used to it.

hotzevzl:
> @hotzevzl I feel like these separate discussions are best addressed as a single thread, because they overlap. Based on what I had in mind, and what you replied, I think this is the best solution:
> [...]
> So basically we ask the user if they want to just export, or export+download - and we do the export to blobs in both cases, and trigger an additional "packing into a single file" functionality in the second one.

👍🏼 yep, that's basically what I had in mind - I wasn't very explicit in my writeup as I wanted to leave this open for feedback, so if this sounds ok to you too, I think we're all set. I will fill in the gaps/undecided bits in my original text to clarify.

relatedly, my thinking was that for the "zip all the things" I would suggest to use cloud functions - it's the kind of thing for which I wouldn't want to clog the "local" filesystem/volumes by bringing back the original pieces, running zip and uploading the result back to bucket... and also having to take care of retries (in case k8s needs to kill a pod that is midway through preparation of a zip archive, and anything else that can go Murphy-wrong). As much as I have tried to avoid cloud functions so far in this project, I think in this case they would make sense. Not sure if this should be described in this PR (in case it looks ok to you and other reviewers, in any case).

> In terms of data volume for the users, we could do one of two things:
> 
>     * yolo it, and assume users have really big disks
> 
>     * extract the data into multiple blobs, evaluate the size of that, and tell the user "the uncompressed volume is x GB. are you sure you want to download?" before packing the data into a single file. We could optimize this by adding a threshold below which the message would not be displayed.
> 
> Personally, I think "yolo" is the way to go. Downloading this sort of data is something that I think only technical users will use, and those don't scare easily when they see a multiple GB download - they are probably used to it.

+1 for yolo. let's keep this simple. SSDs are cheap.

docs/download-upload-clone/storage/high-level-design.md
```diff
@@ -0,0 +1,163 @@
...
+cases they may choose to first download a copy of the final archive: for
+example, for offline safeguarding/archiving, or to upload the archive to a
+different MarxanCloud instance.
+
+- checking the integrity of an uploaded archive
+
+Although some users may wish to edit project archives offline in order to tweak
+some of the data/metadata, MarxanCloud instances will prevent this by default in
+order to avoid data issues that may lead to uploading projects in an
+inconsistent state.
```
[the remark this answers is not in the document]
hotzevzl: yep, thanks! I'll add a short bit about an index file with checksums.

PR #776, opened 2022-01-24
Feat/marxan 1128 ensure acl for exporting/getting

angelhigueraacid:
### Feature relevant tickets

[Ensure ACL for exporting/getting](https://vizzuality.atlassian.net/browse/MARXAN-1128)
[System should allow to export public projects](https://vizzuality.atlassian.net/browse/MARXAN-891)

[MARXAN-1128]: https://vizzuality.atlassian.net/browse/MARXAN-1128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MARXAN-891]: https://vizzuality.atlassian.net/browse/MARXAN-891?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -41,6 +41,16 @@ export class ProjectAclService implements ProjectAccessControl {
     ProjectRoles.project_owner,
   ];
   private readonly canDeleteProjectRoles = [ProjectRoles.project_owner];
+  private readonly canExportProjectRoles = [
+    ProjectRoles.project_owner,
+    ProjectRoles.project_contributor,
+    ProjectRoles.project_viewer,
```
kgajowy, line 47: double-check this guys - as far as I remember, this blocks editing scenarios so it may be something that blocks contributors/owners

AaronPerezPerez: Would you mind to elaborate this? I don't really understand what do you mean 🤔 
These are just the roles on ACL allowed to request an export for a project

AaronPerezPerez: Do you mean to be more restrictive with the allowed roles?

kgajowy: just a generic thought @AaronPerezPerez - to reconsider who can export at all

hotzevzl: yep, we'll let viewers DDOS project/scenario editing, but I still think it's ok

api/apps/api/src/modules/projects/projects.service.ts
```diff
@@ -167,14 +170,26 @@ export class ProjectsService {
...
+    if (isLeft(isPublicProject)) return left(projectNotFound);
 
-    return (
-      await this.commandBus.execute(
-        new ExportProject(new ResourceId(projectId)),
-      )
-    ).value;
+    if (!userCanExportProject && !isPublicProject.right) {
+      return left(forbiddenError);
+    }
```
hotzevzl, line 187: I am not too keen on this - here we're folding out logic that makes two different checks to know if we can honour the request to create an export for the project, and also returning an error in case the project is not found.

Even though this would break the nice pattern we have of using `this.projectAclService.canBlahBlah()`, I would try to wrap the checks we do (user has permissions on a private project *or* the project is public) in a function that abstracts away the compound check - all we need to know at this stage is whether we should let the request go ahead or not, leaving the exact logic to a function that knows what needs to be done in order to check this.

api/apps/api/src/modules/projects/projects.service.ts
```diff
@@ -212,12 +227,27 @@ export class ProjectsService {
...
+      projectId,
     );
-    // MARXAN-1129
-    return location ? createReadStream(location.value) : null;
+    const isPublicProject = await this.projectChecker.isPublic(projectId);
+
+    if (isLeft(isPublicProject)) return left(projectNotFound);
+
+    if (!userIsAllowedToDownloadExport && !isPublicProject.right)
+      return left(notAllowed);
```
hotzevzl, line 249: same as in the other service function above

api/apps/api/test/project/get-export-file.e2e-spec.ts
```diff
@@ -0,0 +1,241 @@
...
+
+  await fixtures.WhenExportFileIsReady();
+  const response = await fixtures.WhenOwnerUserRequestExportFile();
+
+  fixtures.ThenFileIsDownloaded(response);
+});
+
+test('should permit downloading export file for contributor users ', async () => {
+  await fixtures.GivenProjectWasCreated();
+  await fixtures.GivenContributorWasAddedToProject();
```
hotzevzl, line 94: Minor, but maybe we can simplify this a bit for the checks that involve users with a role on the project - the fact that users with relevant roles have been granted access to the project should be more or less implicit when we check if a user with given role can download the export archive, so I would grant the roles to the users transparently within `GivenProjectWasCreated()`, or keep a separate step for this but grant all roles at once. Though again, I'd avoid exposing the role granting as a `given` step to avoid confusion - at least in my view: doing all this within the project setup step seems to work best, in my view.

api/apps/api/test/project/request-project-export.e2e-spec.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  const response = await fixtures.WhenOwnerUserRequestAnExport();
+
+  fixtures.ThenExportIsLaunched(response);
+});
+
+test('should permit export for contributor users', async () => {
+  await fixtures.GivenProjectWasCreated();
+
+  await fixtures.GivenContributorWasAddedToProject();
```
hotzevzl, line 55: same as in the tests for getting the export files.

api/apps/api/src/modules/projects/project-checker/marxan-project-checker.service.ts
```diff
@@ -85,6 +88,20 @@ export class MarxanProjectChecker implements ProjectChecker {
...
+
+    if (!project) {
+      return left(doesntExist);
+    }
+
+    const publicProject = await this.publishedProjectRepo.findOne(projectId);
+
+    return right(Boolean(publicProject));
+  }
+
```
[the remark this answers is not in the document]
AaronPerezPerez: As we understood, the table `published_projects` has rows of published projects, so if a project exists but there is no row in the other table, it is not public. If this is right, I dont really see correct to introduce the logic for checking wether a project exists or not in `publishedProjectRepof.indOne(...)`

api/apps/api/src/modules/projects/projects.service.ts
AaronPerezPerez: @hotzevzl Mmm I have two main doubts with this approach. Firstly, are we sure we want the ProjectACLService to check more than roles? By definition, it seems correct, but I don't know if on a quick read of the code I would know that inside it is checking for roles and that the project is published. By naming, I'd usually expect that ProjectChecker actually checks if the project if public. But if what I mention about ProjectACLService checking more than roles is not really an issue, we can refactor it quickly without problems! 😄

api/apps/api/test/project/get-export-file.e2e-spec.ts
AaronPerezPerez: Double checking this, having a step called `WhenINSERT_ROLEUserRequestExportFile` is more than enought 👌🏽

api/apps/api/src/modules/projects/projects.service.ts
hotzevzl: I'd lean towards keeping the ACL logic (whether this depends on roles or other conditions such as a project being published) in a single place, so I would favour wrapping the `isPublic()` check within `canExportProject()` (maybe we can name it `canExportProjectOrIsProjectPublic()` if we wanted to make things more explicit).

My main note about this bit here is actually about doing two checks (which could be fine) but also using the results of one of these checks to make sure the project actually exists.

Also right now we kind-of-duplicate this compound logic twice only, though as we add more features (cancelling a running export? deleting a still-valid export?, etc.) my feeling is that avoiding to unfold the logic to determine whether we can honour these requests in different service functions may help avoid accidental drifts on how we check these access control questions.

AaronPerezPerez: Nice, we'll take this approach, ensure before all that the project exists and moving the rest to ACL service

PR #785, opened 2022-01-26
MARXAN-1154-Filter-Deleted-users

rubvalave:
-Filters out deleted users when getting users on project/scenarios
-Prevent changes on user roles to deleted users in projects/scenarios.
-Adds tests for them.

Couple of notes:

The ACL check when deleting a user to check that no project was left without owners is not anywhere to be found due to a cyclical dependency problem that I left in a test branch as requested by @hotzevzl

### DIFF
--- a/api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
+++ b/api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
@@ -152,6 +152,7 @@ export class ProjectAclService implements ProjectAccessControl {
       .where({
         projectId,
       })
+      .andWhere('userId.isDeleted is false')
       .select([
         'users_projects.roleName',
         'userId.displayName',
--- a/api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
+++ b/api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
@@ -122,15 +122,21 @@ export class ProjectAclService implements ProjectAccessControl {
   }
 
   async hasOtherOwner(userId: string, projectId: string): Promise<Permit> {
-    const otherOwnersInProject = await this.roles.count({
-      where: {
+    const query = this.roles
+      .createQueryBuilder('users_projects')
+      .leftJoin('users_projects.user', 'userId')
+      .where({
         projectId,
         roleName: ProjectRoles.project_owner,
         userId: Not(userId),
-      },
-    });
+      })
+      .andWhere('userId.isDeleted is false');
+
+    const otherOwnersInProject = await query.getCount();
+
     return otherOwnersInProject >= 1;
   }
+
   /**
    * @debt This module should not involve user details and it should deal with
    * it using a standalone module that will access the data just to read it. We
--- a/api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
+++ b/api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
@@ -214,7 +214,11 @@ export class ProjectAclService implements ProjectAccessControl {
         .leftJoinAndSelect('users_projects.user', 'userId')
         .select(['users_projects.roleName', 'userId.isDeleted'])
         .getOne();
-
+      /**
+       * If a role was already granted to the user, but the user is marked
+       * as deleted, we don't want to touch their existing role: we consider it,
+       * for the time being, as an archived fact, kept untouched.
+       */
       if (existingUserInProject?.user?.isDeleted) {
         return left(transactionFailed);
       }
--- a/api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
+++ b/api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
@@ -205,16 +205,19 @@ export class ProjectAclService implements ProjectAccessControl {
     await apiQueryRunner.startTransaction();
 
     try {
-      const existingUserInProject = await apiQueryRunner.manager.findOne(
-        UsersProjectsApiEntity,
-        undefined,
-        {
-          where: {
-            projectId,
-            userId,
-          },
-        },
-      );
+      const existingUserInProject = await apiQueryRunner.manager
+        .createQueryBuilder(UsersProjectsApiEntity, 'users_projects')
+        .where({
+          projectId,
+          userId,
+        })
+        .leftJoinAndSelect('users_projects.user', 'userId')
+        .select(['users_projects.roleName', 'userId.isDeleted'])
+        .getOne();
+
+      if (existingUserInProject?.user?.isDeleted) {
+        return left(transactionFailed);
+      }
 
       if (!existingUserInProject) {
         const userRoleToSave = new UsersProjectsApiEntity();
--- a/api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.service.ts
+++ b/api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.service.ts
@@ -162,6 +162,7 @@ export class ScenarioAclService implements ScenarioAccessControl {
       .where({
         scenarioId,
       })
+      .andWhere('userId.isDeleted is false')
       .select([
         'users_scenarios.roleName',
         'userId.displayName',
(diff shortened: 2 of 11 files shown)


api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
```diff
@@ -198,16 +205,19 @@ export class ProjectAclService implements ProjectAccessControl {
...
-      );
+      const existingUserInProject = await apiQueryRunner.manager
+        .createQueryBuilder(UsersProjectsApiEntity, 'users_projects')
+        .where({
+          projectId,
+          userId,
+        })
+        .leftJoinAndSelect('users_projects.user', 'userId')
+        .select(['users_projects.roleName', 'userId.isDeleted'])
+        .getOne();
```
hotzevzl, line 216: another way to address this may be to make the "user has a role but user is deleted" case as an "anomaly" and only expose to the rest of the logic whether an active user already has a role, e.g.

```suggestion
      const isUserActiveAndHasAlreadyARoleInProject = await apiQueryRunner.manager
        .createQueryBuilder(UsersProjectsApiEntity, 'users_projects')
        .where({
          projectId,
          userId,
        })
        .leftJoinAndSelect('users_projects.user', 'userId')
        .select(['users_projects.roleName', 'userId.isDeleted'])
        .getOne()
        .then(user => {
          // handle the user is deleted case here
          // otherwise return true if the user is active and has a role, false if it doesn't have a role
        })
        .catch(...);
```

hotzevzl: btw don't feel like you have to go down this madness route 😁

api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.service.ts
```diff
@@ -210,16 +216,19 @@ export class ScenarioAclService implements ScenarioAccessControl {
...
+          scenarioId,
+          userId,
+        })
+        .leftJoinAndSelect('users_scenarios.user', 'userId')
+        .select(['users_scenarios.roleName', 'userId.isDeleted'])
+        .getOne();
+
+      if (existingUserInScenario?.user?.isDeleted) {
+        return left(transactionFailed);
+      }
```
hotzevzl, line 231: same as for projects, I suggest a kindness-to-future-selves note

api/apps/api/test/access-control/project-acl/project-acl-get.e2e-spec.ts
```diff
@@ -57,3 +57,14 @@ test(`getting project users as owner with a wrong search query`, async () => {
...
+
+test(`getting project users should only show those who are not deleted`, async () => {
+  const projectId = await fixtures.GivenProjectWasCreated();
+  await fixtures.GivenContributorWasAddedToProject(projectId);
+  await fixtures.GivenUserWasAddedToProject(projectId);
+  let response = await fixtures.WhenGettingProjectUsersAsOwner(projectId);
+  fixtures.ThenAllUsersBeforeDeletingAnyFromAppAreReturned(response);
+  await fixtures.GivenUserIsDeleted();
+  response = await fixtures.WhenGettingProjectUsersAsOwner(projectId);
+  fixtures.ThenAllUsersExceptDeletedAreReturned(response);
```
hotzevzl, line 69: minor, maybe we could split this test into two - one that tests the first `Then...` assertion in the current code, one that doesn't assert that but goes one deleting the user and asserting that `ThenAllUsersExceptDeletedAreReturned`.

api/apps/api/test/access-control/project-acl/project-acl-patch.e2e-spec.ts
```diff
@@ -106,10 +106,21 @@ test(`adds non-existent userId`, async () => {
...
+  fixtures.ThenQueryFailedIsReturned(nonExistentUserIdResponse);
 });
 
-test(`adds and deletes users alternately`, async () => {
+test(`changes user role after user is soft-deleted from the app`, async () => {
+  const projectId = await fixtures.GivenProjectWasCreated();
+  await fixtures.GivenUserWasAddedToProject(projectId);
+  await fixtures.GivenUserIsDeleted();
+
+  const response = await fixtures.WhenChangingUserRoleFromDeletedUser(
```
hotzevzl, line 117: maybe?
```suggestion
  const response = await fixtures.WhenChangingUserRoleForDeletedUser(
```

api/apps/api/test/access-control/scenario-acl/scenario-acl-get.e2e-spec.ts
```diff
@@ -57,3 +57,14 @@ test(`getting scenario users as owner with a wrong search query`, async () => {
...
+test(`getting scenario users should only show those who are not deleted`, async () => {
+  const scenarioId = await fixtures.GivenScenarioWasCreated();
+  await fixtures.GivenContributorWasAddedToScenario(scenarioId);
+  await fixtures.GivenUserWasAddedToScenario(scenarioId);
+  let response = await fixtures.WhenGettingScenarioUsersAsOwner(scenarioId);
+  fixtures.ThenAllUsersBeforeDeletingAnyFromAppAreReturned(response);
+  await fixtures.GivenUserIsDeleted();
+  response = await fixtures.WhenGettingScenarioUsersAsOwner(scenarioId);
+  fixtures.ThenAllUsersExceptDeletedAreReturned(response);
+});
```
hotzevzl, line 70: same as above for projects (up to you)

api/apps/api/test/access-control/scenario-acl/scenario-acl-patch.e2e-spec.ts
```diff
@@ -98,7 +98,18 @@ test(`adds non-existent userId`, async () => {
...
-  fixtures.ThenQueryFailedReturned(nonExistentUserIdResponse);
+  fixtures.ThenQueryFailedIsReturned(nonExistentUserIdResponse);
+});
+
+test(`changes user role after user is soft-deleted from the app`, async () => {
+  const scenarioId = await fixtures.GivenScenarioWasCreated();
+  await fixtures.GivenUserWasAddedToScenario(scenarioId);
+  await fixtures.GivenUserIsDeleted();
+
+  const response = await fixtures.WhenChangingUserRoleFromDeletedUser(
```
hotzevzl, line 109:
```suggestion
  const response = await fixtures.WhenChangingUserRoleForDeletedUser(
```

api/apps/api/test/steps/given-user-is-created.ts
```diff
@@ -0,0 +1,35 @@
+import { AccessToken } from '@marxan-api/modules/authentication/authentication.service';
+import { User } from '@marxan-api/modules/users/user.api.entity';
+import { INestApplication } from '@nestjs/common';
+import { getRepositoryToken } from '@nestjs/typeorm';
+import { hash } from 'bcrypt';
+import * as faker from 'faker';
+import * as request from 'supertest';
+import { Repository } from 'typeorm';
+
+export const GivenUserIsCreated = async (
```
hotzevzl, line 10: 👏🏼 I sense this is going to be the groundwork for long overdue and welcome changes to how we set up users for tests 😉

hotzevzl (approved):

api/apps/api/test/access-control/project-acl/project-acl-get.e2e-spec.ts
rubvalave: The first `Then` is actually there just as a check of the current users added to the project as expected for this test (1 owner, 1 contributor and 1 new added user). There is actually already a test to check for the regular `Getting Project Users As Owner` a little bit above this code.

In this case I specified it because I was creating a new user because of the principle of keeping the seeds as they are when starting the application (because I have to delete a user and deleting one of the 'aa' 'bb' etc users can interfere in other tests).

So long story short: first `Then` is already tested in a different way elsewhere and the truly important part is the last `Then` here.

api/apps/api/src/modules/access-control/projects-acl/project-acl.service.ts
rubvalave: By keeping `.select(['users_projects.roleName', 'userId.isDeleted']).getOne()` and then adding the `then()`would not it be already filtering out the deleted user?

I assume there is a mistake in your suggestion or I don't totally understand what you meant here.

hotzevzl: in the `.then()`, the parameter received should be the found `UsersProjectsApiEntity` (if found, and only the selected fields) - why would the result we get in the `.then()` be different than what we currently store in `existingUserInProject`?

but yes, let's not go down this madness route 😅

PR #788, opened 2022-01-27
MARXAN-1087-endpoint-acquire-scenarios-lock

rubvalave:
-Adds lock creation endpoint for scenarios.
-Adds tests for lock creation with different roles on a scenario.
-Minor fixes and improvements.

### DIFF
--- a/api/apps/api/src/modules/scenarios/locks/lock.service.spec.ts
+++ b/api/apps/api/src/modules/scenarios/locks/lock.service.spec.ts
@@ -4,12 +4,22 @@ import { Either, left, right } from 'fp-ts/Either';
 
 import { FixtureType } from '@marxan/utils/tests/fixture-type';
 import { AcquireFailure, lockedScenario, LockService } from './lock.service';
-import { ScenarioLockEntity } from './scenario.lock.api.entity';
+import { ScenarioLockEntity } from './entity/scenario.lock.api.entity';
 import { forbiddenError } from '@marxan-api/modules/access-control';
 import { ScenarioAccessControl } from '@marxan-api/modules/access-control/scenarios-acl/scenario-access-control';
 import { ScenarioAccessControlMock } from './__mocks__/scenario-access-control.mock';
+import { ScenarioLockDto } from './dto/scenario.lock.dto';
 
 let fixtures: FixtureType<typeof getFixtures>;
+interface LockInterface {
+  userId: string;
+  scenarioId: string;
+  createdAt: Date;
+}
+
+interface saveMethod {
+  save: () => LockInterface;
+}
 
 beforeEach(async () => {
   fixtures = await getFixtures();
@@ -72,7 +82,7 @@ async function getFixtures() {
         provide: getRepositoryToken(ScenarioLockEntity),
         useValue: {
           manager: {
-            transaction: (fn: (em: any) => Promise<void>) =>
+            transaction: (fn: (em: saveMethod) => Promise<void>) =>
               fn(mockEntityManager),
           },
           count: jest.fn(),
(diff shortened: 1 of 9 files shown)


api/apps/api/test/project-scenarios.e2e-spec.ts
```diff
@@ -325,6 +356,19 @@ async function getFixtures() {
...
+        .set('Authorization', `Bearer ${ownerToken}`),
+    WhenAcquiringLockForScenarioAsContributor: async () =>
+      await request(app.getHttpServer())
+        .post(`/api/v1/scenarios/${scenarioId}/lock`)
+        .set('Authorization', `Bearer ${contributorToken}`),
+    WhenAcquiringLockForScenarioAsViewer: async () =>
+      await request(app.getHttpServer())
+        .post(`/api/v1/scenarios/${scenarioId}/lock`)
+        .set('Authorization', `Bearer ${viewerToken}`),
+
```
hotzevzl, line 371: very minor, but to stick to the "rule of three" here I would suggest to extract the API call code to a `AcquireLockForScenarioAsUserWithToken(token)` and avoid the small duplication. But ok as is as well for me - it's test fixtures, not app code, and it's all code next to each other so quite easy to keep in sync in case any changes are ever needed.

api/apps/api/src/modules/scenarios/locks/scenario.lock.api.entity.ts
```diff
@@ -22,3 +23,8 @@ export class ScenarioLockEntity {
   })
   createdAt!: Date;
 }
+
+export class ScenarioLockResult {
+  @ApiProperty()
+  data!: ScenarioLockEntity;
+}
```
hotzevzl, line 30: I am not sure I get this - are you shaping the reponse as JSON:API? If so, that's great - but you'd need some more properties, and I would then assert more in the test method `ThenScenarioLockInfoIsReturned()` - at least that `data.type` matches what would be the JSON:API type for these lock response payloads.

If this is for something else, then ignore me (but I would not be clear what it is for.

If it's the JSON:API, have a look at other examples in bits such as scenario entity: examples from CRUD parts can help with this (or ask me if you get stuck).

api/apps/api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -1091,4 +1098,34 @@ export class ScenariosController {
...
+  ): Promise<ScenarioLockResult> {
+    const result = await this.lockService.acquireLock(id, req.user.id);
+
+    if (isLeft(result)) {
+      switch (result.left) {
+        case forbiddenError:
+          throw new ForbiddenException();
+        case lockedScenario:
+          throw new BadRequestException(
+            `Scenario ${id} is already being edited`,
```
hotzevzl, line 1120:
```suggestion
            `Scenario ${id} is already being edited.`,
```

Extremely minor and cosmetic only. I don't think we have a guideline on this nor clear unique convention across existing code, but I'd personally lean on using a full stop at the end of messages in general.

api/apps/api/src/modules/scenarios/locks/lock.service.spec.ts
```diff
@@ -64,13 +72,17 @@ async function getFixtures() {
         provide: getRepositoryToken(ScenarioLockEntity),
         useValue: {
           manager: {
-            transaction: (fn: (em: Partial<EntityManager>) => Promise<void>) =>
+            transaction: (fn: (em: any) => Promise<void>) =>
```
hotzevzl, line 75: I'm sure you have good reasons for this (including preserving your sanity against TS's type checker 😬), but switching to `any` always rings loud alerts in my brain both when I do it myself and when I do CRs... can we keep this more strict?

PR #796, opened 2022-02-02
feat(front): Protected areas new workflow[MARXAN-1241]

anamontiaga:
### Feature relevant tickets

[MARXAN-1241](https://vizzuality.atlassian.net/browse/MARXAN-1241)


anamontiaga:
So we have some problems with Protected Areas:

**API**
1 - To invalidate query and be able to update the protected areas after uploading a new custom one, we would need a new job “onUploadCustomProtectedArea” to invalidate when it succeed.

2 - We should need a PATCH instead a POST to be able to send only the threshold and not the areas + threshold. (or an other separately async job).

**FRONT**
1 - From some reason once the protected areas process is finish, the flow stands still on threshold subtab although saved subtab is planing-unit-preview

app/layout/scenarios/edit/name/component.tsx
```diff
@@ -61,7 +61,7 @@ export const ScenariosSidebarName: React.FC<ScenariosSidebarNameProps> = () => {
               solutions: 'empty',
             },
             tab: 'planning-unit',
-            subtab: 'planning-unit-preview',
+            subtab: 'pu-protected-areas-preview',
```
mbarrenechea, line 64: At the moment we create a scenario I think the `subtab` should be null

app/layout/scenarios/edit/wdpa/component.tsx
```diff
@@ -126,7 +126,7 @@ export const ScenariosSidebarEditWDPA: React.FC<ScenariosSidebarEditWDPAProps> =
               <ScenariosSidebarWDPACategories
                 onSuccess={() => {
                   setStep(1);
```
mbarrenechea, line 128: Remove `setStep`. We are going to handle everything by subtab

app/layout/scenarios/edit/wdpa/component.tsx
```diff
@@ -57,7 +57,7 @@ export const ScenariosSidebarEditWDPA: React.FC<ScenariosSidebarEditWDPAProps> =
   });
 
   useEffect(() => {
```
mbarrenechea, line 59: we don't need to handle steps anymore

mbarrenechea (changes requested):

mbarrenechea (approved):

PR #803, opened 2022-02-05
upsert when uploading same protected area for project [MARXAN-1245]

hotzevzl:
https://vizzuality.atlassian.net/browse/MARXAN-1245

### DIFF
--- a/api/apps/api/test/scenario-protected-areas/fixtures.ts
+++ b/api/apps/api/test/scenario-protected-areas/fixtures.ts
@@ -147,7 +147,10 @@ export const getFixtures = async () => {
           FROM (
                  SELECT json_array_elements($1::json -> 'features') AS features
                ) AS f
-          RETURNING id;
+          ON CONFLICT ON CONSTRAINT unique_custom_protected_area_geometries_per_project
+          DO UPDATE
+            SET full_name = EXCLUDED.full_name
+          RETURNING "id";
         `,
         [
           readFileSync(__dirname + `/nam-protected-area.geojson`),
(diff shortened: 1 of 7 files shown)


hotzevzl:
@aagm I am asking you to review this to validate the soundness of the db changes.

The original implementation used a stricter check:

```sql
create unique index wpdpa_project_geometries_project_check on wdpa(coalesce(wdpaid, -1), hash, coalesce(project_id, '00000000-0000-0000-0000-000000000000'))
```

However, at least when checking user uploads (that is, when `project_id is not null`, hence no `coalesce()` needed), my assumption is that the unique index above is overly restrictive, because if a user reuploads the exact same geometries, we don't really need to reject it and we can instead upsert, which will result in updating the protected area `full_name` (for this we need a full constraint - not a partial one - and without expressions - such as the `coalesce()` parts of the original unique index).

For ETL-imported protected areas, maybe we should keep a partial unique index (`where project_id is null`) such as `create unique index unique_wdpa_areas on wdpa(wdpaid, hash) where project_id is null;` - this isn't in my PR right now but it should restore all the strictness we had in the initial implementation while breaking it down in project-specific constraint (which we can use for the upsert proposed in this PR, improving the UX) and ETL-imported areas (where in my opinion we still want to fail hard in case we try to insert duplicated geometries with the same `wdpaid`, because we should then take a decision/check data/check pipeline/etc).

hotzevzl:
@aagm if CI and the Poltertests/Schrödingertests phenomena agree, from my point of view this should be ready to go - I have added the unique index on the WDPA areas which I mentioned in my previous comment, so with the latest changes in this PR we should get back the strictness of the original index, but now split into a unique constraint on project-specific areas only (which we use for the upsert) plus a unique index for the WDPA areas (which should avoid to get accidental duplicates in when importing WDPA areas).

PR #809, opened 2022-02-07
MARXAN-1236-releaseLock-ScenarioLockInACLmodule

rubvalave:
-Implements releaseLock endpoint for scenarios.
-Includes tests for various cases on locking/unlocking scenarios.

-Moves Scenario-Lock entity and service under scenarios-ACL module.
-Adds more context and proper conditions to release/unlock a scenario.
-Adds Project contributors and owners as roles that can unlock a scenario if needed.


api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.service.ts
```diff
@@ -161,6 +180,89 @@ export class ScenarioAclService implements ScenarioAccessControl {
...
+    scenarioId: string,
+  ): Promise<Either<typeof forbiddenError | AcquireFailure, ScenarioLockDto>> {
+    if (!(await this.canEditScenario(userId, scenarioId))) {
+      return left(forbiddenError);
+    }
+    const result = await this.lockService.acquireLock(scenarioId, userId);
+    if (isLeft(result)) {
+      return result;
+    }
+    return result;
```
tiagojsag, line 194: ??

api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.service.ts
```diff
@@ -161,6 +180,89 @@ export class ScenarioAclService implements ScenarioAccessControl {
...
+     * Using named variables instead of direct usage of functions
+     * for clarity when understanding the steps to release a lock
+     */
+    const scenarioIsLocked = await this.lockService.isLocked(scenarioId);
+    const canReleaseLockFromProjectLevel = await this.canReleaseLock(
+      userId,
+      projectId,
+    );
+    const canEditScenario = await this.canEditScenario(userId, scenarioId);
+    const scenarioIsLockedByUser = await this.lockService.isLockedByUser(
```
tiagojsag, line 217:
```suggestion
    const scenarioIsLockedByCurrentUser = await this.lockService.isLockedByUser(
```

api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.service.ts
```diff
@@ -161,6 +180,89 @@ export class ScenarioAclService implements ScenarioAccessControl {
...
+    if (!canEditScenario && !canReleaseLockFromProjectLevel) {
+      return left(forbiddenError);
+    }
+    if (!scenarioIsLocked) {
+      return left(noLockInPlace);
+    }
+    if (
+      !canReleaseLockFromProjectLevel &&
+      canEditScenario &&
+      !scenarioIsLockedByUser
```
tiagojsag, line 230:
```suggestion
      !scenarioIsLockedByCurrentUser
```

tiagojsag (changes requested):

api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.service.ts
rubvalave: The underneath method returns a monad, so in the first case result will be a left object and in the second case it will be a right object. It did confuse me at the beginning but I got so used to it that now I feel nothing.

api/apps/api/src/modules/scenarios/scenarios.service.ts
```diff
@@ -269,19 +273,25 @@ export class ScenariosService {
...
     });
     if (isLeft(scenario)) {
       return left(forbiddenError);
     }
-    if (!(await this.scenarioAclService.canEditScenario(userId, scenarioId))) {
-      return left(forbiddenError);
-    }
-    if (!(await this.lockService.isLockedByUser(scenarioId, userId))) {
-      return left(lockedScenario);
+    const aclAndLockLogic = await this.scenarioAclService.canEditScenarioAndLockLogicIsCorrect(
```
hotzevzl, line 289: see my earlier nitpick about the name of the function we call here - to combine nitpicking with nitpicking, I suggest to simplify the naming here as all we want to know is whether we can go ahead with the requested update... maybe
```suggestion
    const userCanEditScenario = await this.scenarioAclService.canEditScenarioAndLockLogicIsCorrect(
```

api/apps/api/src/modules/scenarios/scenarios.service.ts
```diff
@@ -947,4 +957,27 @@ export class ScenariosService {
...
+      typeof forbiddenError | typeof lockedByAnotherUser | typeof noLockInPlace,
+      void
+    >
+  > {
+    const scenario = await this.getById(scenarioId, {
+      authenticatedUser: { id: userId },
+    });
+    if (isLeft(scenario)) {
+      return left(forbiddenError);
+    }
```
hotzevzl, line 975: I am not too sure about this: `scenario`'s type is `Either<GetScenarioFailure | typeof forbiddenError, Scenario>`, but here we're compressing the possible left values to `forbiddenError` - I'd pass the upstream `left` through here.

api/apps/api/test/project-scenarios.e2e-spec.ts
```diff
@@ -126,59 +128,107 @@ describe('ScenariosModule (e2e)', () => {
...
 
+  it('Releases scenario lock correctly', async () => {
+    await fixtures.GivenScenarioWasCreated();
+    const userToken = await fixtures.GivenUserIsLoggedIn('owner');
+
+    let response = await fixtures.WhenAcquiringLockForScenarioAsOwner();
+    fixtures.ThenScenarioLockInfoForOwnerIsReturned(response);
+
+    response = await fixtures.WhenReleasingLockForScenario(userToken);
+    fixtures.ThenNoContentIsReturned(response);
```
hotzevzl, line 192: minor, but I find this slightly misleading - what we are asserting here is really that the operation is successful, so maybe
```suggestion
    fixtures.ThenLockIsSuccessfullyReleased(response);
```

hotzevzl: same for similar cases below

api/apps/api/test/project-scenarios.e2e-spec.ts
```diff
@@ -126,59 +128,107 @@ describe('ScenariosModule (e2e)', () => {
...
+
+  it('Releases scenario lock correctly because user is not owner of lock but is project owner', async () => {
+    await fixtures.GivenScenarioWasCreated();
+    await fixtures.GivenContributorWasAddedToScenario();
+    const userToken = await fixtures.GivenUserIsLoggedIn('owner');
+
+    let response = await fixtures.WhenAcquiringLockForScenarioAsContributor();
+    fixtures.ThenScenarioLockInfoForContributorIsReturned(response);
+
+    response = await fixtures.WhenReleasingLockForScenario(userToken);
```
hotzevzl, line 203: renaming this may make the intent even clearer (I'd say it's ok as is because things can be easily figured out by lifting one's eyes by a few lines of code), though we'd have to also refactor the function to avoid passing in the `userToken` (or it'd be actually worse):
```suggestion
    response = await fixtures.WhenReleasingLockForScenarioAsOwner();
```

hotzevzl: same below

api/apps/api/test/project-scenarios.e2e-spec.ts
```diff
@@ -126,59 +128,107 @@ describe('ScenariosModule (e2e)', () => {
...
+    const userToken = await fixtures.GivenUserIsLoggedIn('contributor');
+
+    let response = await fixtures.WhenAcquiringLockForScenarioAsOwner();
+    fixtures.ThenScenarioLockInfoForOwnerIsReturned(response);
+
+    response = await fixtures.WhenReleasingLockForScenario(userToken);
+    fixtures.ThenNoContentIsReturned(response);
+  });
+
+  it('Fails to release scenario lock as its not the same user and it is not a project owner/contributor', async () => {
```
hotzevzl, line 220:
```suggestion
  it('Fails to release scenario lock as it is not owned by the same user and this is not a project owner/contributor', async () => {
```

api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.service.ts
```diff
@@ -161,6 +180,87 @@ export class ScenarioAclService implements ScenarioAccessControl {
...
+    if (!scenarioIsLocked) {
+      return left(noLockInPlace);
+    }
+    if (
+      !canReleaseLockFromProjectLevel &&
+      canEditScenario &&
+      !scenarioIsLockedByCurrentUser
+    ) {
+      return left(lockedByAnotherUser);
+    }
```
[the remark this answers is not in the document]
hotzevzl: aaaaah, I had to read the test for this to realise that the compound check in your implementation was indeed correct 😅 

But then, I'd say, we can still skip checking `!canReleaseLockFromProjectLevel`, no? After all, as long as the user can edit the scenario, do we care if they would not be able to release the lock at project level?

Moreover, these rules for who can release a lock are of our own choice and I'd suggest that we may want to add a comment on top of this function to explain the rationale behind these choices (can be brief and high-level). I know, comments should not explain the _how_ but the _why_ and I'm kind of suggesting to stray on the how side of things by unfolding the choices behind the logic in the function itself.

Also, at least from my point of view what we are doing here is a partial "steal-lock" operation: a contributor or admin who doesn't own the lock can release it, and then possibly acquire a lock for themselves, but for this, wouldn't it be more ergonomic to add another method such as `stealLock()` that takes care of both steps (force-releasing someone else's lock and acquiring a lock for oneself)? These choices are somewhat down to preference, so I have no strong opinion here and in the interest of merging this PR quickly I'd say we can keep things as they are, and polish in a second step.

hotzevzl (approved):
despite the long list of comments, please take this as a substantive 👍🏼 so we can move this PR forward - see if you wish to incorporate any of the minor naming suggestions, but please do check the logical operators in `scenario-acl.service.ts` - that's the only essential bit for you to double check (`if (!canEditScenario && !canReleaseLockFromProjectLevel)`).

api/apps/api/src/modules/access-control/scenarios-acl/locks/__mocks__/scenario-access-control.mock.ts
```diff
@@ -0,0 +1,77 @@
...
+  async acquireLock(
+    userId: string,
+    scenarioId: string,
+  ): Promise<Either<typeof forbiddenError | AcquireFailure, ScenarioLockDto>> {
+    if (this.mock(userId, scenarioId)) {
+      return left(lockedScenario);
+    }
+    if (this.mock(userId, scenarioId)) {
+      return left(forbiddenError);
+    }
```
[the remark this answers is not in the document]
rubvalave: I think it is, but I thought I needed it for the mock to not complain about not returning one of the types of error. Apparently this does not work like that in the mock as I checked now and it does not complain so yes, it is redundant.

api/apps/api/src/modules/access-control/scenarios-acl/scenario-access-control.ts
```diff
@@ -11,4 +19,28 @@ export abstract class ScenarioAccessControl {
...
+    userId: string,
+    scenarioId: string,
+    projectId: string,
+  ): Promise<
+    Either<
+      typeof forbiddenError | typeof lockedByAnotherUser | typeof noLockInPlace,
+      void
+    >
+  >;
+  abstract canEditScenarioAndLockLogicIsCorrect(
```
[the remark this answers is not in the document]
rubvalave: I think I left this name because previously I was trying to attempt including the permissions of a Project Owner/Contributor in this function and I didn't rename it when I opted for leaving that out because it was getting quite confusing, but I think yeah your name is a better suggestion then.

api/apps/api/test/project-scenarios.e2e-spec.ts
rubvalave: Should it be enough if I rename the token as `ownerToken`? It's because I used a similar pattern on PR 815 for some other endpoints. I wanted to avoid what you pointed out in a previous PR of the repetition of having the same endpoint 3 times for each role inside an entity.

api/apps/api/src/modules/access-control/scenarios-acl/scenario-acl.service.ts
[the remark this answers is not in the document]
hotzevzl: sure. As above, let's keep this as is and maybe revisit later.

PR #814, opened 2022-02-09
MARXAN-1090-scenarioLockInfo

rubvalave:
-Adds GET endpoints `projects/:projectId/editing-locks` and `scenarios/:scenarioId/editing-locks` to check on scenario locks info.
-Refactors locks e2e tests into their own file and adds new tests for GET endpoints.


api/apps/api/src/modules/projects/projects.controller.ts
```diff
@@ -523,4 +523,22 @@ export class ProjectsController {
...
+  async findLocksForProject(
+    @Param('projectId', ParseUUIDPipe) projectId: string,
+    @Req() req: RequestWithAuthenticatedUser,
+  ): Promise<ScenarioLockResultPlural> {
+    const result = await this.projectsService.findAllLocks(
+      projectId,
+      req.user.id,
+    );
+    if (isLeft(result)) {
+      throw new ForbiddenException();
```
hotzevzl, line 539: i'm not sure this will cover (and keep covering, in case we add/change stuff upstream in the call stack) all the possible errors we may want to expose to API consumers, but on the other hand I'd not want to go down the bulldozer route of exhaustive checks here, given that this is a fairly unexciting corner of the API. Just an invitation to think whether we may need to handle other kinds of HTTP responses here.

hotzevzl: tl;dr I think we're covering all we need and all we can expect right now - if we add more failure cases upstream in the future, we can think about whether we need different HTTP error statuses

api/apps/api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -1221,4 +1224,19 @@ export class ScenariosController {
...
     }
   }
+
+  @Get(':scenarioId/editing-locks')
+  @ApiOperation({ summary: 'Get all locks' })
+  async findLocksForScenario(
+    @Param('scenarioId', ParseUUIDPipe) scenarioId: string,
+    @Req() req: RequestWithAuthenticatedUser,
+  ): Promise<ScenarioLockResultPlural> {
+    const result = await this.service.findAllLocks(scenarioId, req.user.id);
```
hotzevzl, line 1234: can we actually have more than one lock per scenario? IIRC we have a unique constraint at db level (but it may make sense to keep the payload structure similar between projects and scenarios, so in any case I don't think that returning an array in `data` here is an issue).

hotzevzl (approved):
well done for also being a good scout and tidying up the lock-related test files 🚀 

just see about the issue of giving viewers access to lock data, and I think this PR is done 😃 

and thanks for your patience through today's rebase mayhem 😬

api/apps/api/src/modules/scenarios/scenarios.controller.ts
rubvalave: I think I misunderstood something on the MARXAN-1090 comment you posted:

> This could be queried in two ways:
>
>`GET /api/v1/projects/:projectId/editing-locks`
>
>`GET /api/v1/scenarios/:scenarioId/editing-locks`
>
>In the first case, we look up the list of scenarios in the project by the project id and return the lock status of them all.
>
>In the second case, we look up the parent project of whatever scenarioId we get, then likewise return the lock status >of all the scenarios in the project (so all the sibling scenarios of the one that happened to be used to query this data). >Of course we could instead only return the lock metadata of the given scenario, if a typical use case is to check only >the current lock status of the scenario the user is currently editing (so we can ignore changes to locks of sibling >scenarios). Query time will be negligible in all cases, but if users have hundreds of scenarios :exploding_head: then >we could save on response payload sizes here.

As you can see, the second case (scenario) should also return all the scenarios in the project and that's why I was returning an array as well here in `data`.

tiagojsag (commented):
I approve because andrea approves :P

api/apps/api/src/modules/scenarios/scenarios.controller.ts
hotzevzl: lol, I had forgotten about my own evil plan. sorry.

so, great - functionally it's all good. Just see if you wish to make the evilness of the evil plan more explicit by renaming the functions involved in this "find on parent project by current scenario id".

api/apps/api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -1221,4 +1224,31 @@ export class ScenariosController {
...
```
hotzevzl, line 1234:
```suggestion
  @Get(':scenarioId/editing-locks')
  @ApiOperation({ summary: 'Get all locks for scenarios of this scenario\'s parent project' })
  async findLocksForScenariosWithinParentProject(
    @Param('scenarioId', ParseUUIDPipe) scenarioId: string,
    @Req() req: RequestWithAuthenticatedUser,
  ): Promise<ScenarioLockResultPlural> {
    const result = await this.service.findAllLocksForScenariosWithinParentProject(scenarioId, req.user.id);
```

hotzevzl (approved):

PR #817, opened 2022-02-10
Infrastructure - Initial TF setup, container registry and docker image build + push

tiagojsag:
This PR adds the basics of TF (remote state) and the docker image registry on Azure. It also has a k8s cluster, but that will be iterated on later, right now it's pretty much barebones.

It bumps the nodejs version on the FE app to 14.18 (to match the API, just because) and it adds a dockerfile to it too

In parallel, I removed all references to user ids and group ids from all docker images, because honestly those are probably a bad solution to whatever problem someone had and did not document - we have other projects running just fine without using these hacks.

Last but not least, GH actions to build and push docker images for the 3 apps to the azure container registry.


hotzevzl (approved):
🙌🏼 it's alive! or about to be, once these CI pipelines start chugging away 🚀 

with the caveat that I cannot say that it's mostly Greek to me only because I know (ancient) Greek and the non-TF part looks ok, let's get this chugging!

tiagojsag:
@hotzevzl any clue why tests are failing? too much output, and too many things that seem like expected errors, to the point where I cant tell apart what's expected and whats not

hotzevzl:
> @hotzevzl any clue why tests are failing? too much output, and too many things that seem like expected errors, to the point where I cant tell apart what's expected and whats not

It looks like the usual poltergeist tests (polterteists?) that we've been getting more of recently. I've set up a runner on a spare VM but i'm scared as hell of trying it out even only temporarily (i did remove the ability for passersby to open PRs, but still...)

hotzevzl:
polterteists passed on fourth attempt, let's MERGE ALL THE THINGS!

hotzevzl:
forgot to mention - not for this PR of course, but I suggest to complement the documentation in this PR with a separate bit listing the account/service principal/etc requirements that are needed in order to manage the infra via TF on Azure (=> which kind of least-privilege access needs to be granted in order for engineers to be able to tf apply all the things - will also be relevant for TNC to prepare things for Vizz to set up an environment on TNC's Azure account)

PR #820, opened 2022-02-10
front(feat): features new workflow[MARXAN-1260]

anamontiaga:
### Designs

[Figma Designs](https://www.figma.com/file/DpHfvKRBNQylK71c0nq2FD/Marxan-Visual_Internal?node-id=6717%3A10847)

### Feature relevant tickets
[MARXAN-1260](https://vizzuality.atlassian.net/browse/MARXAN-1260)


anamontiaga:
There is a problem when the intersection between features is made and the job is executed, apparently it is executed the geofeatureStratification job and we should need to be included on status. 

I think this is happening also in the previous version

mbarrenechea (approved):

PR #825, opened 2022-02-14
MARXAN-1261 Requested changes for scenario locks endpoints

rubvalave:
-Changes `PATCH /scenarios/:scenarioId/release-lock` into `DELETE /scenarios/:scenarioId/lock`.
-Changes `GET /scenarios/:scenarioId/editing-locks` payload. No more an array of every scenario lock of the parent project of the scenario. It returns either the current lock in place (`{userId, scenarioId, createdAt}`) or null if unlocked.
-Adapt e2e tests to endpoint changes.

### DIFF
--- a/api/apps/api/src/modules/scenarios/scenarios.controller.ts
+++ b/api/apps/api/src/modules/scenarios/scenarios.controller.ts
@@ -1189,15 +1189,15 @@ export class ScenariosController {
 
   @ApiOperation({
     description: `End Scenario Editing session.`,
-    summary: `Releases the lock created on a scenario so other
+    summary: `Deletes the lock created on a scenario so other
     users can start editing the scenario.`,
   })
   @ApiNoContentResponse({
     status: 204,
-    description: 'Lock was released correctly',
+    description: 'Lock was deleted correctly',
   })
   @HttpCode(HttpStatus.NO_CONTENT)
-  @Patch(`:id/release-lock`)
+  @Delete(`:id/lock`)
   async endScenarioEditingSession(
     @Param('id', ParseUUIDPipe) id: string,
     @Req() req: RequestWithAuthenticatedUser,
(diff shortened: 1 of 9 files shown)


api/apps/api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -1189,15 +1189,15 @@ export class ScenariosController {
...
   @ApiOperation({
     description: `End Scenario Editing session.`,
-    summary: `Releases the lock created on a scenario so other
+    summary: `Deletes the lock created on a scenario so other
     users can start editing the scenario.`,
   })
   @ApiNoContentResponse({
     status: 204,
-    description: 'Lock was released correctly',
+    description: 'Lock was deleted correctly',
```
hotzevzl, line 1197: Super minor: I'd say there's no need to rephrase this - we use `DELETE` because that's the verb we get from mighty HTTP, but in domain language, we still release the lock.

hotzevzl (approved):

api/apps/api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -1232,7 +1232,7 @@ export class ScenariosController {
   async findLocksForScenariosWithinParentProject(
     @Param('scenarioId', ParseUUIDPipe) scenarioId: string,
     @Req() req: RequestWithAuthenticatedUser,
-  ): Promise<null | ScenarioLockResultSingular> {
+  ): Promise<{ data: null } | ScenarioLockResultSingular> {
```
hotzevzl, line 1235: mh, what about adding to `ScenarioLockResultSingular` and `ScenarioLockResultPlural` the option for data to be undefined, instead? it's a small thing, but I'd try to avoid ad-hoc inline types if possible.

PR #831, opened 2022-02-15
MARXAN-1237-scenarioLocks-on-all-endpoints

rubvalave:
-Adds new lock logic check into every POST/PATCH/DELETE endpoint. Uses a flag for DELETE.
-Implements poor person's version of LOFU (Lock on first use)
-Adapts tests where needed for this new logic check, though with LOFU this would not be as needed as before.
-Refactors `aclErrorHandler` to address every type of error needed for `scenarios.controller` using exhaustive pattern (maybe the name should be changed into just  `errorHandler` or `exhaustiveErrorHandler`?).

### DIFF
--- a/api/apps/api/src/modules/scenarios/scenarios.controller.ts
+++ b/api/apps/api/src/modules/scenarios/scenarios.controller.ts
@@ -1,7 +1,5 @@
 import {
-  BadRequestException,
   Body,
-  ConflictException,
   Controller,
   Delete,
   ForbiddenException,
@@ -9,8 +7,6 @@ import {
   Header,
   HttpCode,
   HttpStatus,
-  InternalServerErrorException,
-  NotFoundException,
   Param,
   ParseUUIDPipe,
   Patch,
--- a/api/apps/api/src/modules/scenarios/scenarios.controller.ts
+++ b/api/apps/api/src/modules/scenarios/scenarios.controller.ts
@@ -63,11 +63,7 @@ import {
 import { UpdateScenarioPlanningUnitLockStatusDto } from './dto/update-scenario-planning-unit-lock-status.dto';
 import { ShapefileGeoJSONResponseDTO } from './dto/shapefile.geojson.response.dto';
 import { ApiConsumesShapefile } from '@marxan-api/decorators/shapefile.decorator';
-import {
-  projectDoesntExist,
-  projectNotReady,
-  ScenariosService,
-} from './scenarios.service';
+import { ScenariosService } from './scenarios.service';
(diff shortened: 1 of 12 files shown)


AaronPerezPerez (approved):

api/apps/api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -308,23 +270,10 @@ export class ScenariosController {
...
-          );
-        case blmCreationFailure:
-          throw new InternalServerErrorException(
-            `Could not create initial BLM for scenario`,
-          );
-
-        default:
-          throw new InternalServerErrorException();
-      }
+      return aclErrorHandler(result.left, {
```
hotzevzl, line 273: formally, `aclErrorHandler()`'s return type should be `Never`, so the `return` here (and in all the occurrences in this controller) should be redundant and slightly misleading to the reader of this piece of code.

another opinionated opinion here is that although `aclErrorHandler()` tidies up a lot of scattered error handling in one single neatly organized place, we end up throwing out of sight. I'm looking at this on GH so I'm likely to suggest some nonsense, but what about instead using `aclErrorHandler()` as an error translation layer (so it could be `mapAclDomainErrorToHttpError()`, maybe?), where each branch of the `switch` statement would *return* a `new HttpErrorOfSomeKind(options)`, which we would then `throw` in the controller functions?

so here specifically, this would be

```
throw mapAclDomainErrorToHttpError(result.left, {
  userId: req.user.id,
  resourceType: 'scenarios',
});
```

I am not super comfortable with this either, and I haven't really ever seen anything like this in JS or other languages, so I may just be hallucinating. The idea of a domain-to-http error translation layer sounds appealing to me in terms of readability/maintainability, but it may be just me 😬 

Also, `resourceType` may be set from `scenarioResource.name.plural`

hotzevzl (approved):

api/apps/api/src/modules/scenarios/scenarios.controller.ts
hotzevzl: well, since I had that nagging feeling that I was reinventing something from first principles, my brain went on a further neural flare and I eventually _remembered_ that NestJS gives us the lovely _exception filters_ functionality.

All is good and let's merge - but, maybe, maybe, the use of expressive errors via `Maybe`/`Either` could be pushed to the very end by throwing a generic `AclError` (or something like this) everywhere in these controllers, passing to it whatever `left` values we have there, and move the mapping from domain errors to HTTP exceptions to an exception filter.

Absolutely not _now_ of course, but I have the feeling this would be the ultimate logical evolution of the `Either` setup. Something to chew on 😄

PR #835, opened 2022-02-16
feat: complete import piece handler

angelhigueraacid:
## complete import piece handler

This PR adds: 

- `CompleteImportPieceHandler` which is executed when a `CompleteImportPiece` command is dispatched. `CompleteImportPieceHandler` completes the piece transactionally so there are not concurrency issues when completing the pieces of the import.

### DIFF
--- a/api/apps/api/src/modules/clone/import/application/complete-import-piece.handler.ts
+++ b/api/apps/api/src/modules/clone/import/application/complete-import-piece.handler.ts
@@ -28,14 +28,12 @@ export class CompleteImportPieceHandler
   }
 
   async execute({ importId, componentId }: CompleteImportPiece): Promise<void> {
-    let importInstance: Import | undefined;
     await this.importRepository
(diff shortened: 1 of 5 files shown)


api/apps/api/src/modules/clone/import/application/complete-import-piece.handler.ts
```diff
@@ -0,0 +1,72 @@
...
+              return;
+          }
+        }
+
+        await repo.save(importAggregate);
+
+        importAggregate.commit();
+      })
+      .catch(() => {
+        if (importInstance) {
```
AaronPerezPerez, line 67: This can be simplified, we don't need the instance to publish de event, and also we could only log the error in the .catch to clean the transaction.

AaronPerezPerez (changes requested):

aciddaute (approved):

aciddaute (approved):

PR #839, opened 2022-02-17
Feat/import archive command handler

angelhigueraacid:
## Feat/import archive command handler

This PR transforms `ImportArchive` into `ImportArchiveHandler`. Now when a new import is requested by a user the `ImportArchiveHandler` will be executed when a `ImportArchive` command is dispatched.


aciddaute (approved):

api/apps/api/src/modules/clone/import/application/import-archive.handler.spec.ts
```diff
@@ -66,117 +69,54 @@ const getFixtures = async () => {
...
-        ),
+      importResourcePieces.mockEqualPieces();
+    },
+    WhenRequestingImport: async () => {
+      const importResult = await sut.execute(
+        new ImportArchive(new ArchiveLocation(`whatever`)),
       );
+      if (isRight(importResult))
+        resourceId = new ResourceId(
+          repo.entities[(importResult as Right<string>).right].resourceId,
```
aciddaute, line 106: As it is already checked that `importResult` is right, we can remove `as Right<...>` 🤔  

```suggestion
          repo.entities[importResult.right].resourceId,
```

api/apps/api/src/modules/clone/import/adapters/archive-reader.adapter.ts
```diff
@@ -35,22 +25,6 @@ export class ArchiveReaderAdapter implements ArchiveReader {
       new RegExp(ClonePieceRelativePaths[ClonePiece.ExportConfig].config),
     );
     if (isLeft(exportConfigOrError)) return left(archiveCorrupted);
```
aciddaute, line 27: As I see it, extracting export config file logic should be placed outside `ArchiveReaderAdapter`, if we truly want to reuse `ArchiveReaderAdapter` we should return a `Readable`. Another thing that i've noticed is that the potential usage of this adapter will be mostly in geoprocessing service (in piece importers), maybe we should add it to `lib` package like `FileRepository` (port and temporary adapter).

PR #844, opened 2022-02-18
MARXAN-1091-lock-expiration-userToken

rubvalave:
-Adds migration to add a `token_id` column for  `lock` entity.
-Assigns expiration to lock based on expiration of token (on a ON CASCADE DELETE relationship between entities)
-Adds tests to check that lock is effectively removed when user token expires.

### DIFF
--- a/api/apps/api/src/migrations/api/1645026803969-AddTokenIdColumnToLocks.ts
+++ b/api/apps/api/src/migrations/api/1645026803969-AddTokenIdColumnToLocks.ts
@@ -1,0 +1,22 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddTokenIdColumnToLocks1645026803969
+  implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "scenario_locks"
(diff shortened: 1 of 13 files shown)


hotzevzl (approved):
I like this - small and elegant 🚀 

I had to think through the implications of linking a lock to the most-recently-issued (at the time of linking) token we have for an user, but I think with the current setup this should work.

Please do a last sanity check considering that issued tokens are not deleted from db immediately after expiration but only the next time _any_ user logs in *or* refreshes a token... so I think there may be some arbitrary time warp, but probably not something really relevant in normal use:

- user A logs in at 4AM (because, four in the morning...)
- then they do some stuff
- and acquire a scenario lock L for scenario S at 5:58, which is linked to the token T issued at 4AM and set to expire at 5:59:59AM
- and doze off right after
- user B logs in at 5:59AM
- scenario S should still be locked via lock L, because token T is still around even if expired
- any user log in or token refresh action after 6AM will still cause token T to be purged and with it the lock L

If my analysis is correct, I think we can live with this - user B (if they had ACL access to scenario S and wanted to work on it at 5:59AM) could still explicitly steal the lock if they have permissions to do so... otherwise... they will wait for any other user to log in, I suppose.

PR #847, opened 2022-02-18
PoC webshot service [MARXAN-1269]

hotzevzl:
https://vizzuality.atlassian.net/browse/MARXAN-1269

Adapted (and ported to TS) from MGIS' webshot service.

This is _not_ intended for production use, but _only_ to test the PDF generation workflow with the work currently being done within the app.

This PoC allows to point puppeteer to *any* url (to which, however, a templated path is appended) *and* to inject an arbitrary `cookie` header in the request to the eventual destination. This is of course *evil*, but should be fit for the intended purpose, that is, letting us `POST /api/v1/projects/:projectId/scenarios/:scenarioId/solutions/report` with a payload such as

```
{
    "baseUrl": "https://marxan-git-nw-marxan-828-report-vizzuality1.vercel.app",
    "cookie": "a __Secure-next-auth.session-token cookie grabbed from an authenticated session on marxan-git-nw-marxan-828-report-vizzuality1.vercel.app"
}
```

and get a PDF rendering of the solutions report for the given project+scenario.

Actually - is allowing to provide a cookie (and a JWT through it) _really evil? 🤔  We're passing through information that the user already has access too - cookies.

The main concern I can see here is that the cookie may *not* match the user which is logged in (so, the user whose JWT is presented for the API request for the PDF). So, given that basically we're passing through a JWT, wrapped in Next cookie wrapping paper, I'd generate the cookie in `WebshotService` from the JWT that has been presented. For this PoC however I'm going to leave this in as is - in case the flexibility of providing a _specific_ cookie is needed while frontend devs develop the interaction with this new API part. I'll welcome nasty comments on the vulnerabilities that this opens the floodgates to, and ideas on how to address the full authentication hop-around, if the 'assemble cookie from JWT' idea above doesn't sound good.

After tidying things up a bit from the very first commits, I started cursing myself for not having used NestJS for this, even if it's such a nano service - ok, Express is more than enough... but I miss the convenience of the NestJS way. At least this is a bit of a hidden service, and core validation of request parameters and body payload is done in the proxying controller handler of the API. But still...

Also, given that there is very little going on besides pointing the Puppeteer cannon to the target URL, I am not sure what would be wise to test here. I'll add ACL checks in a separate PR and that deserves its small suite of tests (in the API), but besides that... I'd hate to have to get into checking the generated PDF.


rubvalave (approved):

webshot/.gitignore
```diff
@@ -0,0 +1 @@
+dist
```
tiagojsag, line 1: any reason why this is here and not at the root level?

hotzevzl: we've kept the app and api gitignores distinct (without agreeing explicitly on this, but I guess it's an unspoken agreement to avoid bickering between FE and BE 😁 ) and I have kept this distinct for webshot too - I see you've added tf ignores in the root level one, and I agree with that - I'd stick to per-"application module" gitignores (and top level for infra and editor/IDE rubbish, which is more or less what is now, except a couple of python and node-specific excludes, though they're quite ubiquitous in any case so top level also makes sense), but if you prefer this could go as `/webshot/dist` in the top level gitignore too.

tiagojsag (approved):

PR #855, opened 2022-02-22
MARXAN-1020-platformAdmins

rubvalave:
-Adds migration to introduce platform admins table.
-Adds Endpoints to get, create and delete admins.
-Adds tests for admins, refactors user-e2e-tests to fit GWT pattern.

### DIFF
--- a/api/apps/api/src/modules/users/users.controller.ts
+++ b/api/apps/api/src/modules/users/users.controller.ts
@@ -177,7 +177,7 @@ export class UsersController {
     @Request() req: RequestWithAuthenticatedUser,
     @Param('id', ParseUUIDPipe) userIdToDelete: string,
   ): Promise<void> {
-    const result = await this.service.addAdmin(req.user.id, userIdToDelete);
+    const result = await this.service.deleteAdmin(req.user.id, userIdToDelete);
     if (isLeft(result)) {
       throw new ForbiddenException();
     }
--- a/api/apps/api/src/modules/users/users.controller.ts
+++ b/api/apps/api/src/modules/users/users.controller.ts
@@ -4,6 +4,7 @@ import {
   Delete,
   ForbiddenException,
   Get,
+  NotFoundException,
   Param,
   ParseUUIDPipe,
   Patch,
@@ -13,7 +14,11 @@ import {
   ValidationPipe,
 } from '@nestjs/common';
 import { User, userResource, UserResult } from './user.api.entity';
-import { UsersService } from './users.service';
+import {
+  forbiddenError,
+  userNotFoundError,
+  UsersService,
+} from './users.service';
 
 import {
   ApiBearerAuth,
@@ -133,7 +138,7 @@ export class UsersController {
   }
 
   @ApiOperation({
-    description: 'Get list of admins in platform.',
+    description: 'Get list of users with plaform admin role.',
   })
   @ApiOkResponse()
   @ApiUnauthorizedResponse()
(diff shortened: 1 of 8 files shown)


api/apps/api/src/modules/users/users.controller.ts
```diff
@@ -125,4 +131,55 @@ export class UsersController {
   ): Promise<void> {
     return this.service.markAsDeleted(req.user.id);
   }
+
+  @ApiOperation({
+    description: 'Get list of admins in platform.',
```
hotzevzl, line 136: maybe?
```suggestion
    description: 'Get list of users with platform admin role.',
```

api/apps/api/src/modules/users/users.controller.ts
```diff
@@ -125,4 +131,55 @@ export class UsersController {
...
+  ): Promise<PlatformAdminEntity[]> {
+    const result = await this.service.getPlatformAdmins(req.user.id);
+    if (isLeft(result)) {
+      throw new ForbiddenException();
+    }
+    return result.right;
+  }
+
+  @ApiOperation({
+    description: 'Add user as admin.',
```
hotzevzl, line 153: maybe (mirroring the existing operation description in the delete handler below
```suggestion
    description: 'Grant platform admin role to user.',
```

api/apps/api/src/modules/users/users.service.ts
```diff
@@ -216,4 +221,47 @@ export class UsersService extends AppBaseService<
...
+  ): Promise<Either<typeof forbiddenError, void>> {
+    return this.adminRepo.manager.transaction(async (entityManager) => {
+      if (!(await this.isPlatformAdmin(loggedUserId))) {
+        return left(forbiddenError);
+      }
+
+      const admin = new PlatformAdminEntity();
+      admin.userId = userId;
+
+      await entityManager.save(admin);
```
hotzevzl, line 250: by now we only know (because of `ParseUUIDPipe` that `admin.userId` is a UUID, but we don't know if there's any user with such `id`. we can let this fail (but we'd need to catch and handle the possible exception) or check validate the `userId` first.

personally, I'd let it fail and handle the failure, which _also_ helps in with Murphy's law (a user I am about to make a platform admin deletes their account in the very millisecond between when we check if such user exists and when we try to grant them platform admin role... yeah, as if...)

hotzevzl: also, why the transaction here? we only have one single `INSERT`, after all.

hotzevzl (approved):
thanks!

with the caveat that we should check the UUIDs that we get for users to whom we want to grant or from whom we want to revoke the platform admin role, it all looks in good order to me (and thanks for persevering through the test refactor dramas 😬)

An `isPlatformAdmin` flag on the user entity could have been enough too, but I don't see any issues with the route you chose.

As a separate concern (i.e. not in this PR, and this will require a bit of planning), it may be a good idea to either create a first platform admin user via a migration (and show in logs which initial password was generated for it) - right now we're granting this role to user `dd` mostly for tests, but I'd say that a "production" instance should not use these SQL seed for users, especially with hardcoded and very crackable passwords, and come "out of the box" with an initial platform admin user.

api/apps/api/src/modules/users/users.service.ts
rubvalave: Wouldn't the transaction protect the user from being deleted until this has passed? So when this is triggered, that nanosecond between deletion and addition should not matter because of the transaction?

hotzevzl: yes, you are right. if we get a request to delete the user by the time we've started this transaction, the user will only be deleted after the transaction is committed 🚀

PR #870, opened 2022-02-25
make puids stable across scenarios [MARXAN-1242]

hotzevzl:
https://vizzuality.atlassian.net/browse/MARXAN-1242

This PR specifically addresses the "stable" part of the story's requirements.

### DIFF
--- a/api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-linker-service.ts
+++ b/api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-linker-service.ts
@@ -171,6 +171,37 @@ export class ScenarioPlanningUnitsLinkerService {
     // upstream yet, and since the frontend app will (should) always send all
     // that is needed, this should be fine for a first pass.
     if (queryPartsForLinker) {
+      /**
+       * @debt Ordering `planning_units_geom` by uuid will ensure that every
+       * time a scenario is created from the same project, its planning units
+       * will be selected in the same order, therefore making the `puid`
+       * attribute of each planning unit stable across all the scenarios of the
+       * same project.
+       *
+       * This could be done more robustly by linking planning units _to
+       * projects_, so that puids can be assigned at that moment, and then
+       * propagated to individual scenarios. However, this is an intrusive
+       * change that will affect many parts of the API which should be avoided
+       * at this stage.
+       *
+       * Without this ordering (the initial implementation did not enforce an
+       * order), planning units would be selected in database order, and so
+       * their `puid` attribute _may_ be stable in most cases, except when any
+       * query updates relevant rows of `planning_units_geom`, which may result
+       * in the database order to be changed.
+       *
+       * Even with ordering, this setup relies on some assumptions that should
+       * be ignored in practice for the time being, such as that no updates to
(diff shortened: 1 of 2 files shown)


hotzevzl:
I have thought about adding an e2e test that creates two scenarios and checks that the resulting PUs in each of them all have the same `puid`, but... for the reasons explained in the inline comment in the changed bit of code, this test would have passed even before this change, so short of purposefully tampering with the data in the db, it will just test that PostgreSQL doesn't change row ordering randomly...

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units-linker-service.ts
```diff
@@ -171,11 +171,47 @@ export class ScenarioPlanningUnitsLinkerService {
...
+       * `planning_unit_geom` rows to be selected for scenarios created after a
+       * GADM geometries update. As ingesting GADM data is an ETL operation that
+       * is currently supposed to be done only once at the time a Marxan Cloud
+       * instance is set up, the logic below should still be sound, but it would
+       * need to be revisited if assumptions we rely on change.
+       */
+      const query = `
+                     with pug as (
+                       select * from planning_units_geom
+                       order by id
```
aciddaute, line 208: By ordering by `id` property we've stabilized puids across scenarios in the same Marxan instance but we need to ensure that those puids are equal in different Marxan instances in order to successfully import planning area grid. When planning area grid import piece has to be processed we only know old puid and geometry id for each planning unit. We need a mechanism to link the same exact geometry to the same data, if we order by the `id` property we cannot assume that the order (and therefore, the puid) of the geometries will be the same, because the `id` property of the geometries in the target Marxan instance will be automatically generated when importing planning units geometries. We've in mind two possible solutions:

1. Share geometries ids across different Marxan instances. When importing planning units geometries we could insert the same uuids for each record
2. Instead of ordering by id, we can order by `the_geom_hash` column, as it will be the same value across different Marxan instances

We can discuss later the drawbacks of these solutions and see if together we can find a better one

hotzevzl: excellent point. ULIDs or UUIDv7 could help here, but besides not being widely supported (nor even finalised, in the case of the UUIDv7 RFC), I can foresee trouble with subtle details related to timing of db insertion, etc.

I'd say we could order by hash, *or* go back to the idea of compiling a "master" list of PUs with their associated stable puid at the time of creation of a first scenario (`projects_pu`) and derive all the `scenarios_pu_data` for all the successive scenarios from this one (I'd be inclined towards this second option as the first one may still give us some trouble in the future if geoms are somehow adjusted 🤔)

aciddaute: I see the PUs list solution as the better one, but I think that those lists should be generated when a project is created instead of when the first scenario is created 🤔

aagm (approved):

hotzevzl:
closing as this was implemented fully through the recent project/scenario/pu linking refactor (https://vizzuality.atlassian.net/browse/MARXAN-1337)

PR #873, opened 2022-02-28
1.5x heap allocation for API service

hotzevzl:
https://vizzuality.atlassian.net/browse/MARXAN-1328


hotzevzl:
If the interwebs are correct (https://medium.com/geekculture/node-js-default-memory-settings-3c0fe8a9ba1), the default heap allocation should have been bumped to 4GiB from NodeJS v14 onwards, at least for systems with at least 16GiB physical memory: https://github.com/v8/v8/commit/b2f75b008d14fd1e1ef8579c9c4d2bc7d374efd3

GitHub Actions runners should be allocated with 7GiB memory, so the earlier 2GiB default should still apply. Bumping this to 6GiB as this PR does may be too much (as we also run other stuff on the same VM), so I'm open to feedback on this. Maybe I could see if a 3GiB allocation does solve our intermittent OOM issues when running tests for the API service, and be done with this if it kind of fixes things for the time being.

hotzevzl:
@alexeh I'm dragging you kicking and screaming in this PR because I have been told you are _the_ definitive v8 memory wizard in LG, so here we go.

I am not sure how many times I should push the "re-run all jobs" button in GH actions until we can be satisfied that this change actually "kinda fixes" the intermittent OOM issue we've been having with these tests... I'll keep pushing it for a while, in any case.

tiagojsag (approved):
MOAR MEMORY!!!!!!

PR #885, opened 2022-03-02
Marxan 1164 implicit permissions

rubvalave:
-Adds migration to include triggers on `users_projects` changes and `scenarios` creation.
-Tweaks some differences in tests due to not-slightly-accurate test data.
-Changes columns definition in `user-scenarios` API entity to avoid duplicate key errors.
-Tests for implicit-permissions. [WIP]

### DIFF
--- a/api/apps/api/src/migrations/api/1645550554581-ImplicitRolesFunctionsAndTriggers.ts
+++ b/api/apps/api/src/migrations/api/1645550554581-ImplicitRolesFunctionsAndTriggers.ts
@@ -109,53 +109,31 @@ export class ImplicitRolesFunctionsAndTriggers1645550554581
     `);
 
     await queryRunner.query(`
-      CREATE OR REPLACE FUNCTION change_implicit_scenario_role_to_explicit()
-        RETURNS trigger AS
-      $$
-      -- If we are handling an INSERT for a new explicit role that should
-        -- replace an existing implicit role, we simply flip the is_implicit
-        -- column to false
-      BEGIN
-        IF EXISTS (SELECT 1 FROM users_scenarios WHERE user_id = NEW.user_id AND scenario_id = NEW.scenario_id) THEN
-          IF TG_OP = 'INSERT' AND TG_TABLE_NAME = 'users_scenarios' AND NEW.is_implicit IS FALSE THEN
-            UPDATE users_scenarios SET
-              is_implicit = FALSE
-              role_id = NEW.role_id
-              WHERE user_id = NEW.user_id AND scenario_id = NEW.scenario_id;
-            RETURN NULL;
-          END IF;
-        END IF;
-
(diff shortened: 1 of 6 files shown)


api/apps/api/src/migrations/api/1645550554581-ImplicitRolesFunctionsAndTriggers.ts
```diff
@@ -0,0 +1,171 @@
...
+        END IF;
+      END;
+      $$ language 'plpgsql';
+    `);
+
+    await queryRunner.query(`
+      CREATE OR REPLACE FUNCTION manage_existing_project_roles_for_scenario()
+      RETURNS trigger as $$
+      DECLARE
+        users_to_grant role_on_entity[];
```
hotzevzl, line 115: mostly to be kind to our future selves - should we create (yet) another type, such as `CREATE TYPE user_and_role AS ( user_id uuid, role_id varchar);` and use this instead? the current implementation works just fine as the shape of data is the same, but whoever may need to read this in a few months may expel fewer WTFs if we reference `user_to_grant.user_id` rather than `user_to_grant.entity_id` in the body of this function.

api/apps/api/src/migrations/api/1645550554581-ImplicitRolesFunctionsAndTriggers.ts
```diff
@@ -0,0 +1,171 @@
...
+        upper_project_id uuid;
```
hotzevzl, line 116: nitpick: should we stick to _parent_ (rather than _upper_, which I don't think we've used much, if at all, in the rest of the codebase)?

api/apps/api/test/implicit-permissions/implicit-permissions.e2e-spec.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  await fixtures.WhenChangingAllUsersRole();
+  const response = await fixtures.WhenGettingScenarioUsers(
+    scenarioId,
+    creatorToken,
+  );
+  fixtures.ThenUsersWithChangedRolesShouldBeReturned(response);
+});
+
+test(`getting all locks of scenarios as scenario owner`, async () => {
```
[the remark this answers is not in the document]
rubvalave: I realized that a couple of tests do not run properly. I think there were some problems on how I approached the tests and I made some changes because maybe they were the reason.

I also introduced a DELETE FROM into the manage_existing_projects SQL trigger function, because when testing locally I found some weird messages appearing, but If you consider it is not necessary just delete it.

If tests pass and you're happy with them, I think this can be merged, as I don't want it to take any more time of my life. Locally I got it working by applying the actions manually. If tests keep failing maybe I overlooked something in how the fixtures are built or something.

PR #887, opened 2022-03-03
Feat/planning area grid custom piece importer

angelhigueraacid:
This PR adds planning area grid custom piece importer

### DIFF
--- a/api/apps/geoprocessing/src/import/pieces-importers/planning-area-custom-grid.piece-importer.ts
+++ b/api/apps/geoprocessing/src/import/pieces-importers/planning-area-custom-grid.piece-importer.ts
@@ -1,0 +1,132 @@
+import { geoprocessingConnections } from '@marxan-geoprocessing/ormconfig';
+import { ClonePiece, ImportJobInput, ImportJobOutput } from '@marxan/cloning';
+import {
+  ComponentLocationSnapshot,
+  ResourceKind,
+} from '@marxan/cloning/domain';
+import { FileRepository } from '@marxan/files-repository';
+import { PlanningUnitGridShape } from '@marxan/scenarios-planning-unit';
+import { Injectable, Logger } from '@nestjs/common';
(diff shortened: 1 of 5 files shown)


aciddaute (approved):

api/apps/geoprocessing/src/import/pieces-importers/planning-area-custom-grid.piece-importer.ts
```diff
@@ -0,0 +1,142 @@
...
+                  .map(this.parseGridFileLine);
+
+                const values = geomPUs
+                  .map((pu, index) => `($1,$2,ST_GeomFromEWKB($${index + 3}))`)
+                  .join(',');
+
+                const buffers = geomPUs.map((pu) => Buffer.from(pu.geom));
+                await this.geoprocessingEntityManager.query(
+                  `
+                  INSERT INTO planning_units_geom (type,project_id,the_geom)
```
hotzevzl, line 119: quick note on this and in general as you progress through imports of geo data which in some real cases may span tens of thousands of records, time a few columns: let's consider both PostgreSQL's limits on query parameters (IIRC depends on versions, but easy to hit at the 10k scale, I think) and the efficiency of TypeORM at composing queries with "high" (for some value of high" numbers of params.

My general suggestion would be to open a transaction (if not in one already) and split the inserts into batches considering the number of parameters needed to safely set the batch size according to PostgreSQL's limits.

I also suggest to always add a short note about this in a comment next to the batching logic, so that whoever may update the query next and for example add more params could take that into account and see if it may be necessary to reduce the batch length according (e.g. if doubling the number of params needed for each row, halve the batch size)

aciddaute: Wrapping the insert operations inside a transaction makes totally sense 👍 . However, as we are already working with streams the inserts are being made in batches (chunks) right?

PR #890, opened 2022-03-04
MARXAN-1124-add-users-to-project-workflow

rubvalave:
-Adds users endpoint to search for full email.
-Adds tests for this endpoint.

### DIFF
--- a/api/apps/api/src/modules/users/users.controller.ts
+++ b/api/apps/api/src/modules/users/users.controller.ts
@@ -36,7 +36,7 @@ import {
   FetchSpecification,
   ProcessFetchSpecification,
 } from 'nestjs-base-service';
-import { UpdateUserDTO } from './dto/update.user.dto';
+import { ExactEmailParamsDTO, UpdateUserDTO } from './dto/update.user.dto';
 import { UpdateUserPasswordDTO } from './dto/update.user-password';
 import { IsMissingAclImplementation } from '@marxan-api/decorators/acl.decorator';
 import { PlatformAdminEntity } from './platform-admin/admin.api.entity';
@@ -206,12 +206,11 @@ export class UsersController {
   @ApiOkResponse()
   @ApiUnauthorizedResponse()
   @ApiForbiddenResponse()
-  @Get('by-email')
+  @Get('by-email/:email')
   async findByEmail(
-    @Body() dto: UpdateUserDTO,
+    @Param() params: ExactEmailParamsDTO,
   ): Promise<UserResult | undefined> {
-    assertDefined(dto.email);
-    const result = await this.service.findByEmail(dto.email);
+    const result = await this.service.findByExactEmail(params.email);
     if (result) {
       return await this.service.serialize(result);
     }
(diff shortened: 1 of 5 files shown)


api/apps/api/src/modules/users/users.controller.ts
```diff
@@ -198,4 +199,21 @@ export class UsersController {
...
+
+  @ApiOperation({
+    description: 'Find users by full email address.',
+  })
+  @ApiOkResponse()
+  @ApiUnauthorizedResponse()
+  @ApiForbiddenResponse()
+  @Get('by-email')
+  async findByEmail(
+    @Body() dto: UpdateUserDTO,
```
hotzevzl, line 211: I would suggest to accept the email address as a `@Query()` (`?email=abc@example.com`) or `@Param()` (`/api/v1/users/by-email/abc@example.com`) - we shouldn't rely on a body in a GET request.

Also, and following from that, we would accept a plain string (in terms of TS typing), not a subset of a DTO such as UpdateUserDTO (in general, I'd recommend against repurposing a DTO like this, in any case - best to create a small ad-hoc DTO if really needed, or using tiny-types for expressivity).

api/apps/api/src/modules/users/users.service.ts
```diff
@@ -7,7 +7,7 @@ import {
 } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 import { ILike, Repository } from 'typeorm';
-import { User, userResource } from './user.api.entity';
+import { User, userResource, UserResult } from './user.api.entity';
```
hotzevzl, line 10: this is an unused import

api/apps/api/test/users.e2e-spec.ts
```diff
@@ -434,4 +434,44 @@ describe('UsersModule (e2e)', () => {
...
+      expect(
+        WhenSearchingUserByFullMail.body.data.attributes.displayName,
+      ).toEqual('User A A');
+    });
+
+    test('Partial email searches return undefined', async () => {
+      const WhenSearchingUserByFullMail = await request(app.getHttpServer())
+        .get('/api/v1/users/by-email')
+        .set('Authorization', `Bearer ${adminToken}`)
+        .send({ email: 'aa@example' });
```
hotzevzl, line 472: what if the API user sent us `{ email: 'aa@example%' }` though? 😉 

I suggest to create a new method such as `UsersService.findByExactEmail()`, or replace the existing `findByEmail()` method altogether, in either case dropping the `ILike` operator - basically all we should do in that service function is `'... where lower(email) = :email', { email: email.toLowerCase() }` or something like that.

api/apps/api/src/modules/users/users.controller.ts
hotzevzl: I would also add an `@IsEmail()` validator (this is also missing in my original `CreateUserDTO`, although it is present in `LoginDto` at least 🤦🏼)

hotzevzl (changes requested):

hotzevzl (approved):

PR #891, opened 2022-03-04
fix: add fake project id to planning area when uploading planning area

angelhigueraacid:
This PR fixes inconsistent `project_id` in `planning_areas` table value when uploading planning area. Now the id of the newly planning area and its associated project_id are the same.

### DIFF
--- a/api/apps/geoprocessing/test/integration/planning-areas/custom-planning-area.repository.e2e-spec.ts
+++ b/api/apps/geoprocessing/test/integration/planning-areas/custom-planning-area.repository.e2e-spec.ts
@@ -46,10 +46,20 @@ describe(`when entity created`, () => {
         38.20365531807149,
         29.22889003019423,
       ],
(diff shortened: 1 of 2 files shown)


hotzevzl (approved):

hotzevzl:
@angelhigueraacid please ignore my previous approval of this PR - I need to check it more thoroughly (and it's actually best for @aagm to have a close look too)

aciddaute (approved):

hotzevzl (commented):
@angelhigueraacid this looks good to me (just check the spelling mismatch I noted inline).

All works as expected when used via the frontend, including garbage collection - however, I need @aagm's input on this as I am not fully clear about what wasn't working as expected and why, since we did use this functionality in the past, and I am therefore not even sure where to check in case changing the assumptions of the original implementation may be causing some regression elsewhere.

aagm (approved):
Thank you!

PR #893, opened 2022-03-08
MARXAN-1361-block-users

rubvalave:
-Adds flag `isBlocked` for users table.
-Adds endpoint to block users in batches using array of Ids.
-Modifies instances where users can't be logged in if they are blocked/deleted.
-Adds tests for blocking endpoint.

### DIFF
--- a/api/apps/api/src/modules/users/users.service.ts
+++ b/api/apps/api/src/modules/users/users.service.ts
@@ -58,6 +58,7 @@ export class UsersService extends AppBaseService<
         'displayName',
         'avatarDataUrl',
         'isActive',
+        'isBlocked',
         'isDeleted',
         'metadata',
         'projects',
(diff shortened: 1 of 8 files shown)


api/apps/api/src/modules/users/users.service.ts
```diff
@@ -262,4 +273,20 @@ export class UsersService extends AppBaseService<
...
+  ): Promise<Either<typeof forbiddenError, void>> {
+    if (!(await this.isPlatformAdmin(loggedUserId))) {
+      return left(forbiddenError);
+    }
+
+    await Promise.all(
+      usersId.map(async (userId) => {
+        await this.markAsBlocked(userId);
+      }),
+    );
```
hotzevzl, line 289: as this is, an admin can lock themselves out (and possibly any admins, if they are the only one). For role deletions we enforce that the last standing owner of an entity cannot remove themselves - here the use case is a bit different and I would simplify this by rejecting (400 with a meaningful error message) if the admin is trying to block themselves, whether there would be other admins left standing or not - I don't see blocking oneself here as an action an admin would need to do in practice, so it would be enough to prevent that. What do you think?

hotzevzl: also, very much nitpicking: I'd call the argument `userIds`

hotzevzl: if going the "admin cannot block themselves" route, I'd add a test for that too - but up to you.

hotzevzl (approved):

PR #900, opened 2022-03-10
add brief and HLD for BLM preview maps

hotzevzl:
https://vizzuality.atlassian.net/browse/MARXAN-1277


rubvalave (approved):

hotzevzl:
@aagm we discussed originally the idea of "injecting" content on a template page via the webshot service - however I didn't have notes on the rationale for this, and the full workflow I have described in the HLD still seems ok to me (except the need to get a frontend auth token, which could be done via a cookie as we do for the PDF webshot module already)... do you remember which other advantages the "content injection" route would give us? Does the flow described here look ok to you?

PR #901, opened 2022-03-10
avoid possible fk constraint violations

hotzevzl:
https://vizzuality.atlassian.net/browse/MARXAN-1367

using BullMQ job ids to assemble `externalId` values for API events may lead to FK constraint violations as these job ids cannot be guaranteed to be forever monotonic.

This PR removes the use of `externalId` altogether for one kind of API events where this should not be needed, and adds `@todo` annotations to remove the dependency on job ids in other relevant code paths.

### DIFF
--- a/api/apps/api/src/modules/clone/infra/export/export-piece.events-handler.ts
+++ b/api/apps/api/src/modules/clone/infra/export/export-piece.events-handler.ts
@@ -57,7 +57,6 @@ export class ExportPieceEventsHandler
     return {
       topic: data.componentId,
       kind,
-      externalId: ApiEventsService.composeExternalId(data.componentId, kind),
       data: {
         kind,
         ...output,
@@ -80,7 +79,6 @@ export class ExportPieceEventsHandler
     return {
       topic: componentId,
       kind,
-      externalId: ApiEventsService.composeExternalId(componentId, kind),
       data: {
         exportId,
         resourceId,
--- a/api/apps/api/src/modules/clone/infra/export/mark-export-as-failed.handler.ts
+++ b/api/apps/api/src/modules/clone/infra/export/mark-export-as-failed.handler.ts
@@ -37,7 +37,6 @@ export class MarkExportAsFailedHandler
     await this.apiEvents.createIfNotExists({
       kind,
       topic: resourceId.value,
-      externalId: ApiEventsService.composeExternalId(exportId.value, kind),
       data: {
         exportId: exportId.value,
         resourceId: resourceId.value,
--- a/api/apps/api/src/modules/clone/infra/import/import-piece.events-handler.ts
+++ b/api/apps/api/src/modules/clone/infra/import/import-piece.events-handler.ts
@@ -51,7 +51,6 @@ export class ImportPieceEventsHandler
     return {
       topic: data.componentId,
       kind,
-      externalId: ApiEventsService.composeExternalId(data.componentId, kind),
       data: {
         kind,
         ...output,
@@ -75,7 +74,6 @@ export class ImportPieceEventsHandler
     return {
       topic: componentId,
       kind,
-      externalId: ApiEventsService.composeExternalId(componentId, kind),
       data: {
         importId,
         pieceResourceId,
--- a/api/apps/api/src/modules/geo-features/processing/run.service.ts
+++ b/api/apps/api/src/modules/geo-features/processing/run.service.ts
@@ -78,9 +78,6 @@ export class RunService {
   ) {
     const job = await queue.add(`run`, data);
     await this.apiEvents.create({
-      // @todo: remove dependency on job.id - may lead to FK constraint
-      // violations as job ids are not guaranteed to be forever monotonic
-      externalId: job.id + kind,
       kind,
       topic: data.scenarioId,
       data: {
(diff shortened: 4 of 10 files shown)


hotzevzl:
@aciddaute I would appreciate some help from the AT crew on this: I started this PR from the related Jira task (https://vizzuality.atlassian.net/browse/MARXAN-1367) to fix a potential issue with possible clashes of `externalId` values for the `API_EVENT_KINDS.scenario__protectedAreas__submitted__v1__alpha` events (see diff) but in general, the same issue may affect other parts of the codebase where we fire events that include a `externalId` based on sources that are not guaranteed to be monotonic or unique.

In most of the cases we assemble `externalId` from UUIDs (for example, all the instances that you and the AT crew added using  `ApiEventsService.composeExternalId()`, IIRC), however I would be keen to try to actually _remove_ dependencies on `externalId` being set where this is not strictly necessary.

For context, this property was introduced by Dominik (IIRC) when working on the geoprocessing of geofeatures (split, stratification and so on) in order to make sure we would not try to process events twice when they originate from external sources, combining a unique constraint on the column with handling conflicts via `ApiEventsService.createIfNotExists()`.

tl;dr I would like to make sure we try to use `externalId` only when strictly necessary for the purposes it was created for, and in any case that we don't rely on BullMQ job ids or other "external" ids that cannot be guaranteed not to lead to unnecessary constraint violations (because the duplication would come from a reset of the external source of "uniqueness", for example, as in the case with BullMQ job ids) - would you mind reviewing the uses of `externalId` you and the AT crew added, to confirm if they are all needed and guaranteed to be unique (unless we're *genuinely* trying to pick up an event from an external source more than once, for whatever reason)? I can take care of the rest.

aciddaute:
Sure! I'll review them 👍

aciddaute:
The only justified use of `externalId` property within the api events that we've worked with is in `MarkExportAsFailedHandler` ([here](https://github.com/Vizzuality/marxan-cloud/blob/0767a815f8d24e9d1ba135df81dc6ca4709155a9/api/apps/api/src/modules/clone/infra/export/mark-export-as-failed.handler.ts#L40)) . When a `export-piece` job fails a `ExportPieceFailed` event is emitted. `ExportPieceFailedSaga` handles that event and sends a `MarkExportAsFailed` command which is handled by `MarkExportAsFailedHandler`. In case `N` jobs fails before they are cancelled this command handler will be called `N` times, creating `N` "export-failed" api events. In any case, generated `externalId`s will be unique since they are composed by a uuid (export identifier).

List of unnecessary `externalId`s:

- [Export piece finished event in `ExportPieceEventsHandler`](https://github.com/Vizzuality/marxan-cloud/blob/0767a815f8d24e9d1ba135df81dc6ca4709155a9/api/apps/api/src/modules/clone/infra/export/export-piece.events-handler.ts#L60)
- [Export piece failed event in `ExportPieceEventsHandler`](https://github.com/Vizzuality/marxan-cloud/blob/0767a815f8d24e9d1ba135df81dc6ca4709155a9/api/apps/api/src/modules/clone/infra/export/export-piece.events-handler.ts#L83)
- [Export piece failed event in `MarkExportPiecesAsFailed` command handler](https://github.com/Vizzuality/marxan-cloud/blob/0767a815f8d24e9d1ba135df81dc6ca4709155a9/api/apps/api/src/modules/clone/infra/export/mark-export-pieces-as-failed.handler.ts#L42)
- [Import piece finished event in `ImportPieceEventsHandler`](https://github.com/Vizzuality/marxan-cloud/blob/0767a815f8d24e9d1ba135df81dc6ca4709155a9/api/apps/api/src/modules/clone/infra/import/import-piece.events-handler.ts#L56)
- [Import piece failed event in `ImportPieceEventsHandler`](https://github.com/Vizzuality/marxan-cloud/blob/0767a815f8d24e9d1ba135df81dc6ca4709155a9/api/apps/api/src/modules/clone/infra/import/import-piece.events-handler.ts#L80)
- [Blm calibration finished event in `BlmCalibrationEventsService`](https://github.com/Vizzuality/marxan-cloud/blob/0767a815f8d24e9d1ba135df81dc6ca4709155a9/api/apps/api/src/modules/scenarios/blm-calibration/blm-calibration-events.service.ts#L32)
- [Blm calibration failed event in `BlmCalibrationEventsService`](https://github.com/Vizzuality/marxan-cloud/blob/0767a815f8d24e9d1ba135df81dc6ca4709155a9/api/apps/api/src/modules/scenarios/blm-calibration/blm-calibration-events.service.ts#L43)
- [Blm calibration submitted event in `StartBlmCalibrationHandler `](https://github.com/Vizzuality/marxan-cloud/blob/0767a815f8d24e9d1ba135df81dc6ca4709155a9/api/apps/api/src/modules/scenarios/blm-calibration/start-blm-calibration.handler.ts#L49)

hotzevzl:
> The only justified use of `externalId` property within the api events that we've worked with is in `MarkExportAsFailedHandler` ([here](https://github.com/Vizzuality/marxan-cloud/blob/0767a815f8d24e9d1ba135df81dc6ca4709155a9/api/apps/api/src/modules/clone/infra/export/mark-export-as-failed.handler.ts#L40)) . When a `export-piece` job fails a `ExportPieceFailed` event is emitted. `ExportPieceFailedSaga` handles that event and sends a `MarkExportAsFailed` command which is handled by `MarkExportAsFailedHandler`. In case `N` jobs fails before they are cancelled this command handler will be called `N` times, creating `N` "export-failed" api events. In any case, generated `externalId`s will be unique since they are composed by a uuid (export identifier).
>
> List of unnecessary `externalId`s:

ok, understood - thanks for cross-checking! I'll remove them when I get back to this PR on Monday.

PR #911, opened 2022-03-17
document required postgresql extensions

hotzevzl:


tiagojsag (changes requested):
I think you need to take this into account: http://typeorm-doc.exceptionfound.com/interfaces/postgresconnectionoptions.html#uuidextension

example: http://typeorm-doc.exceptionfound.com/interfaces/postgresconnectionoptions.html#uuidextension

hotzevzl:
> I think you need to take this into account: http://typeorm-doc.exceptionfound.com/interfaces/postgresconnectionoptions.html#uuidextension
> 
> example: http://typeorm-doc.exceptionfound.com/interfaces/postgresconnectionoptions.html#uuidextension

👍🏼 ad516c644de16f018516c880d337045552ec84a0

And thanks to your comment, TIL that TypeORM would use this when a `@Column()` is decorated with `@Generated('uuid')` (but not for columns annotated as  `@PrimaryGeneratedColumn('uuid')` or `@PrimaryColumn('uuid')`, apparently - here the generated SQL uses the `DEFAULT` value), except that it doesn't even work like that, in practice... but well, I have given up with TypeORM's documentation a long time ago 😢

PR #913, opened 2022-03-17
MARXAN-1380-unpublish-public-project

rubvalave:
-Adds endpoint for `project_owners` to unpublish a public project.
-Adds conditions to when clearing a project from under moderation to be able to outright unpublish the project straightaway.
-Adds tests for the new endpoint and changes in the workflow.

Depends on #896 being closed beforehand.

### DIFF
--- a/api/apps/api/src/modules/published-project/published-project-crud.service.ts
+++ b/api/apps/api/src/modules/published-project/published-project-crud.service.ts
@@ -48,15 +48,16 @@ export class PublishedProjectCrudService extends AppBaseService<
     fetchSpecification: FetchSpecification,
     info?: ProjectsRequest,
   ): Promise<SelectQueryBuilder<PublishedProject>> {
-    let showUnderModerationPublishedProjects = false;
     const id = info?.authenticatedUser?.id;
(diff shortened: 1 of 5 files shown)


api/apps/api/src/modules/published-project/published-project-crud.service.ts
```diff
@@ -48,15 +48,16 @@ export class PublishedProjectCrudService extends AppBaseService<
...
 
-    query.andWhere('published_project.underModeration = :underModeration', {
-      underModeration: showUnderModerationPublishedProjects,
-    });
+    /*
+      If we are listing projects for non-authenticated requests or for
+      authenticated users who are not admin, projects under moderation
+      will be hiding from the listing.
+    */
+    if (!id || !(await this.usersService.isPlatformAdmin(id))) {
```
hotzevzl, line 58: while we are revisiting this, let's make the name more explicit
```suggestion
    if (!userId || !(await this.usersService.isPlatformAdmin(userId))) {
```

hotzevzl (approved):
thanks! IIRC you asked for naming suggestions on a different PR but I ended up suggesting a couple of naming changes here too... feel free to decide.

all the rest looks good!

PR #920, opened 2022-03-21
document NETWORK_CORS_ORIGINS env var + set it for k8s api container

hotzevzl:
The TF part is me doing something I have absolutely no idea about.


tiagojsag (commented):
your previous TF changes didn't make a lot of sense, as they affected unrelated modules, and ultimately broke things. My commit should hopefully fix that. Please review and merge

PR #921, opened 2022-03-21
Ref/planning units creation

aciddaute:
This PR refactors projects planning units creation process. `project_id` column of `projects_pu` table no longer contains planning area id either way if the project is being created from a shapefile or the project is being created with a regular grid:

- Added `planning_area_id` column to `projects_pu`
- Now `SetProjectGridFromShapefileHandler` updates `projects_pu` `project_id` column based on `planning_area_id` value
- Now `PlanningAreaGarbageCollector` deletes `projects_pu` filtering by `planning_area_id` instead of `project_id` column
- Now if the project has a regular planning units grid and a custom planning area, `projects_pu` records will be inserted with values in both columns (`project_id`, `planning_area_id`). If the project has a gadm planning area `planning_area_id` columns will contain null values.

### DIFF
--- a/api/apps/api/src/modules/projects/planning-unit-grid/set-project-grid-from-shapefile.handler.ts
+++ b/api/apps/api/src/modules/projects/planning-unit-grid/set-project-grid-from-shapefile.handler.ts
@@ -1,3 +1,11 @@
+import { ApiEventsService } from '@marxan-api/modules/api-events';
+import { Project } from '@marxan-api/modules/projects/project.api.entity';
+import { DbConnections } from '@marxan-api/ormconfig.connections';
+import { ProjectsPuEntity } from '@marxan-jobs/planning-unit-geometry';
+import { API_EVENT_KINDS } from '@marxan/api-events';
+import { PlanningArea } from '@marxan/planning-area-repository/planning-area.geo.entity';
+import { PlanningUnitSet } from '@marxan/planning-units-grid';
+import { PlanningUnitGridShape as ProjectGridShape } from '@marxan/scenarios-planning-unit';
 import {
   CommandHandler,
   EventBus,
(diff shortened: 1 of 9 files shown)


angelhigueraacid (approved):

api/apps/api/src/modules/projects/planning-unit-grid/set-project-grid-from-shapefile.handler.ts
```diff
@@ -42,23 +42,12 @@ export class SetProjectGridFromShapefileHandler
...
-
-      await manager.query(
-        `
-          UPDATE "planning_areas"
-          SET "project_id" = $1
-          WHERE "project_id" = $2
-        `,
-        [projectId, planningAreaId],
-      );
+      await manager
```
AaronPerezPerez, line 45: 👏🏽

api/apps/geoprocessing/src/migrations/geoprocessing/1647859641978-AddProjectsPuPlanningAreaIdColumn.ts
```diff
@@ -0,0 +1,27 @@
...
+      ALTER COLUMN project_id DROP NOT NULL;
+    `);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+      ALTER TABLE projects_pu
+      DROP COLUMN planning_area_id;
+    `);
+
```
AaronPerezPerez, line 21: At the moment we couldn't rollback this transaction as it would fail with the PUs that have null project_ids. To bypass this error I would add a default fake project uuid like "0000-0000..."

AaronPerezPerez (changes requested):

AaronPerezPerez (approved):

PR #923, opened 2022-03-21
Marxan 1374

rubvalave:
-Creates migration to add a protected_pu_ids array to partial results.
-Adds new service to handle output_best.csv information per run.
-Adds unit tests for parser of output_best.

@hotzevzl I don't know if we should keep this PR open till the rest of the work is done? and make smaller PR's to this one? It's because I don't know if something else will be broken if this is merged (I guess not but just in case)

Also, naming of things is based on my understanding on what everything does, but feel free to suggest better namings!


hotzevzl (approved):
@rubvalave looks great. Could you please rebase on `develop` and merge?

PR #929, opened 2022-03-24
MARXAN-1374-webshot-as-library

rubvalave:
-Removes webshot service from inside the api and moves it to a library under `libs` folder.
-Adds endpoint to ask for PDF scenario report to scenario-controller.
-Adds new webshot config folder to different configs (jest-e2e.json, package.json, tsconfig).


api/apps/api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -1215,4 +1220,59 @@ export class ScenariosController {
...
+        userId: req.user.id,
+        resourceType: scenarioResource.name.plural,
+      });
+    }
+
+    const pdfStream = await this.webshotService.getSummaryReportForScenario(
+      scenarioId,
+      scenario.right.projectId,
+      configForWebshot,
+    );
```
hotzevzl, line 1268: what do you think about moving this into a small function in `ScenariosService`? this way we avoid adding ACL logic in the controller (though I think there will be a hopefully manageable circular dependency lurking if we do so)

api/apps/api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -1215,4 +1220,59 @@ export class ScenariosController {
...
+     */
+    const configForWebshot = appSessionToken
+      ? {
+          ...config,
+          cookie: appSessionToken,
+        }
+      : config;
+    // @debt Refactor to use @nestjs/common's StreamableFile
+    // (https://docs.nestjs.com/techniques/streaming-files#streamable-file-class)
+    // after upgrading NestJS to v8.
```
hotzevzl, line 1251: for sanity of our future selves reading this in months from now, this comment belongs just above the stream part below

api/apps/api/src/modules/scenarios/scenarios.controller.ts
hotzevzl: I mean, at least the `getById()` and `if(isLeft(scenario))` bits - the pdfstream thingy can stay, but basically it would use `this.scenariosService` which would pass arguments through to `webshotService`, and in turn pass back the stream from `webshotService`.

hotzevzl: ok, I see you've dealt with circular deps between scenarios service and webshot service already below

hotzevzl (approved):
awesome, thank you!

I would only ask you to think about moving the acl checks out of the controller handler as I suggest in the inline comments, and otherwise this looks ok to go.

(please rebase on `develop` though - this PR includes changes from the first BLM preview PR)

api/apps/api/src/modules/scenarios/scenarios.controller.ts
hotzevzl: as discussed, I've done this - 5175ac5aa66ad5923dea5aafba7322fe5a72c4af

api/apps/api/src/modules/scenarios/scenarios.controller.ts
hotzevzl: 2208771b4c59372c2a5147b18156b7d3831414f6

PR #939, opened 2022-03-29
proposal: refactor project scenario feature linking [MARXAN-1405]

hotzevzl:
https://vizzuality.atlassian.net/browse/MARXAN-1405


hotzevzl:
@aagm @AaronPerezPerez @aciddaute @angelhigueraacid before my brain melts fully on this - here is a draft of the refactor (more of an overlay, let's say - I tried to touch as little as possible of the current data architecture in order to minimize risks)...

I had started with a detailed implementation suggestion to make sure I would address all the requirements, edge cases, caveats and so on, but since this was taking me forever, I tried today to summarise the key points in the _Implementation - summary_ section (only for user-uploaded features, going to add bits for platform features now). You can read through from the beginning up to the  _Implementation - summary_ section, and ignore all that comes after that.

I have dropped some ideas I have explored and tested along the way (for example, a monotonic bigint id), which are still in the _Implementation (full rationale)_ section - just ignore that final madness, please.

hotzevzl:
@aagm one thing I left undefined on purpose is what `hash` is - from a even a small sample of data, I think that that through `<feature_class_name>` and `hash(the_geom)` we *cannot* unequivocally identify `features_data` rows (we would have the `feature_class_name` of their "parent feature" easily by processing features by going through the scenario-relevant `(apidb)features` one by one), as this doesn't address the case where `features_data.the_geom` is not unique across all the `features_data` rows for a given `(apidb)feature` - this is the case in my local env, for example:

```
marxan-geo-api=# select distinct md5(the_geom::bytea), feature_id, count(*) from features_data group by md5(the_geom::bytea), feature_id order by count desc;

               md5                |              feature_id              | count
----------------------------------+--------------------------------------+-------
 fccd47ef31cf55e3f8553da341f3c5ed | d180c56b-2747-4e82-b292-51adab26b4a6 |     6
 03c236657ded2e2a3a3f55b911d4bb7f | d180c56b-2747-4e82-b292-51adab26b4a6 |     5
 2771d1d635e1943691e318ff4fea1cc9 | d180c56b-2747-4e82-b292-51adab26b4a6 |     5
 5010a325850f732bbbe6c994e4318dc9 | d180c56b-2747-4e82-b292-51adab26b4a6 |     5
 b5f40873a81ce718e0c0f5550ac7779e | d180c56b-2747-4e82-b292-51adab26b4a6 |     5
 bb6c7d399e8dba20de0ad57b3b211ba4 | d180c56b-2747-4e82-b292-51adab26b4a6 |     5
 04ba6937b97a94307ae6da932eb7cfc5 | d180c56b-2747-4e82-b292-51adab26b4a6 |     4
 1057a86be8a0b8b5572b5e647eacbac9 | d180c56b-2747-4e82-b292-51adab26b4a6 |     4
 2f10ccda9ff9dcf360a37b79d2e74894 | d180c56b-2747-4e82-b292-51adab26b4a6 |     4
 439c4b06f354cc1f5897b698ad294275 | d180c56b-2747-4e82-b292-51adab26b4a6 |     4
 6461a8d6f51cb4deee5cc0656ad766bd | d180c56b-2747-4e82-b292-51adab26b4a6 |     4
[...]
```

I thought about using `(md5(concat(the_geom::bytea, properties)))` but `concat()` is not `immutable` so it cannot be used as is for a generated column, so that's not an option without risking a huge act of hybris such as wrapping concat in an immutable function... even when knowing that we would never feed it content that may be expanded non-deterministically... though something like this may be needed eventually 🤷🏼

hotzevzl:
> [...]
> I thought about using `(md5(concat(the_geom::bytea, properties)))` but `concat()` is not `immutable` so it cannot be used as is for a generated column, so that's not an option without risking a huge act of hybris such as wrapping concat in an immutable function... even when knowing that we would never feed it content that may be expanded non-deterministically... though something like this may be needed eventually 🤷🏼

ok, stepping back from this and some fresh air helped 😄 - this can work, I think:

```
alter table features_data add column hash text not null generated always as (md5(the_geom::bytea::text || properties::text)) stored;
create unique index unique_feature_data_per_project on features_data(hash, feature_id);
```

I'm happy 😄

PR #940, opened 2022-03-29
fix: throws a bad request when giving a create project dto with regular pu shape, without GADM id or country id

AaronPerezPerez:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/test/project-planning-areas-tiles/world.ts
+++ b/api/apps/api/test/project-planning-areas-tiles/world.ts
@@ -27,7 +27,7 @@ export const createWorld = async (app: INestApplication) => {
 
   const regularProject = await ProjectsTestUtils.createProject(app, jwtToken, {
     ...E2E_CONFIG.projects.valid.minimalInGivenAdminArea({
-      countryCode: 'NAM',
+      countryId: 'NAM',
       adminAreaLevel1Id: 'NAM.8_1',
       adminAreaLevel2Id: 'NAM.8.6_1',
     }),
--- a/api/apps/api/test/project/projects-world.ts
+++ b/api/apps/api/test/project/projects-world.ts
@@ -23,7 +23,7 @@ export const createWorld = async (app: INestApplication, jwtToken: string) => {
   const projectWithCountry = (
     await ProjectsTestUtils.createProject(app, jwtToken, {
       ...E2E_CONFIG.projects.valid.minimalInGivenAdminArea({
-        countryCode: country,
+        countryId: country,
       }),
       organizationId,
     })
@@ -32,7 +32,7 @@ export const createWorld = async (app: INestApplication, jwtToken: string) => {
   const projectWithGid1 = (
     await ProjectsTestUtils.createProject(app, jwtToken, {
       ...E2E_CONFIG.projects.valid.minimalInGivenAdminArea({
-        countryCode: country,
+        countryId: country,
         adminAreaLevel1Id: l1AdminArea,
       }),
       organizationId,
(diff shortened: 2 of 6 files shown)


aciddaute (approved):

api/apps/api/test/e2e.config.ts
```diff
@@ -118,7 +119,7 @@ export const E2E_CONFIG: {
       minimal: () => ({
         name: faker.random.words(5),
         organizationId: faker.random.uuid(),
-        planningUnitGridShape: PlanningUnitGridShape.Hexagon,
+        planningUnitGridShape: PlanningUnitGridShape.Irregular,
```
hotzevzl, line 122: 🤔 even if this _happens not to interfere_ with other tests, I recommend to create a distinct fixture for the use we're making in the new tests in this PR. `PlanningUnitGridShape.Irregular` is not even supposed to be used anywhere in the platform for the time being, so it would be at least confusing to have this as a default for this fixture.

I'd either create a sibling fixture such as `projects.valid.minimalWithIrregularGridShape` or use the default `projects.valid.minimal` in the new test with a spread to overwrite the `planningUnitGridShape` prop with the desired `PlanningUnitGridShape.Irregular` value

api/apps/api/src/utils/acl.utils.ts
```diff
@@ -113,6 +115,10 @@ export const mapAclDomainToHttpError = (
       return new ForbiddenException(
         `User with ID: ${options?.userId} is not allowed to perform this action on ${options?.resourceType}.`,
       );
+    case projectIsMissingInfoForRegularPus:
+      return new BadRequestException(
+        `Project with regular planning unit shape is missing GADM id or country id.`,
```
hotzevzl, line 120: I suggest to use the descriptive error message from my ill-fated decorator-based PR:
```suggestion
        `When a regular planning grid is requested (hexagon or square) either a custom planning area or a GADM area gid must be provided.`,
```

(needs to be adjusted in tests as well)

api/apps/api/test/e2e.config.ts
hotzevzl: actually, looking at the new test, I'm not even sure about the need for the change I'm commenting about for that test... so anyway - wouldn't keeping `planningUnitGridShape: PlanningUnitGridShape.Hexagon` work here?

hotzevzl (approved):
other than the minor suggestion for test fixture and for error message noted inline, looks good to me, thank you!

api/apps/api/test/e2e.config.ts
AaronPerezPerez: As `Hexagon` is a regular shape, we would need to add a `planningUnitId` or any of the GADM id so the controller doesn't return a 400. Maybe adding a pu id is enought?

hotzevzl: ah ok, I see - yes, the main issue here may be with `createWorld.WhenCreatingProjectWithoutAdminAreas()` in `api/apps/api/test/project-planning-units/world.ts`, which is used in `api/apps/api/test/project-planning-units/planning-units.e2e-spec.ts` because this will fail before even getting to the assertion.

we also have one occurrence in `projects.e2e-spec.ts` and one in `organizations.e2e-spec.ts` where we should be ok with using `minimalInGivenAdminArea()` instead.

in these cases, using `Irregular` as grid shape would be cheating a bit, and the current behaviour (accepting a project without admin area nor planning area id) is incorrect, so it's best to remove the use of `projects.valid.minimal` instead.

PR #946, opened 2022-03-31
ref: shared features data entity

aciddaute:
This PR adds `GeoFeatureGeometry` entity to `libs` and removes duplicated entity from `api` and `geoprocessing`.


api/apps/geoprocessing/src/import/db-cleanup/db-cleanup.processor.ts
```diff
@@ -61,7 +60,7 @@ export class DbCleanupProcessor {
     const apiQb = this.apiEntityManager.createQueryBuilder();
     const geoQb = this.geoprocessingEntityManager.createQueryBuilder();
 
-    const scenarios: Scenario[] = await apiQb
+    const scenarios: { id: string }[] = await apiQb
```
hotzevzl, line 63: can we keep a reference to `Scenario`?
```suggestion
    const scenarios: Pick<Scenario, "id">[] = await apiQb
```
(but I think the imported `Scenario` here was unrelated to what we wanted to do - the full diff would be

```
diff --git a/api/apps/geoprocessing/src/import/db-cleanup/db-cleanup.processor.ts b/api/apps/geoprocessing/src/import/db-cleanup/db-cleanup.processor.ts
index a037969db..560d4cc82 100644
--- a/api/apps/geoprocessing/src/import/db-cleanup/db-cleanup.processor.ts
+++ b/api/apps/geoprocessing/src/import/db-cleanup/db-cleanup.processor.ts
@@ -1,3 +1,4 @@
+import { Scenario } from '@marxan-api/modules/scenarios/scenario.api.entity';
 import { ResourceKind } from '@marxan/cloning/domain';
 import { FailedImportDbCleanupJobInput } from '@marxan/cloning/job-input';
 import { FailedImportDbCleanupJobOutput } from '@marxan/cloning/job-output';
@@ -60,7 +61,7 @@ export class DbCleanupProcessor {
     const apiQb = this.apiEntityManager.createQueryBuilder();
     const geoQb = this.geoprocessingEntityManager.createQueryBuilder();

-    const scenarios: { id: string }[] = await apiQb
+    const scenarios: Pick<Scenario, "id">[] = await apiQb
       .select('id')
       .from('scenarios', 's')
       .where('project_id = :projectId', {
```

hotzevzl (approved):

api/apps/geoprocessing/src/import/db-cleanup/db-cleanup.processor.ts
hotzevzl: ah yes, you're right. no, let's leave it as in your PR.

PR #952, opened 2022-04-04
add k8s_storage module with initial storage class for Azure Files CSI over NFS [MARXAN-1428]

hotzevzl:
https://vizzuality.atlassian.net/browse/MARXAN-1428

### DIFF
--- a/infrastructure/kubernetes/main.tf
+++ b/infrastructure/kubernetes/main.tf
@@ -39,6 +39,9 @@ locals {
   k8s_client_certificate     = base64decode(data.azurerm_kubernetes_cluster.k8s_cluster.kube_config.0.client_certificate)
   k8s_client_key             = base64decode(data.azurerm_kubernetes_cluster.k8s_cluster.kube_config.0.client_key)
   k8s_cluster_ca_certificate = base64decode(data.azurerm_kubernetes_cluster.k8s_cluster.kube_config.0.cluster_ca_certificate)
+  backend_storage_class      = "azurefile-csi-nfs"
+  backend_storage_size       = "100Gi"
+  backend_storage_pvc        = "backend-shared-spatial-data-storage"
 }
 
 module "k8s_namespaces" {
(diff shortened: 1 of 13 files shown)


infrastructure/kubernetes/main.tf
```diff
@@ -39,6 +39,9 @@ locals {
   k8s_client_certificate     = base64decode(data.azurerm_kubernetes_cluster.k8s_cluster.kube_config.0.client_certificate)
   k8s_client_key             = base64decode(data.azurerm_kubernetes_cluster.k8s_cluster.kube_config.0.client_key)
   k8s_cluster_ca_certificate = base64decode(data.azurerm_kubernetes_cluster.k8s_cluster.kube_config.0.cluster_ca_certificate)
+  backend_storage_class      = "azurefile-csi-nfs"
+  backend_storage_size       = "100Gi"
```
hotzevzl, line 43: way oversized, but:
- we may want to sometimes keep temporary files around for inspection (I added a change yesterday that enabled this for staging, and for whatever dev envs via env vars)
- this is using standard locally-redundant storage for now, but if we wanted to use premium later on, 100Gi is (from what I understand) the minimum allocation
- storage is cheap

infrastructure/kubernetes/modules/storage/main.tf
```diff
@@ -0,0 +1,11 @@
+resource "kubernetes_storage_class" "azurefile_csi_nfs" {
+  metadata {
+    name = var.backend_storage_class
+  }
+  storage_provisioner = "file.csi.azure.com"
+  reclaim_policy      = "Delete"
+  parameters = {
+    protocol = "nfs"
```
hotzevzl, line 8: `azurefile_csi` (based on SMB rather than NFS) comes included with AKS clusters, but
- it is case-preserving, case-insensitive - from what I understand from the docs 🙄 
- NFS shares use Posix semantics for file access, and I wanted to avoid a possible small discrepancy between dev and staging/prod

infrastructure/kubernetes/main.tf
tiagojsag: put this into a var, so it's easy to upscale in the future

tiagojsag (changes requested):

PR #955, opened 2022-04-05
MARXAN-1375-webshot-service-image-url

rubvalave:
-Adds png-image generation service for webshot.
-Adds url to be used for png-generation.


hotzevzl (approved):
great, thanks!

we won't know if it works until the frontend page is ready, but I can volunteer to test this by pointing it to some other page 😅 

only a cosmetic comment - I would create a `domain/blm-previews` folder and move the `png-image.ts` source in there (this new code is not about solutions report, so putting it in that folder may be a big confusing - though all of this is nonsense in any case)

a possible improvement (for this _and_ my earlier webshot handler) would be to properly pass on any error message we generate _or_ collect (for example, if `page.waitForFunction()` times out) to the webshot response, so that when this is received by the `WebshotService` on the API/geoprocessing side we may include more useful error messages than the current generic text in responses to the final consumer (mostly relevant for the PDF one, as the geoprocessing service is the actual consumer of this new handler, so any error messages won't bubble up to the user in any case).

but this could be for a separate PR, if at all, if you think this would be useful.

PR #985, opened 2022-04-12
feat(front): download project[MARXAN-1477]

anamontiaga:
### Overview

We are waiting for the call `/api/v1/projects/{projectId}/export` to return a `userId` so we can identify which user is downloading it.

In addition, I find a problem when downloading the project from the list, since the blocking screen of the async process would not appear in that list, since there are more projects.

Right now I have removed the `stopPropagation` on `handleDownload` function so that it can access the project dashboard, but it looks a bit strange.

### Feature relevant tickets

[MARXAN-1477](https://vizzuality.atlassian.net/browse/MARXAN-1477)


mbarrenechea:
@anamontiaga  I don't know what I did... but having 122 commits and 377 files to merge is not ok to me. I'm going to close this PR and open a new one with your 6 commits you did... SORRY

PR #999, opened 2022-04-20
MARXAN-1274-publish-project-data

rubvalave:
-Adds new columns to publish project entity.
-Adapts tests to test if new data is included.
-Changes how tests are published by including a DTO.


hotzevzl (approved):
great 😃 

I must confess I am not fully sold on splitting up document-like content across several JSONB columns (I would have used a single JSONB column and stuffed all the stuff in there, combining the specific types), but your proposed solution should work perfectly in any case, and given that we don't expect much of a change in these data structures, it should also make the implementation simpler (especially for partial updates where only some fields are populated).

Let's merge!

PR #1007, opened 2022-04-22
MARXAN-1274-publicMap-screenshot

rubvalave:
-Adds screenshot generation to publish action.
-Adds new location field to published project entity.
-Adapts tests to new fields.
-Minor removal of redundant fixture data.

### DIFF
--- a/api/apps/api/src/modules/published-project/entities/published-project.api.entity.ts
+++ b/api/apps/api/src/modules/published-project/entities/published-project.api.entity.ts
@@ -5,7 +5,6 @@ import {
   Creator,
   Resource,
 } from '../dto/create-published-project.dto';
-import { string } from 'fp-ts';
 
 @Entity('published_projects')
 export class PublishedProject {
--- a/api/apps/api/src/modules/published-project/entities/published-project.api.entity.ts
+++ b/api/apps/api/src/modules/published-project/entities/published-project.api.entity.ts
@@ -5,6 +5,7 @@ import {
   Creator,
   Resource,
 } from '../dto/create-published-project.dto';
+import { string } from 'fp-ts';
 
 @Entity('published_projects')
 export class PublishedProject {
@@ -17,6 +18,9 @@ export class PublishedProject {
   @Column('character varying')
   description?: string;
 
+  @Column('character varying')
+  location?: string;
+
   @Column('boolean', { name: 'under_moderation', default: false })
   underModeration?: boolean;
 
@@ -28,6 +32,9 @@ export class PublishedProject {
 
   @Column('jsonb')
   creators?: Creator[];
+
+  @Column('character varying', { name: 'png_data' })
+  pngData?: string;
 
   @OneToOne(() => Project)
   @JoinColumn({ name: 'id' })
--- a/api/apps/api/src/modules/published-project/published-project.service.ts
+++ b/api/apps/api/src/modules/published-project/published-project.service.ts
@@ -13,7 +13,7 @@ import { PublishProjectDto } from './dto/publish-project.dto';
 import { WebshotService } from '@marxan/webshot';
 import { AppConfig } from '@marxan-api/utils/config.utils';
 import { assertDefined } from '@marxan/utils';
-import { isLeft, isRight } from 'fp-ts/lib/These';
+import { isLeft, isRight } from 'fp-ts/lib/Either';
 
 export const notFound = Symbol(`project not found`);
 export const accessDenied = Symbol(`not allowed`);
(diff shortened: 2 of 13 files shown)


api/apps/api/src/modules/published-project/entities/published-project.api.entity.ts
```diff
@@ -5,6 +5,7 @@ import {
   Creator,
   Resource,
 } from '../dto/create-published-project.dto';
+import { string } from 'fp-ts';
```
hotzevzl, line 8: I don't think this is what you meant (though it may be what LSP and vscode thought you meant)

api/apps/api/src/modules/published-project/published-project.service.ts
```diff
@@ -10,6 +10,10 @@ import { PublishedProject } from '@marxan-api/modules/published-project/entities
 import { ProjectAccessControl } from '@marxan-api/modules/access-control';
 import { UsersService } from '@marxan-api/modules/users/users.service';
 import { PublishProjectDto } from './dto/publish-project.dto';
+import { WebshotService } from '@marxan/webshot';
+import { AppConfig } from '@marxan-api/utils/config.utils';
+import { assertDefined } from '@marxan/utils';
+import { isLeft, isRight } from 'fp-ts/lib/These';
```
hotzevzl, line 16:
```suggestion
import { isLeft, isRight } from 'fp-ts/lib/Either';
```

api/apps/api/src/modules/published-project/published-project.service.ts
```diff
@@ -59,13 +64,41 @@ export class PublishedProjectService {
...
+        },
+      },
+      webshotUrl,
+    );
+
+    if (isLeft(pngData)) {
+      console.info(
+        `Map screenshot for public project ${id} could not be generated`,
+      );
+    }
```
hotzevzl, line 94: this would be a good candidate for breaking out into a new method (not sure we could reuse much of it when moving this step to the marxan run workflow)... feel free to leave it as is though, especially as we may move this soon in any case.

api/apps/api/test/fixtures/test-data.sql
```diff
@@ -41,15 +41,6 @@ VALUES
...
 
-INSERT INTO users_scenarios
-(user_id, scenario_id, role_id)
-VALUES
-((SELECT id FROM users WHERE lower(email) = 'aa@example.com'), (SELECT id FROM scenarios WHERE name = 'Example scenario 1 Project 2 Org 2'), 'scenario_owner'),
-((SELECT id FROM users WHERE lower(email) = 'bb@example.com'), (SELECT id FROM scenarios WHERE name = 'Example scenario 1 Project 2 Org 2'), 'scenario_contributor'),
-((SELECT id FROM users WHERE lower(email) = 'cc@example.com'), (SELECT id FROM scenarios WHERE name = 'Example scenario 1 Project 2 Org 2'), 'scenario_viewer'),
-((SELECT id FROM users WHERE lower(email) = 'aa@example.com'), (SELECT id FROM scenarios WHERE name = 'Example scenario 2 Project 2 Org 2'), 'scenario_owner'),
-((SELECT id FROM users WHERE lower(email) = 'bb@example.com'), (SELECT id FROM scenarios WHERE name = 'Example scenario 2 Project 2 Org 2'), 'scenario_contributor'),
-((SELECT id FROM users WHERE lower(email) = 'cc@example.com'), (SELECT id FROM scenarios WHERE name = 'Example scenario 2 Project 2 Org 2'), 'scenario_viewer');
```
hotzevzl, line 52: ❤️

api/apps/api/test/project/projects.fixtures.ts
```diff
@@ -226,15 +247,18 @@ export const getFixtures = async () => {
...
         .post(`/api/v1/projects/${projectId}/publish`)
         .send({
           name: 'example project',
           description: 'fake description',
+          location: 'fake location',
           company: { name: 'logo', logoDataUrl: blmImageMock },
           creators: [{ displayName: 'fake name', avatarDataUrl: blmImageMock }],
           resources: [{ title: 'fake url', url: 'http://www.example.com' }],
+          config: { baseUrl: 'example/png', cookie: 'randomCookie' },
+          scenarioId,
```
hotzevzl, line 261: in hindsight, I think we should have labelled this `featuredScenarioId` in the DTO for this, but let's not get there now 😓

hotzevzl (approved):

api/apps/api/test/project/projects.fixtures.ts
rubvalave: In fact, I think I would change it as you suggested, mainly because once I start the export here, I would also be able to select which scenarios I want for the export and it can get confusing real fast.

api/apps/api/src/modules/published-project/published-project.service.ts
rubvalave: I think for sure what I can reuse is the config stuff, as it is the same for BLM than this images. Apart from that maybe a common method in webshotService lib to distinguish between blm and publishedProject images? I don't know...

hotzevzl: i'd say let's leave it as is in this PR, so that we can merge and go ahead with the full flow between frontend and backend, and if/when we move this to marxan run flow, make it a self-contained method (the marxan run flow is complicated enough already that keeping new steps kind of self-contained would be quite useful)

PR #1020, opened 2022-04-28
Feat/marxan 1469 hash and sign export files

aciddaute:
This PR adds logic for verifying and integrity checking import files signature and manifest files

### Feature relevant tickets

- [check signed hash of zip file contents users try to import](https://vizzuality.atlassian.net/browse/MARXAN-1470)
- [hash and sign exported data in downloadable zip artifacts](https://vizzuality.atlassian.net/browse/MARXAN-1469)
- [configure signing secret for integrity checks of exported data (via env var+config)](https://vizzuality.atlassian.net/browse/MARXAN-1467)

### DIFF
--- a/api/apps/api/src/modules/clone/export/adapters/node-manifest-file-service.adapter.ts
+++ b/api/apps/api/src/modules/clone/export/adapters/node-manifest-file-service.adapter.ts
@@ -1,0 +1,143 @@
+import {
+  CloningFilesRepository,
+  GetFileError,
+  SaveFileError,
+} from '@marxan/cloning-files-repository';
+import { ArchiveLocation } from '@marxan/cloning/domain';
+import {
+  manifestFileRelativePath,
+  signatureFileRelativePath,
+} from '@marxan/cloning/infrastructure/clone-piece-data/manifest-file';
+import { readableToBuffer } from '@marxan/utils';
+import { Inject, Injectable } from '@nestjs/common';
+import { spawnSync } from 'child_process';
+import { createSign } from 'crypto';
+import { Either, left, right } from 'fp-ts/lib/Either';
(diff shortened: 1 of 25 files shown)


api/apps/api/src/modules/clone/export/adapters/node-manifest-file-service.adapter.ts
```diff
@@ -0,0 +1,175 @@
...
+      );
+      const signatureFile = await this.fileRepository.get(signatureFileUri);
+      if (isLeft(signatureFile)) {
+        throw signatureFile;
+      }
+      const signatureFileBuffer = await readableToBuffer(signatureFile.right);
+      const uploadedSignature = signatureFileBuffer.toString();
+
+      if (recalculatedSignature !== uploadedSignature)
+        return left(invalidSignature);
```
hotzevzl, line 103: I think most of this could be done via something like this:
```suggestion
      const signatureFile = await this.fileRepository.get(signatureFileUri);
      if (isLeft(signatureFile)) {
        throw signatureFile;
      }
      const signatureFileBuffer = await readableToBuffer(signatureFile.right);
      const uploadedSignature = signatureFileBuffer.toString();

      const verifier = createVerify('RSA-SHA256');
      verifier.update(manifestFileBuffer);
      return verifier.verify(
        this.cloningSigningSecret,
        uploadedSignature,
        'hex',
      ) ? right(true) : left(invalidSignature);
```

aciddaute: Yes, we will change it 👍

hotzevzl:
@tiagojsag could I ask you to please have a look at the way I handled the new RSA key secret in the Terraform part of things?

tiagojsag (approved):

PR #1024, opened 2022-04-29
Fix/marxan 1246 cost surface templates puids

aciddaute:
This PR fixes cost surface templates `puid` fields

[cost surface templates should use a stable numeric puid for planning unit ids](https://vizzuality.atlassian.net/browse/MARXAN-1246)

### DIFF
--- a/api/apps/geoprocessing/src/modules/surface-cost/ports/shapefile-record.ts
+++ b/api/apps/geoprocessing/src/modules/surface-cost/ports/shapefile-record.ts
@@ -1,4 +1,0 @@
-export interface ShapefileRecord {
-  puid: number;
-  cost: number;
-}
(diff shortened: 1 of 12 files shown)


AaronPerezPerez (approved):

api/apps/geoprocessing/src/modules/surface-cost/ports/shapefile-record.ts
```diff
@@ -0,0 +1,4 @@
+export interface ShapefileRecord {
```
hotzevzl, line 1: very minor - could we rename this to stick more closely to the specific record kind? as is now, its name is quite generic (I know, it's within the `surface-cost` module, but it's not strictly speaking namespaced)

hotzevzl:
```suggestion
export interface CostSurfaceShapefileRecord {
```
maybe?

hotzevzl (approved):

api/apps/geoprocessing/src/modules/surface-cost/ports/shapefile-record.ts
aciddaute: Sure!

PR #1026, opened 2022-04-29
document some more secrets needed for GH actions

hotzevzl:


tiagojsag (changes requested):
line 118 covers the frontend secrets. please remove lines 107-116 of your commit

PR #1039, opened 2022-05-05
Public-Projects-ExportId-Switch

rubvalave:
-Add conditions on findAll/findOne methods in public projects to avoid showing exportId when it is not ready.
-Add specific test to check that exportId is shown if export is finished.

### DIFF
--- a/api/apps/api/src/modules/published-project/published-project-crud.service.ts
+++ b/api/apps/api/src/modules/published-project/published-project-crud.service.ts
@@ -53,6 +53,7 @@ export class PublishedProjectCrudService extends AppBaseService<
         'underModeration',
         'originalProject',
         'pngData',
+        'exportId',
       ],
       keyForAttribute: 'camelCase',
       originalProject: {
--- a/api/apps/api/src/modules/published-project/published-project-crud.service.ts
+++ b/api/apps/api/src/modules/published-project/published-project-crud.service.ts
@@ -87,7 +87,7 @@ export class PublishedProjectCrudService extends AppBaseService<
     if (exportIdString) {
       const exportId = new ExportId(exportIdString);
       const finalExport = await this.exportRepo.find(exportId);
-      if (!finalExport) {
+      if (!finalExport?.hasFinished()) {
         delete entity.exportId;
       }
     }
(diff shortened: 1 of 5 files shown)


api/apps/api/src/modules/published-project/published-project-crud.service.ts
```diff
@@ -73,6 +77,24 @@ export class PublishedProjectCrudService extends AppBaseService<
...
+  ): Promise<PublishedProject> {
+    const exportIdString = entity?.exportId;
+
+    if (exportIdString) {
+      const exportId = new ExportId(exportIdString);
+      const finalExport = await this.exportRepo.find(exportId);
+      if (!finalExport) {
+        delete entity.exportId;
+      }
+    }
```
aciddaute, line 93: If published project has an export id it should always exist on export repository (at least that we are doing something strange when publishing projects). If I remember correctly, Miguel wanted export id property only if the export has finished right? We should check if the export has finished after retrieving it from export repo. We usually check that by inspecting `archiveLocation` property inside `Export` aggregate. I think it is worth to create a method `isFinished(): boolean` to check that condition internally 🤔, I can take care if you want 👍  

```suggestion
    const exportIdString = entity?.exportId;

    if (exportIdString) {
      const exportId = new ExportId(exportIdString);
      const finalExport = await this.exportRepo.find(exportId);
      if (!finalExport?.isFinished()) {
        delete entity.exportId;
      }
    }
```

api/apps/api/src/modules/published-project/published-project.service.ts
```diff
@@ -253,10 +255,19 @@ export class PublishedProjectService {
...
-
     if (!result) {
       return left(notFound);
     }
+
+    if (result.exportId) {
+      const exportId: ExportId = new ExportId(result.exportId);
+
+      const exportResult = await this.exportRepo.find(exportId);
+      if (!exportResult) {
```
aciddaute, line 266: Same as above

api/apps/api/test/project/projects.fixtures.ts
```diff
@@ -25,6 +25,7 @@ import { bootstrapApplication } from '../utils/api-application';
 import { EventBusTestUtils } from '../utils/event-bus.test.utils';
 import { ScenariosTestUtils } from '../utils/scenarios.test.utils';
 import { ScenarioType } from '@marxan-api/modules/scenarios/scenario.api.entity';
+import { async } from 'rxjs';
```
aciddaute, line 28: It seems like a unwanted export 🤔

hotzevzl (commented):
no further comments from me besides what Daute already suggested - once that's taken care of this looks good to me.

aciddaute (approved):

PR #1044, opened 2022-05-06
ref: blm calculation function

aciddaute:
This PR changes blm values calculation function

### Feature relevant tickets

[Change BLM calculation function.](https://vizzuality.atlassian.net/browse/MARXAN-1524)

### DIFF
--- a/api/apps/api/test/scenarios/get-scenario-blm-calibration-results.fixtures.ts
+++ b/api/apps/api/test/scenarios/get-scenario-blm-calibration-results.fixtures.ts
@@ -115,6 +115,8 @@ export const getFixtures = async () => {
       results.forEach((result) => {
         expect(result.id).toBeDefined();
         expect(result.scenarioId).toBe(scenarioId);
+        expect(result.cost).toBeGreaterThanOrEqual(0);
+        expect(result.boundaryLength).toBeGreaterThanOrEqual(0);
       });
     },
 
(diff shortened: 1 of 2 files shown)


api/apps/api/test/scenarios/get-scenario-blm-calibration-results.fixtures.ts
```diff
@@ -115,10 +115,6 @@ export const getFixtures = async () => {
       results.forEach((result) => {
         expect(result.id).toBeDefined();
         expect(result.scenarioId).toBe(scenarioId);
-        expect(result.cost).toBeGreaterThanOrEqual(0);
-        expect(result.blmValue).toBeGreaterThanOrEqual(blmRange[0]);
-        expect(result.blmValue).toBeLessThanOrEqual(blmRange[1]);
-        expect(result.boundaryLength).toBeGreaterThanOrEqual(0);
```
hotzevzl, line 121: why aren't these assertions going to be useful anymore?

hotzevzl (approved):

api/apps/api/test/scenarios/get-scenario-blm-calibration-results.fixtures.ts
hotzevzl: I don't know either - it would be clearer if I looked at the formula in plain mathematical notation rather than as code 😓 but at a quick parse, I'd say that we may be able to at least assert that all the values are between 0 exclusive and... probably the next power-of-ten number higher than the upper bound? But I also think there would be very little value in checking these in practice - we're more likely to going to catch edge cases with floating point arithmetic than actual bugs in the formula. @aagm should we check that the generate values are within a given range at all?

AaronPerezPerez (approved):

PR #1050, opened 2022-05-10
fix: output scenarios summaries cloning

aciddaute:
This PR adds `output_scenarios_summaries` data to cloning process

### Feature relevant tickets

[BE - export/import/clone solutions data missing](https://vizzuality.atlassian.net/browse/MARXAN-1535)


hotzevzl (approved):
I haven't checked the details (columns set and so on) but from a few step back all looks good, thanks!

PR #1057, opened 2022-05-11
fix(front): BLM chart improvements [MARXAN-1392]

anamontiaga:
### Overview
- [x] Add min and max cost values to X axis BLM chart 
- [ ] There is only space to "more" value on yAxis boundaryLength Value and it seems a little bit confused so i decided not to included.
- [ ] Also, removed gradient seems not to be a good idea as it keeps the chart empty with no reference lines.

### Feature relevant tickets

[MARXAN-1392](https://vizzuality.atlassian.net/browse/MARXAN-1392)

---


anamontiaga:
There is only space to "more" value on yAxis boundaryLength Value and it seems a little bit confused so i decided not to included. 
Also, removed gradient seems not to be a good idea as it keeps the chart empty with no reference lines.

PR #1059, opened 2022-05-11
feat(front): add emails to admin panel [MARXAN-1547]

anamontiaga:
### Testing instructions
- Please, test with `Miguel-test project` as it has actually three owners and check if emails match with names and are showing in same order.

### Feature relevant tickets
[MARXAN-1547](https://vizzuality.atlassian.net/browse/MARXAN-1547)

### DIFF
--- a/app/hooks/admin/index.ts
+++ b/app/hooks/admin/index.ts
@@ -5,9 +5,6 @@ import {
   useQuery, useQueryClient,
 } from 'react-query';
 
-import keyBy from 'lodash/keyBy';
-import merge from 'lodash/merge';
-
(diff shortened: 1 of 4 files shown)


app/hooks/admin/index.ts
```diff
@@ -5,6 +5,9 @@ import {
   useQuery, useQueryClient,
 } from 'react-query';
 
+import keyBy from 'lodash/keyBy';
```
mbarrenechea, line 8: why do we need to use a lodash package to join 2 arrays?

mbarrenechea (changes requested):

PR #1090, opened 2022-05-19
ref: change blm default range

aciddaute:
This PR changes default blm range

### Feature relevant tickets

[Update BLM by default values](https://vizzuality.atlassian.net/browse/MARXAN-1551)


hotzevzl (approved):
great, thanks. just see if you wish to use package-relative imports instead of these long sequences of `../`s in paths for the `defaultBlmRange` imports

PR #1098, opened 2022-05-19
fix: synthetic events creation right before marking an import as fini…

aciddaute:
This PR adds synthetic event (`project.planningUnits.finished/v1/alpha`) for cloned projects

### Feature relevant tickets

[cannot create new scenarios in a project cloned from published project](https://vizzuality.atlassian.net/browse/MARXAN-1591)


hotzevzl (approved):

aciddaute:
Sure! 👍

PR #1104, opened 2022-05-23
MARXAN-1596-adds-error-handler-default-body-size

rubvalave:
-Adds error handler to instances of `entity.too.large` errors for express middleware.


api/apps/api/src/filters/all-exceptions.exception.filter.ts
```diff
@@ -24,14 +25,20 @@ import { AppConfig } from '../utils/config.utils';
 export class AllExceptionsFilter implements ExceptionFilter {
   private readonly logger = new Logger();
 
-  catch(exception: Error, host: ArgumentsHost) {
+  private isHttpError(exception: Error | HttpError): exception is HttpError {
```
hotzevzl, line 28: tiny: I would even avoid the `| HttpError` bit here - in my view, it doesn't really matter. When the function is called we know we have an `Error` of some kind, and then we do some narrowing in the chained ternary operator contraption further down.

api/package.json
```diff
@@ -56,6 +56,7 @@
     "@pgtyped/cli": "0.11.0",
     "@pgtyped/query": "0.11.0",
     "@turf/turf": "6.5.0",
+    "@types/http-errors": "^1.8.2",
```
hotzevzl, line 59: I had missed this - it should be a dev-dependency, not a dependency

PR #1110, opened 2022-05-25
MARXAN-1609-projects-sources-column

rubvalave:
-Adds migration to add `sources` column and enum type to `projects` table.

### DIFF
--- a/api/apps/api/src/modules/projects/project.api.entity.ts
+++ b/api/apps/api/src/modules/projects/project.api.entity.ts
@@ -140,8 +140,9 @@ export class Project extends TimeUserEntityMetadata {
   @Column('enum', {
     name: 'sources',
     enum: ProjectSourcesEnum,
+    default: ProjectSourcesEnum.marxanCloud,
   })
-  sources?: string;
+  sources!: ProjectSourcesEnum;
 
   /**
    * JSONB storage for non-relational attributes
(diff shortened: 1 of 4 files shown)


aciddaute:
I think the migration is perfectly fine but I'm missing two things:

- We should update `Project` Typeorm api entity to reflect the new column
- We should create project shell specifying as `sources` value `legacy-import` when importing legacy projects (`createShells` method in `api/apps/api/src/modules/legacy-project-import/application/start-legacy-project-import.handler.ts`)

aciddaute:
I think the migration is perfectly fine but I'm missing two things:

- We should update `Project` Typeorm api entity to reflect the new column
- We should create project shell specifying as `sources` value `legacy-import` when importing legacy projects (`createShells` method in `api/apps/api/src/modules/legacy-project-import/application/start-legacy-project-import.handler.ts`)

api/apps/api/src/modules/projects/project.api.entity.ts
```diff
@@ -131,6 +136,13 @@ export class Project extends TimeUserEntityMetadata {
   @OneToOne(() => ProjectBlm)
   projectBlm!: ProjectBlm;
 
+  @ApiProperty()
+  @Column('enum', {
+    name: 'sources',
+    enum: ProjectSourcesEnum,
+  })
+  sources?: string;
```
aciddaute, line 144: I think we should mark `sources` attribute as mandatory since it can't hold null values. Also we should use the enum as type right?

```suggestion
  sources!: ProjectSourcesEnum;
```

aciddaute (changes requested):

aciddaute (approved):

PR #1115, opened 2022-05-26
Antimeridian and bbox calc fix

aagm:
## Take with care  
This fix is touching almost all intersections based on bbox in the application

### Overview
So this applies 2 different fixes altogether:
BBox crosses the antimeridian, we are overpassing this in 2 steps, first, we properly calculate min and max longitude values by dividing the bbox in 2 bases on the east and west semihemisphers. 
After that both FE and backend will follow the next rule: if min long is > than max long, that means that the bbox crosses the antimeridian and we subdivide them.

Regarding the grid. So as we are already setting the equal area cost surface, I did take the license to every grid that is crossing the antimeridian will be cut in order to avoid it crossing the antimeridian. with this in place calculations should work as expected.

Also, small fixes were applied to everything that depends on bbox filtering in order to be able to overpass the bbox at the antimeridian situation. The only problem here might be a small overhead due to subdividing it into the east and west suboxes.

### Testing instructions

Up locally and setup a project in Fiji and in any other place

### Feature relevant tickets

https://vizzuality.atlassian.net/browse/MARXAN-985

---

- [x ] Meaningful commits and code rebased on `develop`.
- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/geo-features/geo-features.service.ts
+++ b/api/apps/api/src/modules/geo-features/geo-features.service.ts
@@ -28,7 +28,7 @@ import { DbConnections } from '@marxan-api/ormconfig.connections';
 import { v4 } from 'uuid';
 import { UploadShapefileDTO } from '../projects/dto/upload-shapefile.dto';
 import { GeoFeaturesRequestInfo } from './geo-features-request-info';
-import { antimeridianBbox } from '@marxan-geoprocessing/utils/bbox.utils';
+import { antimeridianBbox } from '@marxan/utils/geo/bbox';
 
 const geoFeatureFilterKeyNames = [
   'featureClassName',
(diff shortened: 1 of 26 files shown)


api/libs/utils/src/geo/bbox.ts
```diff
@@ -0,0 +1,38 @@
...
+ * @debt This should be moved to a self-standing
+ * @debt there is also a mismatch in terms of the bbox creation [xmax,xmin,ymax,ymin] vs [xmin,xmax,ymin,ymax]
+ */
+
+/**
+ * conversion operation between bbox [xmin, ymin, xmax, ymax]
+ * to Nominatim bbox  [xmin, xmax, ymin, ymax].
+ *
+ */
+export function bbox2Nominatim(bbox: BBox): BBox {
```
hotzevzl, line 14: you could consider aliasing `BBox` as `NominatimBBox` to improve type expressiveness in function signatures, even though in practice it's still just an array of 4 or 6 numbers.

hotzevzl: nothing to do here though, unless you wish to, of course 😃

hotzevzl (approved):
Great job on this one, and painstaking patience - thank you! 🚀 🙌🏼 

Mostly cosmetic/expressiveness notes from my side just because I can't help 😅

api/apps/api/src/modules/geo-features/geo-features.service.ts
```diff
@@ -167,20 +168,24 @@ export class GeoFeaturesService extends AppBaseService<
...
       const geoFeaturesWithinProjectBbox = await this.geoFeaturesGeometriesRepository
         .createQueryBuilder('geoFeatureGeometries')
         .select('"geoFeatureGeometries"."feature_id"', 'featureId')
         .distinctOn(['"geoFeatureGeometries"."feature_id"'])
         .where(
-          `st_intersects(
-        st_makeenvelope(:xmin, :ymin, :xmax, :ymax, 4326),
+          `(st_intersects(
+            st_intersection(st_makeenvelope(:...eastBbox, 4326),
+            ST_MakeEnvelope(0, -90, 180, 90, 4326)),
```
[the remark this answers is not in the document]
aagm: yes in the future we will need to revisit some of this, as well as at some point pay the debt in regarding sql formulation

PR #1117, opened 2022-05-27
MARXAN-1611-solutions-are-locked

rubvalave:
- Adds `solutions_are_locked` column to scenarios table.
- Sets `solutions_are_locked`column to true in scenario shell of legacy project import process.


api/apps/api/src/modules/scenarios/scenario.api.entity.ts
```diff
@@ -13,9 +13,9 @@ import { User } from '../users/user.api.entity';
 import { IsArray, IsOptional } from 'class-validator';
 import { TimeUserEntityMetadata } from '../../types/time-user-entity-metadata';
 import { BaseServiceResource } from '../../types/resource.interface';
-import { GeoFeatureSetSpecification } from '../geo-features/dto/geo-feature-set-specification.dto';
 import { JsonApiAsyncJobMeta } from '@marxan-api/dto/async-job.dto';
 import { ScenarioBlm } from '@marxan-api/modules/blm/values/repositories/scenario-blm/scenario-blm.api.entity';
+import { boolean } from 'fp-ts';
```
aciddaute, line 18: It seems like one of those sneaky undesired `fp-ts` imports :joy:

api/apps/api/src/modules/legacy-project-import/application/start-legacy-project-import.handler.ts
```diff
@@ -45,6 +45,7 @@ export class StartLegacyProjectImportHandler
       const scenario = await this.scenarioRepo.save({
         name,
         projectId: project.id,
+        solutions_are_locked: true,
```
aciddaute, line 48: As we talked, `soulutions_are_locked` value will be provided by the user. We should add a parameter to `StartLegacyProjectImport` command

aciddaute, line 48: As we talked, `soulutions_are_locked` value will be provided by the user. We should add a parameter to `StartLegacyProjectImport` command

api/apps/api/src/modules/scenarios/scenario.api.entity.ts
```diff
@@ -78,6 +78,9 @@ export class Scenario extends TimeUserEntityMetadata {
   @OneToOne(() => ScenarioBlm)
   scenarioBlm!: ScenarioBlm;
 
+  @Column('boolean')
+  solutionsAreLocked?: boolean;
```
aciddaute, line 82: I think we should set `solutionsAreLocked` as required since it can't hold null values :thinking: 

```suggestion
  solutionsAreLocked!: boolean;
```

aciddaute, line 82: I think we should set `solutionsAreLocked` as required since it can't hold null values :thinking: 

```suggestion
  solutionsAreLocked!: boolean;
```

aciddaute (changes requested):

api/apps/api/src/modules/projects/dto/legacy-project-import.dto.ts
```diff
@@ -14,6 +14,7 @@ export class StartLegacyProjectImportBodyDto {
   })
   @IsString()
   projectName!: string;
+  solutionsAreLocked!: boolean;
```
aciddaute, line 17: `solutionsAreLocked` property should be properly decorated :thinking:. This is causing a false positive in `api/apps/api/test/legacy-project-import/legacy-project-import.e2e-spec.ts`, we are not providing a value for `solutionsAreLocked` (error) but API doesn't complain

aciddaute (changes requested):

aciddaute (approved):

PR #1120, opened 2022-05-31
MARXAN-1610-amount_from_legacy_projects

rubvalave:
- Adds column to `scenarios_feature_data`.
- Changes how puvsprdat.service creates file.


aciddaute:
Migration looks good to me but I'm not sure about `PuvsprDatService` query, maybe @hotzevzl  can help us :wink:. Also, I think we should add `amountFromLegacyProject` property to `scenario-features-data` Typeorm entity :thinking:

hotzevzl:
I think we're all set here after merging in my suggestions - I'll wait for CI to finish and then merge and we'll have `amount_from_legacy_projects` available - sorry for having taken an eternity to provide feedback on this one 😏

PR #1126, opened 2022-06-06
[MARXAN-1610] suggestions on top of #1120

hotzevzl:
- allow floats for `amount_from_legacy_project`
- update query to select `amount_from_legacy_project` alongside `species` subquery data

### DIFF
--- a/api/apps/api/src/modules/scenarios/input-files/puvspr.dat.service.ts
+++ b/api/apps/api/src/modules/scenarios/input-files/puvspr.dat.service.ts
@@ -23,10 +23,20 @@ export class PuvsprDatService {
       amount: number;
     }[] = await this.connection.query(
       `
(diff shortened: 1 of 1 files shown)


rubvalave:
> The rich man implementation of the amount_from_legacy_projects column looks very good!

Though I think the SQL Query is somehow failing, because the tests that are failing are the ones that failed to me as well at first (and that I fixed adding the big about the `scenarios_features_data`).

hotzevzl:
> > The rich man implementation of the amount_from_legacy_projects column looks very good!
> 
> Though I think the SQL Query is somehow failing, because the tests that are failing are the ones that failed to me as well at first (and that I fixed adding the big about the `scenarios_features_data`).

yep, this was me forgetting a stray `,` while rebasing my changes on top of your PR 🤦🏼

PR #1128, opened 2022-06-08
MARXAN-1580-disable-initial-export-published-projects

rubvalave:
-Adds and follows the same logic per `findAll` method that was already used in `getById` for published project entities.
-This means that `exportId` would only be present in a project IF and only IF the export has finished. If not, the entity would not include the `exportId` among its properties.

### DIFF
--- a/api/apps/api/src/modules/published-project/published-project-crud.service.ts
+++ b/api/apps/api/src/modules/published-project/published-project-crud.service.ts
@@ -146,6 +146,12 @@ export class PublishedProjectCrudService extends AppBaseService<
       return entitiesAndCount;
     }
 
+    const extendedEntitiesFromGetById: PublishedProject[] = await Promise.all(
+      entitiesAndCount[0].map(
+        async (entity) => await this.extendGetByIdResult(entity),
+      ),
+    );
+
     const allOwnersOfAllProjectsPlusEmail = await this.usersProjectsRepo.query(
       `
       select up.project_id, up.user_id, u.email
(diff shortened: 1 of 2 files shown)


api/apps/api/src/modules/published-project/published-project-crud.service.ts
```diff
@@ -146,6 +146,12 @@ export class PublishedProjectCrudService extends AppBaseService<
       return entitiesAndCount;
     }
 
+    const extendedEntitiesFromGetById: PublishedProject[] = await Promise.all(
+      entitiesAndCount[0].map(
+        async (entity) => await this.extendGetByIdResult(entity),
+      ),
+    );
```
hotzevzl, line 153: It should do what it says on the tin... I believe 😃 

My suggestion for this extend-all (I think? or am I being Delphine the III here? 😬 non, non, non, non, non...!) to move the `extendGetByIdResult()` logic that deletes `exportId` when export has not finished yet into a small function that does the same thing but for an array of `exportIdString`s, so that we can avoid a `this.exportRepo.find(exportId)` for each of the entities  we have in `entitiesAndCount[0]` here. In practice this shouldn't be a problem in this case as we would have at most `pageSize` entities, so unless a direct API user (other than the frontend, that does things sanely) uses a huge `pageSize` (and we actually do ever have that many results), then the overhead would be minimal.

So my comment is mostly a general recommendation: rather than making an array processing a N-times repetition of processing a single element, process an array by default and make the processing of a single element a special case of this, for an array of size 1 - so that we can do this in `O(1)` (more or less) time rather than `O(N)`.

hotzevzl (approved):

PR #1131, opened 2022-06-09
Feat/marxan 1612 cancel legacy project imports

angelhigueraacid:
This PR adds :  

- `toBeRemoved` and `status` to legacy project import model
- canceled legacy project import api event
- `CancelLegacyProjectImport` command and `HaltLegacyProjectImport` event
- endpoint for canceling legacy project imports
- drops `ON DELETE CASCADE` from `legacy project imports table` on `project_id foreign key`

[Block legacy projects being imported](https://vizzuality.atlassian.net/browse/MARXAN-1612)

### DIFF
--- a/api/apps/api/src/modules/legacy-project-import/application/add-file-to-legacy-project-import.handler.spec.ts
+++ b/api/apps/api/src/modules/legacy-project-import/application/add-file-to-legacy-project-import.handler.spec.ts
@@ -187,7 +187,7 @@ const getFixtures = async () => {
         files,
         pieces,
         status: isAcceptingFiles
-          ? LegacyProjectImportStatuses.AceptingFiles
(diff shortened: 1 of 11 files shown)


api/apps/api/src/modules/legacy-project-import/domain/events/halt-legacy-project-import.events.ts
```diff
@@ -0,0 +1,6 @@
+import { ResourceId } from '@marxan/cloning/domain';
+import { IEvent } from '@nestjs/cqrs';
+
+export class HaltLegacyProjectImport implements IEvent {
```
aciddaute, line 4: Events should have past tense names, they represent something that happened. `HaltLegacyProjectImport` seems like a command name to me (something imperative)

aciddaute (approved):

PR #1136, opened 2022-06-13
forbid changes to legacy project imports [MARXAN-1612]

angelhigueraacid:
This PR adds:

- `LegacyProjectImportChecker` service. Now block guard checks if a project is blocked depending if there is an ongoing legacy project import for the given project.

### Feature relevant tickets
[Block legacy projects being imported](https://vizzuality.atlassian.net/browse/MARXAN-1612)

### DIFF
--- a/api/apps/api/src/modules/legacy-project-import/application/mark-legacy-project-import-as-finished.handler.ts
+++ b/api/apps/api/src/modules/legacy-project-import/application/mark-legacy-project-import-as-finished.handler.ts
@@ -1,3 +1,5 @@
+import { ProjectRoles } from '@marxan-api/modules/access-control/projects-acl/dto/user-role-project.dto';
+import { UsersProjectsApiEntity } from '@marxan-api/modules/access-control/projects-acl/entity/users-projects.api.entity';
 import { ApiEventsService } from '@marxan-api/modules/api-events';
 import { API_EVENT_KINDS } from '@marxan/api-events';
 import { ResourceId } from '@marxan/cloning/domain';
@@ -7,7 +9,9 @@ import {
   CommandHandler,
   IInferredCommandHandler,
 } from '@nestjs/cqrs';
+import { InjectRepository } from '@nestjs/typeorm';
 import { isLeft } from 'fp-ts/lib/Either';
+import { Repository } from 'typeorm';
 import { LegacyProjectImportRepository } from '../domain/legacy-project-import/legacy-project-import.repository';
 import { MarkLegacyProjectImportAsFailed } from './mark-legacy-project-import-as-failed.command';
 import { MarkLegacyProjectImportAsFinished } from './mark-legacy-project-import-as-finished.command';
@@ -19,6 +23,8 @@ export class MarkLegacyProjectImportAsFinishedHandler
     private readonly apiEvents: ApiEventsService,
     private readonly legacyProjectImportRepository: LegacyProjectImportRepository,
     private readonly commandBus: CommandBus,
+    @InjectRepository(UsersProjectsApiEntity)
+    private readonly usersRepo: Repository<UsersProjectsApiEntity>,
     private readonly logger: Logger,
   ) {
     this.logger.setContext(MarkLegacyProjectImportAsFinishedHandler.name);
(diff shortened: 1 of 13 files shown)


angelhigueraacid:
Hi @hotzevzl ! The only thing left that i would like to review about this PR is about using this guard when creating a scenario.
Right now in the `create` method of `api/apps/api/src/modules/scenarios/scenarios.service.ts`, we are not doing `this.blockGuard.ensureThatScenarioIsNotBlocked(scenario.right.id);` so we don't check if any of the blocking operation are being made when creating a new scenario. So,¿Is this the case and we don't want to check if any of these blocking operations are being made? Also, i would say we should at least prevent from creating a scenario when the legacy project import is being executed(on the same project).

api/apps/api/src/modules/legacy-project-import/domain/legacy-project-import-checker/legacy-project-import-checker.service.ts
```diff
@@ -0,0 +1,12 @@
+import { Either } from 'fp-ts/Either';
+
+export const legacyProjectImportDoesntExist = Symbol(
+  `doesn't exist legacy project import`,
+);
+export type LegacyProjectImportDoesntExist = typeof legacyProjectImportDoesntExist;
+
+export abstract class LegacyProjectImportChecker {
+  abstract hasImportedLegacyProjectImport(
```
hotzevzl, line 9: sorry about the picky note, but I think the name is unclear - from what the implementations do I think this could be renamed to something like this?

```suggestion
  abstract isLegacyProjectImportCompletedFor(
```

api/apps/api/src/modules/legacy-project-import/application/mark-legacy-project-import-as-finished.handler.ts
```diff
@@ -61,5 +67,11 @@ export class MarkLegacyProjectImportAsFinishedHandler
         ownerId,
       },
     });
+
+    await this.usersRepo.save({
+      userId: ownerId,
+      projectId: projectId.value,
+      roleName: ProjectRoles.project_owner,
+    });
```
hotzevzl, line 75: no need to change anything, especially at this stage - just noting, this would be a great candidate for a small expressive function to remind our future selves of what is the purpose of these lines 😱 - `setInitialOwnerForImportedProject()` or something like this. Let's leave it as is in any case - we do have a few other similar occurrences around, and it would only make sense to touch them all at once - but definitely not at this stage 😱 😅

api/apps/api/src/modules/legacy-project-import/domain/legacy-project-import-checker/legacy-project-import-checker.service.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+
+it.each(
+  Object.values(LegacyProjectImportStatuses).filter(
+    (kind) => kind !== LegacyProjectImportStatuses.Completed,
+  ),
+)(
+  `hasImportedLegacyProjectImport() should return false if given legacy project import has status %s`,
+  async (kind) => {
+    const id = await fixtures.GivenLegacyProjectImport(
+      LegacyProjectImportStatuses.AcceptingFiles,
```
hotzevzl, line 38: shouldn't this be
```suggestion
      kind,
```
?

Otherwise, we're not making use of the `kind` param we get by iterating over the statuses except the `Completed` one

api/apps/api/src/modules/legacy-project-import/domain/legacy-project-import-checker/legacy-project-import-checker.service.ts
hotzevzl: (if you feel it should be amended for clarity, please remember to update the function name in test descriptions too)

api/apps/api/src/modules/projects/block-guard/marxan-block-guard.service.spec.ts
```diff
@@ -180,8 +196,18 @@ describe('MarxanBlockGuard - ensureThatScenarioIsNotBlocked', () => {
       .ThenAPendingMarxanRunErrorIsThrown();
   });
 
+  it(`throws an exception if the given scenario's parent project has not imported the legacy project import`, async () => {
```
hotzevzl, line 199: maybe
```suggestion
  it(`while a legacy project import process is ongoing, scenarios within the project should be marked as non-editable`, async () => {
```

hotzevzl (approved):
thanks @angelhigueraacid - all good, except for some naming that in my opinion could be improved for clarity.

PR #1138, opened 2022-06-13
add dat file delimiter finder service and adapt legacy importers 

angelhigueraacid:
This PR adds `DatFileDelimiterFinder` service and adapts legacy importers so they can read dat files using commas and tabs as delimiters 

### Feature relevant tickets
[Support tabs and commas as delimiters when reading dat files](https://vizzuality.atlassian.net/browse/MARXAN-1639)

### DIFF
--- a/api/apps/geoprocessing/src/legacy-project-import/legacy-piece-importers/file-readers/dat-file.comma-or-tab-finder.spec.ts
+++ b/api/apps/geoprocessing/src/legacy-project-import/legacy-piece-importers/file-readers/dat-file.comma-or-tab-finder.spec.ts
@@ -1,0 +1,88 @@
+import { FixtureType } from '@marxan/utils/tests/fixture-type';
+import { Test } from '@nestjs/testing';
+import { isRight, isLeft } from 'fp-ts/lib/Either';
+import { Readable } from 'stream';
+import {
+  DatFileCommaOrTab,
+  DatFileCommaOrTabFinder,
+} from './dat-file.comma-or-tab-finder';
+import { invalidDelimiter } from './dat-file.delimiter-finder';
+
+let fixtures: FixtureType<typeof getFixtures>;
+
+beforeEach(async () => {
+  fixtures = await getFixtures();
+});
+
+it('fails when dat file uses invalid delimiter', async () => {
+  const fileLocation = fixtures.GivenADatFileWithInvalidDelimiter();
+  const result = await fixtures.WhenGettingDaDelimiter(fileLocation);
+  fixtures.ThenDatFileFindDelimiterOperationFails(result);
+});
+
+it('succes when dat file uses comma delimiter', async () => {
+  const fileLocation = fixtures.GivenAValidDatFileWithCommas();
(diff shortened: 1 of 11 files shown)


api/apps/geoprocessing/src/legacy-project-import/legacy-piece-importers/file-readers/dat-file.comma-or-tab-finder.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+
+let fixtures: FixtureType<typeof getFixtures>;
+
+beforeEach(async () => {
+  fixtures = await getFixtures();
+});
+
+it('fails when dat file uses invalid delimiter', async () => {
+  const fileLocation = fixtures.GivenADatFileWithInvalidDelimiter();
+  const result = await fixtures.WhenGettingDaDelimiter(fileLocation);
```
hotzevzl, line 19: tiny typo - I suppose you meant `WhenGettingDatDelimiter` (here and in other tests + fixtures)

api/apps/geoprocessing/src/legacy-project-import/legacy-piece-importers/file-readers/dat-file.comma-or-tab-finder.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  private commaOrTabDelimiter = (line: string) => {
+    const tabPattern = /[a-z]\t[a-z]+/;
+    const commaPattern = /[a-z],[a-z]+/;
+
+    const tabDelimiter = tabPattern.test(line);
+    const commaDelimiter = commaPattern.test(line);
+
+    return { tabDelimiter, commaDelimiter };
+  };
```
hotzevzl, line 41: This should work in the specific case of Marxan input files where users didn't do much messing-up of their files - in theory trying to parse (via `fast-csv`) the first line with both comma and tab as separators and check if we have "columns" once parsing should help with (for example) spurious tabs that may have sneaked into the header line of a comma-separated file, but yeah, let's keep this simple (especially as parsing will likely break later on in any case if we "guess" the separator wrong here).
Likewise if a file contains only one column we wouldn't recognize the separator here (but likely not by attempting to parse via `fast-csv` either, unless we handle the case explicitly), but again in practice there should never be any Marxan legacy input files with one single column, so tl;dr I think we're ok here - I wrote this comment up for you @angelhigueraacid  to cross-check, but I think we can merge as is.

angelhigueraacid: Yes, now thinking of it, perhaps we could have use the `fast-csv` library for reading the firs line and skipping the rest. We can always just make a new implementation for `DatFileDelimiterFinder` using `fast-csv`.
For dat files that only contain one column, the result will be `invalidDelimiter` so i think it should be okey as any input file with just one column seems already invalid. I will solve the `fisrtLine` typos that i spread around every legacy importer and merge the PR!

PR #1142, opened 2022-06-17
MARXAN-1643-modify-signup-to-include-bakground

rubvalave:
-Adds two new columns and enums for `users` columns.
-Adapts sign-up endpoint to allow sending `background` and `level` information.


hotzevzl (approved):
My spider sense (https://en.wikipedia.org/wiki/Macrothele_calpeiana#/media/File:Macrothele_calpeiana.JPG) makes me lean on "just stuffing stuff into a JSONB document" kind of approach for this stuff (we should be ok given that we're wrapping things up, but these closed lists often escalate quickly to need to include "other" and then some free user input for "other"... and so on).

So, I guess we can merge as is, in this case.

PR #1144, opened 2022-06-20
Feat/marxan 1653 remove exports and imports rows

angelhigueraacid:
This PR adds :

- `exportId` property to `imports` aggregate and table
- `delete` method in `exports` repository
- If scenario clones success, deletes the associated `export` ( and `import`)

### Feature relevant tickets

[remove exports and imports rows for a cloned scenario once the destination scenario has been fully cloned (or if the export or import fails)](https://vizzuality.atlassian.net/browse/MARXAN-1653)

### DIFF
--- a/api/apps/api/src/modules/clone/import/adapters/entities/imports.api.entity.ts
+++ b/api/apps/api/src/modules/clone/import/adapters/entities/imports.api.entity.ts
@@ -1,3 +1,4 @@
+import { ExportEntity } from '@marxan-api/modules/clone/export/adapters/entities/exports.api.entity';
 import { ResourceKind } from '@marxan/cloning/domain';
 import {
   Column,
@@ -51,6 +52,9 @@ export class ImportEntity {
   })
   isCloning!: boolean;
 
+  @Column({ type: 'uuid', name: 'export_id' })
+  exporttId!: string;
+
   @OneToMany(() => ImportComponentEntity, (component) => component.import, {
     cascade: true,
   })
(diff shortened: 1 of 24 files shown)


angelhigueraacid:
Hi @hotzevzl ! In case there are any imports already created in the database, I have set to them a random exportId `api/apps/api/src/migrations/api/1655657559840-AdsExportIdToImportsTable.ts` . I don't know if this practice is good or we should apply a different strategy

api/apps/api/src/modules/clone/import/adapters/entities/imports.api.entity.ts
```diff
@@ -51,6 +52,9 @@ export class ImportEntity {
   })
   isCloning!: boolean;
 
+  @Column({ type: 'uuid', name: 'export_id' })
+  exporttId!: string;
```
hotzevzl, line 56:
```suggestion
  exportId!: string;
```

hotzevzl:
> Hi @hotzevzl ! In case there are any imports already created in the database, I have set to them a random exportId `api/apps/api/src/migrations/api/1655657559840-AdsExportIdToImportsTable.ts` . I don't know if this practice is good or we should apply a different strategy

@angelhigueraacid should we "just" drop all the imports? what you suggest in the PR will work for now and we're unlikely to ever run into issues given that what is "live" is just a very small set of test/demo projects with all the caveats of the live platform being a beta and so on, but... maybe we can consider all the current imports as "done" and purge them all - what do you think?

(btw please have a look at the migration file name - should be `AddExportId...etc` not `AdsExportId...`)

hotzevzl (approved):
Other than the discussion about what to do with the new `imports` FK, all the rest looks good, thanks!

angelhigueraacid:
> > Hi @hotzevzl ! In case there are any imports already created in the database, I have set to them a random exportId `api/apps/api/src/migrations/api/1655657559840-AdsExportIdToImportsTable.ts` . I don't know if this practice is good or we should apply a different strategy
> 
> @angelhigueraacid should we "just" drop all the imports? what you suggest in the PR will work for now and we're unlikely to ever run into issues given that what is "live" is just a very small set of test/demo projects with all the caveats of the live platform being a beta and so on, but... maybe we can consider all the current imports as "done" and purge them all - what do you think?
> 
> (btw please have a look at the migration file name - should be `AddExportId...etc` not `AdsExportId...`)

Okey! Seems good to consider every import as just done and drop all of them! I will modify the migration (its file name too )

PR #1147, opened 2022-06-23
explicitly select columns we need in response [MARXAN-1665]

hotzevzl:
https://vizzuality.atlassian.net/browse/MARXAN-1665


rubvalave (approved):
LGTM!

PR #1148, opened 2022-06-23
downgrade PostgreSQL to v13.x [MARXAN-1664]

hotzevzl:
**DO NOT MERGE**: even if/when we are happy with these changes after having magnificently verified everything for functional and performance regressions, we should keep these changes in a parallel universe of a PR stack that includes changes needed for infra/tf, so that we can keep fast-tracking urgent changes into `develop` in the meanwhile for any fixes that may be needed for instances where demos/training sessions are running.

https://vizzuality.atlassian.net/browse/MARXAN-1664

Part of spike to check if all works as expected (bar any tolerable performance regressions) if we run both apidb and geodb on PostgreSQL v13, in preparation to moving the database instances in k8s setups to the Azure database for PostgreSQL Flexible service, which currently supports PostgreSQL up to version 13.

We have been using PostGIS v3.1 so far, and this PR also bumps up PostGIS to v3.2 (as this is the version currently supported in the PostgreSQL v13 Azure managed service).

In theory, functional regressions are not expected with this downgrade of PostgreSQL, as the key functionality needed for vector tiles is provided by PostGIS v3.x and we're keeping (and actually bumping up) this extension's version.

Some performance regressions are expected, instead, with the PostgreSQL downgrade at least in terms of some lesser parallelization of queries in some cases: the assumption here is however that we may be trading some tolerable performance regressions in some cases with the ability to run the PostgreSQL databases on the Azure managed service, which provide both functional advantages (backups, restores, other managed affordances) as well as the ability to potentially use read replicas in
the future to better balance read loads.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ details.
 
 - [NodeJS](https://nodejs.org/en/) v14.18 and v16.14
 - [Yarn](https://classic.yarnpkg.com/) v1
-- [PostgreSQL](https://www.postgresql.org/) v14
+- [PostgreSQL](https://www.postgresql.org/) v13
 - [Postgis](https://postgis.net/) v3
 - [Redis](https://redis.io/) v6
 - A [Sparkpost](https://www.sparkpost.com/) account
--- a/api/apps/geoprocessing/src/migrations/geoprocessing/1611221157285-InitialGeoDBSetup.ts
+++ b/api/apps/geoprocessing/src/migrations/geoprocessing/1611221157285-InitialGeoDBSetup.ts
@@ -6,7 +6,6 @@ export class InitialGeoDBSetup1611221157285 implements MigrationInterface {
     // Only CREATEDB privilege required in 13+ rather than SUPERUSER (ht @agnessa)
     if (await PostgreSQLUtils.version13Plus()) {
       await queryRunner.query(`
-    CREATE EXTENSION IF NOT EXISTS tablefunc;
     CREATE EXTENSION IF NOT EXISTS plpgsql;
     CREATE EXTENSION IF NOT EXISTS postgis;
     CREATE EXTENSION IF NOT EXISTS postgis_raster; -- OPTIONAL
@@ -201,7 +200,6 @@ export class InitialGeoDBSetup1611221157285 implements MigrationInterface {
       DROP EXTENSION postgis_raster; -- OPTIONAL
       DROP EXTENSION postgis;
       DROP EXTENSION plpgsql;
-      DROP EXTENSION tablefunc;
       `);
     }
   }
--- a/api/apps/geoprocessing/test/tiles/protected-area.fixtures.ts
+++ b/api/apps/geoprocessing/test/tiles/protected-area.fixtures.ts
@@ -73,7 +73,14 @@ export const getFixtures = async () => {
       );
 
       expect(customFeature.length).toEqual(1);
-      expect(features.length).toEqual(17);
+      /**
+       * Note: this was expected toEqual(17) originally; when switching from
+       * PostgreSQL 14+PostGIS 3.1 to PostgreSQL 13+PostGIS 3.2 (to match the
+       * versions available in Azure Database for PostgreSQL - Flexible Server),
+       * processing of the features leads to a count of 18 features (this is
+       * due, specifically, to the PostGIS 3.1->3.2 part of the down-up-grade).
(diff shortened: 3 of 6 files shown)


tiagojsag:
LGTM I was waiting on a reply from their side to move forward with the infra work on our end.

tiagojsag:
btw we should validate that Azure Postgres thingy has the other plugins we need.

hotzevzl:
> btw we should validate that Azure Postgres thingy has the other plugins we need.

```
marxan-geo-api=# select extname,extversion from pg_extension;
        extname         | extversion
------------------------+------------
 plpgsql                | 1.0
 postgis                | 3.2.1
 postgis_topology       | 3.2.1
 fuzzystrmatch          | 1.1
 postgis_tiger_geocoder | 3.2.1
 pgcrypto               | 1.3
 tablefunc              | 1.0
 postgis_raster         | 3.2.1
(8 rows)
```

In Azure Flexible Server for PostgreSQL v13 (https://docs.microsoft.com/en-us/azure/postgresql/flexible-server/concepts-extensions#postgres-13-extensions):

```
pgpgsql: 1.0
postgis: 3.2.0
postgis_topology: 3.2.0
fuzzystrmatch: 1.1
postgis_tiger_geocoder: 3.2.0
pgcrypto: 1.3
tablefunc: not available
postgis_raster: 3.2.0
```

As far as I can see, `tablefunc` is not actually used (none of its functions are mentioned anywhere in the codebase: https://www.postgresql.org/docs/current/tablefunc.html) - so I think I can remove this extension as part of this PR.

@aagm do you see any issues with this?

hotzevzl:
so @aagm @tiagojsag  I don't understand this:

- postgres14.x/postgis3.x: https://github.com/Vizzuality/marxan-cloud/actions/runs/2596327856 (branch https://github.com/Vizzuality/marxan-cloud/tree/spike/api/MARXAN-1664_downgrade-to-postgresql-13-wtf-edition)
- postgres13.x/postgis3.x: https://github.com/Vizzuality/marxan-cloud/runs/7132228685?check_suite_focus=true#step:4:4995 (branch https://github.com/Vizzuality/marxan-cloud/tree/spike/api/MARXAN-1664_downgrade-to-postgresql-13)

what _seems to fail_ in several tests is an attempt to insert into a calculated column, for which the error in PG13 is (when doing something similar via `psql`):

```
ERROR:  cannot insert into column "hash"
DETAIL:  Column "hash" is a generated column.
```

however, the same operation does likewise (and thankfully) lead to an error in PG14:

```
ERROR:  cannot insert a non-DEFAULT value into column "hash"
DETAIL:  Column "hash" is a generated column.
```

And in both cases, given the hint provided by PG14, inserting into a generated column with a `DEFAULT` value works, which is where I am getting lost, because these tests fail with the only difference being the PostgreSQL server version and nothing else in the code (unless of course TypeORM or the `pg` driver do things differently between pg13 and pg14).

So, this is more of an update on this task: in case there's something obvious that jumps to your eyes, please let me know, otherwise I'll get into overdrive mode trying to understand what's breaking so miserably.

Lastly, this is _not_ "just" a failure artifact related to tests only: given that all the failing tests are related to project cloning/export/import, I could confirm that these operations succeed (in general 🙄) in instances running on PG14 but not on instances (aka my local env, for now) running on PG13.

hotzevzl:
@aagm @tiagojsag pragmatically trying to close these PG14->PG13 downgrade checks, I am personally happy-ish to go ahead with the down-up-grade by updating the only failing expectation: 775139167981b943a2fb2632b1b0ba1c01fa3372

It would be worth understanding why PostGIS 3.2 calculates things differently than PostGIS 3.1 (tests pass, untouched, when using PostgreSQL13+PostGIS3.1 instead of PostgreSQL13+PostGIS3.2 as base image: https://github.com/Vizzuality/marxan-cloud/actions/runs/2611259505) - but 1. I don't think it's a battle worth fighting now, 2. I kinda trust that PostGIS does the right thing in any case, and 3. my focus in this spike was to make sure that a downgrade to PostgreSQL 13 would not cause operational issues or require us to rewrite un-manageable portions of the application's code, so I think we could live with discrepancies in how some things are calculated.

If needed (from a science point of view) we can try to dig further into calculation discrepancies later.

I'd merge this PR at the end of the day today unless either of you thinks we should take the originally failing test as a red flag.

PR #1150, opened 2022-06-24
MARXAN-1647-download-user-data-csv

rubvalave:
-Adds endpoint to download user data in CSV format.
-Adds tests for the endpoint.


api/apps/api/src/modules/users/users.controller.ts
```diff
@@ -49,6 +52,9 @@ import { UpdateUserPasswordDTO } from './dto/update.user-password';
 import { IsMissingAclImplementation } from '@marxan-api/decorators/acl.decorator';
 import { PlatformAdminEntity } from './platform-admin/admin.api.entity';
 import { isLeft } from 'fp-ts/lib/Either';
+import { Writable } from 'utility-types';
```
hotzevzl, line 55: unused import, please remove (also: it would be `import { Writable } from 'stream';` for this use case)

hotzevzl (approved):

PR #1152, opened 2022-06-27
Feat/add api feature id to scenario features data

angelhigueraacid:
The main change this PR introduces is adding `api_feature_id` on `scenario_features_data` table. With this new column, we will be able to know which feature generated the `scenario_features_data` rows as with the current implementation, `feature_id` in `features_data` table holds the `baseFeautrId`. Features obtained by copying operation should give the same result as `feature_id` and `api_feature_id` will hold the same value.

Because we are adding this new column, queries that are grouping by `feature_id` in `features_data` now will group by `api_features_id` in `scenario_features_data`.

The main queries that have changed are:

- In api/apps/api/src/modules/scenarios-features/copy/copy-query.service.spec.ts , when copying a feature the `api_feature_id` value will be set with the `feature_id` value from `features_data` (when copying a feature, the `baseFeatureId` and the `apiFeatureId` are the same)
- In api/apps/api/src/modules/scenarios-features/move-data-from-preparation.handler.ts, when moving `scenario_features_preparation` data to `scenarios_features_data` we must move the new column.
- In api/apps/api/src/modules/scenarios/input-files/spec.dat.service.ts, now we select `api_feature_id` instead of `feature_id` from features_data and also we make the inner join filtering by `scenario_id` (this has been added since the integer `featureId` in `scenario_features_data` could be duplicated, for example when cloning a project )
- In api/apps/geoprocessing/src/export/pieces-exporters/scenario-features-data.piece-exporter.ts , now instead of exporting just `feature_id` from `features_data` we also need to export `api_feature_id` from `scenario_features_data` (this changes affects obviously the import of this file)
- In api/apps/geoprocessing/src/modules/scenario-planning-units-features-aggregate/scenario-planning-units-features-aggregate-processor.ts , now we select `api_feature_id` instead of `feature_id`
- In api/libs/features/src/scenario-features-gap-data.geo.entity.ts. Now we group by `api_feature_id` instead of `feature_id`
- In api/libs/features/src/scenario-features-output-gap-data.geo.entity.ts. Now we group by `api_feature_id` instead of `feature_id`

It is important to note that, no change has been made to features obtained by `split` operation, once the feature is created in features table when splitting  we should use this id as the api_feature_id. 

### Feature relevant tickets

[area of scenario features data originated from a split can't be calculated](https://vizzuality.atlassian.net/browse/MARXAN-1668)


api/apps/geoprocessing/src/migrations/geoprocessing/1655901622494-AddApiFeatureIdToScenarioFeaturesData.ts
```diff
@@ -0,0 +1,36 @@
...
+    await queryRunner.query(`
+      ALTER TABLE scenario_features_data
+      ADD COLUMN api_feature_id uuid NOT NULL;
+    `);
+
+    await queryRunner.query(`
+      alter table scenario_features_preparation
+      ADD COLUMN api_feature_id uuid NOT NULL;
+    `);
+  }
```
hotzevzl, line 19: I think it was me who advised that nuking `scenario_features_data` would be ok, but now seeing this in code, I'm not so sure anymore - true, there's no _real_ production instance, but I think we should try to disrupt existing "public" instances as little as possible.

In a way, given that so far we only have `copy/v1` operations used in the wild (if any `split/v1` has ever been attempted, we can safely consider these as not properly done and discardable), this can work to our advantage: what about adding the `api_feature_id` as nullable first, then setting it to the actual expected value via something like `update scenario_features_data set api_feature_id = (select fd.feature_id from scenario_features_data sfd inner join features_data fd on fd.id = sfd.feature_class_id where sfd.id = scenario_features_data.id)`?

and finally setting the new column as non-nullable.

in theory we should not need to repeat this for `scenario_features_preparation` unless some stale data was left around in an instance's geodb through some preparation step that was interrupted (though IIRC prep is done in a db transaction so it may be guaranteed to complete or be rolled back without leaving stale data around).

hotzevzl (changes requested):

hotzevzl (approved):

PR #1155, opened 2022-07-01
MARXAN-764-onCascade-delete-scenarios-resources

rubvalave:
-Adds proper cleaning for scenarios unused resources in geoprocessing db.
-Adapts scenario deletion endpoint to trigger jobs on geoprocessing.

NOTE: I did not include the migrations to do the following points indicated on the original task because there were already migrations that did the same from some months ago:

- > add FK from output_scenarios_pu_data.scenario_pu_id to scenarios_pu_data.id with cascade (on update, on delete), both via migration and in the relevant entities - this will cascade the delete of the previous point to the relevant output_scenarios_pu_data rows)
- > add FK from scenarios_pu_cost_data.scenarios_pu_data_id to scenarios_pu_data.id with cascade, as above


api/apps/api/src/modules/projects/delete-project/schedule-project-unused-resources-cleanup.handler.ts
```diff
@@ -3,7 +3,7 @@ import { Inject, Logger } from '@nestjs/common';
 import { CommandHandler, IInferredCommandHandler } from '@nestjs/cqrs';
 import { Queue } from 'bullmq';
 import { ScheduleCleanupForProjectUnusedResources } from './schedule-project-unused-resources-cleanup.command';
-import { unusedResourcesCleanupQueueToken } from './unused-resources-cleanup-queue.provider';
+import { unusedResourcesCleanupQueueToken } from '../../../../../../libs/unused-resources-cleanup/src/unused-resources-cleanup-queue.provider';
```
hotzevzl, line 6: `@marxan/...` import

api/apps/api/src/modules/scenarios/delete-scenario/delete-scenario.handler.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+  fixtures.ThenAScenarioDeletedEventIsEmitted(scenarioId);
+});
+
+it('fails to delete a scenario ', async () => {
+  const scenarioId = fixtures.GivenScenarioExits();
+  fixtures.GivenDeleteOperationFails();
+
+  const result = await fixtures.WhenAScenarioIsDeleted(scenarioId);
+
+  fixtures.ThenDeleteScenarioFailedIsReturned(result);
```
hotzevzl, line 41: very picky detail - I'd say this is a `when` (when a scenario deletion is requested and it fails...)

api/apps/api/src/modules/scenarios/scenarios.service.ts
```diff
@@ -112,6 +112,10 @@ import { InjectEntityManager } from '@nestjs/typeorm';
 import { apiConnections } from '@marxan-api/ormconfig';
 import { EntityManager } from 'typeorm';
 import { UserId } from '@marxan/domain-ids';
+import {
+  DeleteScenario as ScenarioUnusedResources,
```
hotzevzl, line 116: picky detail - I am unclear about this import alias... maybe you meant `DeleteScenarioUnusedResources`? but even then, I am not sure I understand the rationale (also, it's late on a Friday afternoon, so I may not understanding *anything*)

hotzevzl: (especially because we do have a `ScenarioUnusedResources` class, in any case) 🤔

hotzevzl (approved):
thanks @rubvalave - all looks good, especially if you are sure it does what it says on the tin 😅 

just some tiny notes about names, please check and decide.

angelhigueraacid (approved):

PR #1157, opened 2022-07-05
fix: wrong update of exisiting scenario features data in migration

angelhigueraacid:
This PR fixes the update of existing `scenario_features_data` rows when adding `api_feature_id`

### Feature relevant tickets
[set api_feature_id for existing features: fix migration](https://vizzuality.atlassian.net/browse/MARXAN-1674)


hotzevzl (approved):
thanks @angelhigueraacid - this saves my planned deployments for the day (and unbreaks local dev environments if any dev has features by the time this migration tries to run 😬)

PR #1158, opened 2022-07-05
Feat/marxan 946 update puvspr.dat entries

angelhigueraacid:
This PR changes the way `puvspr.dat` file is generated. Now, the calculations are made once a `created specification` is made, for each of the features selected in the specification, the calculation will be performed and stored, in order to retrieve this results when creating puvspr.dat file.

In case the project is a legacy project, no calculations will be performed for any feature within the specification. Also, this PR adds `SplitCreateFeatures` used for creating one feature per subset contained within a split feature.
 
### Feature relevant tickets

[update puvspr.dat entries only for changed/added/deleted features](https://vizzuality.atlassian.net/browse/MARXAN-946)


api/apps/api/test/integration/single-config-feature-value.hasher.e2e-spec.ts
```diff
@@ -0,0 +1,113 @@
...
+      sut.getHashAndStrippedConfigFeature(input),
+    ThenSingleSplitConfigFeatureValueIsHashed: ({
+      hash,
+      canonical,
+    }: HashAndStrippedConfigFeature) => {
+      expect(hash).toBeDefined();
+      expect(hash.length).toBeGreaterThan(0);
+    },
+    ThenSingleSplitConfigFeatureValueAreTheSame: (
+      fisrtResult: HashAndStrippedConfigFeature,
```
hotzevzl, line 100:
```suggestion
      firstResult: HashAndStrippedConfigFeature,
```

api/apps/geoprocessing/src/migrations/geoprocessing/1655819134885-AddsPuvsprCalculationsTable.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export class AddPuvsprCalculationsTable1655819134885
+  implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+      CREATE TABLE "puvspr_calculations"(
+        "id" uuid PRIMARY KEY DEFAULT gen_random_uuid(),
+        "project_id" uuid NOT NULL,
+        "feature_id" uuid NOT NULL,
+        "pu_id" integer NOT NULL,
```
hotzevzl, line 11: can we add a FK to `projects_pu.id` with cascade so that whenever the corresponding `projects_pu` row is deleted we also get the relevant `puvspr_calculations` rows deleted, without having to add another task to the GC code to clean up this table when the parent is deleted?

(if keeping as is - and in any case we should do this in a separate PR, if going with the FK - then we should also update the regular cleanups on project deletion to also GC the orphaned rows from this table - again, if at all, in a separate PR)

api/libs/puvspr-calculations/src/puvspr-calculations.service.ts
```diff
@@ -0,0 +1,143 @@
...
+  ) {
+    /**
+     * @TODO further performance savings: limiting scans to planning_units_geom
+     * by partition (we need to get the grid shape from the parent project); use
+     * && operator instead of st_intersects() for bbox-based calculation of
+     * intersections.
+     *
+     * @TODO Calculate `amount` correctly from `amount_from_legacy_project`. The
+     * initial implementation is simply a placeholder so that we can use
+     * `amount_from_legacy_projects` in legacy project piece importers.
```
hotzevzl, line 48: this can be removed as you've basically taken care of it via the `computeLegacyAmountPerPlanningUnit` method below

api/libs/puvspr-calculations/src/puvspr-calculations.service.ts
```diff
@@ -0,0 +1,143 @@
...
```
hotzevzl, line 102: this can be removed, as above

hotzevzl (approved):
thanks @angelhigueraacid!

all looks good - I have left a bunch of typo and minor comments.

the two main bits to consider more carefully are my comment about `PuvsprDatLegacyProject.getLegacyAndMarxanFeatures()` and on whether we should add a FK from the calculations table to `projects_pu` (at the expense of... an extra join or queries to get the value to write?)

api/apps/api/src/modules/scenarios/input-files/puvspr.dat.processor/puvspr.dat.legacy-project.ts
```diff
@@ -0,0 +1,89 @@
...
+      .select('id')
+      .from(GeoFeature, 'features')
+      .where('is_legacy = true')
+      .andWhere('id IN (:...featureIds)', { featureIds })
+      .execute();
+
+    const legacyFeatureIds = legacyFeatures.map(({ id }) => id);
+    const marxanFeatureIds = featureIds.filter(
+      (id) => !legacyFeatureIds.includes(id),
+    );
```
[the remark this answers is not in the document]
angelhigueraacid: It's is not really intentional, i would like to change this to use the features repo instead and find all features `where id IN featureIds`, this way, we will get if the feature is legacy or not and also ensure that we only use featureIds that exists. Thanks for noticing it, didn't even consider that case.

api/apps/geoprocessing/src/migrations/geoprocessing/1655819134885-AddsPuvsprCalculationsTable.ts
angelhigueraacid: So instead of using `pu_id integer`, you would prefer using the `id` of the `projects_pu` table? If it is a better solution we can do it, in case we go with this change, when generating puvspr.dat file we would need to make an extra query for getting the `integer puids` for both legacy and marxan projects, also this change will affect a little bit the export and import of this data.

PR #1167, opened 2022-07-11
remove feature tags [MARXAN-1679]

hotzevzl:
https://vizzuality.atlassian.net/browse/MARXAN-1679


rubvalave (approved):
LGTM! It is just a case of removing completely all the references to a specific column everywhere along the app, right?

hotzevzl:
> LGTM! It is just a case of removing completely all the references to a specific column everywhere along the app, right?

yep, and the related enums (both in db and TS code).

PR #1168, opened 2022-07-12
mark as failed stuck async jobs [MARXAN-1554]

angelhigueraacid:
This PR adds async jobs garbage collector. So in case an async job gets stuck , a failed api events for the given async job will be send. We will check wether an async jobs is stuck or not when users logs in for every project and scenario where the user has any role.

For every kind of async job, a default 8 hours interval is set to define if the async job is stuck or not, this behavior can be easily updated for each individual async job kind changing the implementation of `getMaxHoursForAsyncJob` for the corresponding async job.

### Feature relevant tickets

[If an async job gets "stuck", it should be marked as failed automatically](https://vizzuality.atlassian.net/browse/MARXAN-1554)

### DIFF
--- a/api/apps/api/src/modules/async-jobs-garbage-collector/async-jobs/async-job.spec.ts
+++ b/api/apps/api/src/modules/async-jobs-garbage-collector/async-jobs/async-job.spec.ts
@@ -122,7 +122,7 @@ class AsyncJobFake extends AsyncJob {
   protected getAllAsyncJobStates(): API_EVENT_KINDS[] {
     return this.allAsyncJobStates;
   }
-  protected getEndAsynJobStates(): API_EVENT_KINDS[] {
+  protected getEndAsyncJobStates(): API_EVENT_KINDS[] {
     return this.endAsyncJobStates;
   }
   protected getFailedAsyncJobState(
--- a/api/apps/api/src/modules/async-jobs-garbage-collector/async-jobs/async-job.spec.ts
+++ b/api/apps/api/src/modules/async-jobs-garbage-collector/async-jobs/async-job.spec.ts
@@ -1,0 +1,136 @@
+import { ApiEventsService } from '@marxan-api/modules/api-events';
+import { ApiEventByTopicAndKind } from '@marxan-api/modules/api-events/api-event.topic+kind.api.entity';
+import { API_EVENT_KINDS } from '@marxan/api-events';
+import { FixtureType } from '@marxan/utils/tests/fixture-type';
+import { Injectable, NotFoundException } from '@nestjs/common';
+import { Test } from '@nestjs/testing';
+import { v4 } from 'uuid';
+import { AsyncJob } from './async-job';
+
+let fixtures: FixtureType<typeof getFixtures>;
+
+beforeEach(async () => {
+  fixtures = await getFixtures();
+});
+
+it(`does not send a failed api event when async job has not started`, async () => {
+  fixtures.GivenAsyncJobHasNotStarted();
+  await fixtures.WhenSendingFailedApiEventForStuckAsyncJob();
+  fixtures.ThenNoApiEventHasBeenSent();
+});
+
+it(`does not send a failed api event for a finished async job`, async () => {
+  fixtures.GivenAsyncJobHasFinished();
+  await fixtures.WhenSendingFailedApiEventForStuckAsyncJob();
+  fixtures.ThenNoApiEventHasBeenSent();
+});
+
+it(`does not send a failed api event for a non stuck async job`, async () => {
+  fixtures.GivenAsyncJobHasNotFinished({ isStuck: false });
+  await fixtures.WhenSendingFailedApiEventForStuckAsyncJob();
+  fixtures.ThenNoApiEventHasBeenSent();
+});
+
+it(`sends a failed api event for a stuck async job`, async () => {
+  fixtures.GivenAsyncJobHasNotFinished({ isStuck: true });
+  await fixtures.WhenSendingFailedApiEventForStuckAsyncJob();
+  fixtures.ThenApiEventHasBeenSent();
+});
+
+const getFixtures = async () => {
+  const getLatestEventForTopicMock = jest.fn();
+  const createIfNotExistsMock = jest.fn();
+  const sandbox = await Test.createTestingModule({
+    imports: [],
+    providers: [
+      AsyncJobFake,
+      {
+        provide: ApiEventsService,
+        useValue: {
+          getLatestEventForTopic: getLatestEventForTopicMock,
+          createIfNotExists: createIfNotExistsMock,
+        },
+      },
+    ],
+  }).compile();
+
+  await sandbox.init();
+
+  const projectId = v4();
+
+  const sut = sandbox.get(AsyncJobFake);
+
+  function subtract2HoursFromNow() {
+    const date = new Date();
+    date.setHours(date.getHours() - 2);
+    return date;
+  }
+
+  const finishedApiEvent = API_EVENT_KINDS.project__grid__finished__v1__alpha;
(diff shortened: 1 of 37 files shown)


api/apps/api/src/modules/async-jobs-garbage-collector/async-jobs/async-job.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+  public allAsyncJobStates: API_EVENT_KINDS[] = [];
+  public endAsyncJobStates: API_EVENT_KINDS[] = [
+    API_EVENT_KINDS.project__grid__finished__v1__alpha,
+  ];
+  public failedAsyncJob = API_EVENT_KINDS.project__grid__failed__v1__alpha;
+  public maxHoursForAsyncJob = 0;
+  protected getAllAsyncJobStates(): API_EVENT_KINDS[] {
+    return this.allAsyncJobStates;
+  }
+  protected getEndAsynJobStates(): API_EVENT_KINDS[] {
```
hotzevzl, line 125:
```suggestion
  protected getEndAsyncJobStates(): API_EVENT_KINDS[] {
```

api/apps/api/src/modules/async-jobs-garbage-collector/async-jobs/async-job.ts
```diff
@@ -0,0 +1,83 @@
...
+import { Injectable, NotFoundException } from '@nestjs/common';
+import { In } from 'typeorm';
+
+@Injectable()
+export abstract class AsyncJob {
+  constructor(protected readonly apiEvents: ApiEventsService) {}
+
+  protected abstract getAllAsyncJobStates(): API_EVENT_KINDS[];
+
+  protected abstract getEndAsynJobStates(): API_EVENT_KINDS[];
```
hotzevzl, line 14:
```suggestion
  protected abstract getEndAsyncJobStates(): API_EVENT_KINDS[];
```

api/apps/api/src/modules/async-jobs-garbage-collector/async-jobs/async-job.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  public async sendFailedApiEventForStuckAsyncJob(resourceId: string) {
+    const latestApiEvent = await this.findLatestApiEvent({
+      topic: resourceId,
+      kind: In(this.getAllAsyncJobStates()),
+    });
+
+    if (!latestApiEvent) return;
+
+    const hasAsyncJobFinsihed = this.hasAsyncJobFinsihed(latestApiEvent);
```
hotzevzl, line 30:
```suggestion
    const hasAsyncJobFinished = this.hasAsyncJobFinished(latestApiEvent);
```

api/apps/api/src/modules/async-jobs-garbage-collector/async-jobs/scenario-async-jobs/cost-surface.async-job.ts
```diff
@@ -0,0 +1,43 @@
...
+      API_EVENT_KINDS.scenario__costSurface__finished__v1_alpha1,
+    ];
+  }
+  getFailedAsyncJobState(): CostSurfaceApiEvents {
+    /*
+    At the moment in the codebase, we are not using this two api events:
+    API_EVENT_KINDS.scenario__costSurface__shapeConverted__v1_alpha1 
+    API_EVENT_KINDS.scenario__costSurface__shapeConversionFailed__v1_alpha1
+    In case we start using them, we migh have to do add new state to getEndAsynJobStates()
+    and check the latestApitEvent when executing getFailedAsyncJobState()
```
hotzevzl, line 36: ❤️

api/apps/api/src/modules/async-jobs-garbage-collector/garbage-collect-async-jobs.handler.ts
```diff
@@ -0,0 +1,59 @@
...
+        this.projectAsyncJobs.sendFailedApiEventsForStuckAsyncJobs(projectId),
+      ),
+    );
+
+    const scenariosIds = await this.getScenariosIdsOf(userId);
+    await Promise.all(
+      scenariosIds.map((scenarioId) =>
+        this.scenarioAsyncJobs.sendFailedApiEventsForStuckAsyncJobs(scenarioId),
+      ),
+    );
```
hotzevzl, line 41: this is very tempting, and is ok - thinking however about broader use of the platform later on, I wonder if we could trade some of the performance gains we can hope when count of projects and scenarios is rather low (by firing all the promises at once waiting for all the async work to be done) for some control over potential db query stampedes, and use a `for await...` here instead (accepting to process each project and scenario sequentially)

api/apps/api/src/modules/async-jobs-garbage-collector/scenario-async-jobs.garbage-collector.ts
```diff
@@ -0,0 +1,60 @@
...
+    private readonly planningUnitsInclusionAsyncJob: PlanningUnitsInclusionAsyncJob,
+    private readonly protectedAreasAsyncJob: ProtectedAreasAsyncJob,
+    private readonly runAsyncJob: RunAsyncJob,
+    private readonly scenarioCloneAsyncJob: ScenarioCloneAsyncJob,
+    private readonly scenarioExportAsyncJob: ScenarioExportAsyncJob,
+    private readonly scenarioImportAsyncJob: ScenarioImportAsyncJob,
+    private readonly specificationAsyncJob: SpecificationAsyncJob,
+  ) {}
+  public async sendFailedApiEventsForStuckAsyncJobs(scenarioId: string) {
+    await Promise.all([
```
hotzevzl, line 34: since this is a more or less fixed list, and even if we want to go for a `for await` approach for projects and scenarios, we could still keep this `Promise.all()` here - or maybe err on the side of caution and process each job type sequentially too, since we can kind of afford to take our time as all of this is scheduled to happen behind the scenes in any case, so whenever it finishes it will be ok, and we can instead prioritise trying to avoid overloading the db (especially when the `api_events` table may have grown a lot)

hotzevzl (commented):
thanks @angelhigueraacid - great PR 🙌🏼 

my only concern is about trying to do too much at once - see my inline notes on whether we should consider processing each project and scenario check sequentially, and each job kind check within each of them sequentially too, since we shouldn't be too concerned about the checks completing as soon as possible (and conversely if we unleash potentially hundreds of checks at very short interval on a large `api_events` table, we may actually overload the db 🤔

api/apps/api/src/modules/async-jobs-garbage-collector/garbage-collect-async-jobs.handler.ts
angelhigueraacid: I 

> thanks @angelhigueraacid - great PR 🙌🏼
> 
> my only concern is about trying to do too much at once - see my inline notes on whether we should consider processing each project and scenario check sequentially, and each job kind check within each of them sequentially too, since we shouldn't be too concerned about the checks completing as soon as possible (and conversely if we unleash potentially hundreds of checks at very short interval on a large `api_events` table, we may actually overload the db 🤔

I will gladly change the code to process project and scenarios one by one and each async job kind within each of the sequentially. I had this same concern and i believe this change will make it safer. Thanks for the suggestion

angelhigueraacid:
> thanks @angelhigueraacid - great PR 🙌🏼
> 
> my only concern is about trying to do too much at once - see my inline notes on whether we should consider processing each project and scenario check sequentially, and each job kind check within each of them sequentially too, since we shouldn't be too concerned about the checks completing as soon as possible (and conversely if we unleash potentially hundreds of checks at very short interval on a large `api_events` table, we may actually overload the db 🤔

I will gladly change the code to process project and scenarios one by one and each async job kind within each of the sequentially. I had this same concern and i believe this change will make it safer. Thanks for the suggestion

PR #1170, opened 2022-07-13
MARXAN-1484-minimum-password-strength

rubvalave:
-Adds small utility to check that passwords used are over 80 in entropy levels.
-Adds small test to check lower than 80 password can't be used.

### DIFF
--- a/api/apps/api/src/modules/authentication/authentication.service.ts
+++ b/api/apps/api/src/modules/authentication/authentication.service.ts
@@ -26,8 +26,8 @@ import { Mailer } from '@marxan-api/modules/authentication/password-recovery/mai
 import { API_EVENT_KINDS } from '@marxan/api-events';
 import { EventBus } from '@nestjs/cqrs';
 import { UserLoggedIn } from './user-logged-in.event';
-import { passwordEntropyValidation } from '@marxan-api/utils/passwordValidation.utils';
-import { isLeft } from 'fp-ts/lib/These';
+import { isStringEntropyHigherThan } from '@marxan-api/utils/passwordValidation.utils';
+import { isLeft } from 'fp-ts/lib/Either';
 
 /**
  * Access token for the app: key user data and access token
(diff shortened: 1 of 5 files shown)


api/apps/api/src/modules/authentication/authentication.service.ts
```diff
@@ -23,6 +24,8 @@ import { v4 } from 'uuid';
 import * as ApiEventsUserData from '@marxan-api/modules/api-events/dto/apiEvents.user.data.dto';
 import { Mailer } from '@marxan-api/modules/authentication/password-recovery/mailer';
 import { API_EVENT_KINDS } from '@marxan/api-events';
+import { passwordEntropyValidation } from '@marxan-api/utils/passwordValidation.utils';
+import { isLeft } from 'fp-ts/lib/These';
```
hotzevzl, line 28:
```suggestion
import { isLeft } from 'fp-ts/lib/Either';
```

hotzevzl (approved):
thanks @rubvalave - it does (or should do) what it says on the tin - I would suggest:

- making the threshold more easily configurable at consumption point, even if hardcoded

for example, define it as a class constant in `AuthenticationService`, or even outside of the class declaration as a top-level constant in that same file (`authentication.service.ts`)

- passing on this threshold to the utility function
- using a verb for the function name (`isStringEntropyHigherThan(value: number)`), for example
- importing the `fast-password-entropy` via a top-level `import`

rubvalave:
I don't exactly understand what you meant with this:

- importing the `fast-password-entropy` via a top-level import

You meant importing the library in the service as well as I did with the threshold? I apply the other changes, but I do not know what this meant and the purpose of it. Do you mean importing it in the service and passing it to the little utility function I added?

hotzevzl:
> I don't exactly understand what you meant with this:
> 
>     * importing the `fast-password-entropy` via a top-level import
> 
> 
> You meant importing the library in the service as well as I did with the threshold? I apply the other changes, but I do not know what this meant and the purpose of it. Do you mean importing it in the service and passing it to the little utility function I added?

sure, my comment was indeed not clear - what does top-level even mean in general? 🤦🏼 

i just meant moving the import to the top of the utility file, and using an `import` instead of a `require` - like in f8ae3a66f73eaa70f74403f0b187e8b106f0c162

but ignore this - it's just a style thingy.

would you mind rebasing/fixing conflicts and merging, please?

PR #1183, opened 2022-07-15
Refactor infrastructure to use Azure Flexible SQL as server db

tiagojsag:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


infrastructure/base/modules/database/main.tf
```diff
@@ -0,0 +1,39 @@
+locals {
+  postgres_secret_json = {
+    username = "postgres"
+    password = random_password.postgresql_admin_generator.result
+  }
```
hotzevzl, line 5: should we rely on the stock `postgres` user?
in practice, this should be equivalent to the current setup where we create an ad-hoc user, except perhaps that `postgres` may have additional permissions granted via `pg_hba` (on the `postgres` db, for example, though in practice I can successfully `\c postgres` as `marxan-api` user already, in a local db), and also we would forego (if that's an issue to start with - probably not) the separation of users and permissions across `apidb` and `geodb` when they are within the same physical PostgreSQL server.

tl;dr I think it's ok, just listing some things to cross check.

hotzevzl (approved):

infrastructure/base/modules/database/main.tf
hotzevzl: ah right, this is the `base` layer - all good then.

infrastructure/base/vars/terraform.tfvars
```diff
@@ -7,3 +7,10 @@ github_org               = "Vizzuality"
 github_repo              = "marxan-cloud"
 github_production_branch = "main"
 github_staging_branch    = "staging"
+key_vault_access_users  = ["tiago.garcia_vizzuality.com#EXT#", "andrea.rota_vizzuality.com#EXT#"]
+
+production_db_instance_size = "GP_Standard_D32d_v4"
```
hotzevzl, line 12: @tiagojsag just noticed this while trying to remember how many cores I could try to max out: shouldn't this be `GP_Standard_D32ds_v4`?

PR #1188, opened 2022-07-19
feat: identify expired resources in exports table [MARXAN-1649]

rubvalave:
- Adds ExportCleanup module and service to API.
- Adds config interval to check for validity of creation of exports.
- Adds job that identifies expired exports based on time, its presence on the `published_projects` table and the scenario exports.

### DIFF
--- a/api/apps/api/src/modules/export-cleanup/export-cleanup.service.ts
+++ b/api/apps/api/src/modules/export-cleanup/export-cleanup.service.ts
@@ -5,7 +5,6 @@ import { EntityManager } from 'typeorm';
 import { ExportCleanup } from './export-cleanup';
 import { apiConnections } from '@marxan-api/ormconfig';
 import { AppConfig } from '@marxan-api/utils/config.utils';
-import { execSync } from 'child_process';
 
 const validityIntervalInHours = AppConfig.get<string>(
   'storage.cloningFileStorage.artifactValidityInHours',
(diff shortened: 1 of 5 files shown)


api/apps/api/src/modules/export-cleanup/export-cleanup.service.ts
```diff
@@ -0,0 +1,52 @@
...
+    @InjectEntityManager(apiConnections.default)
+    private readonly apiEntityManager: EntityManager,
+  ) {}
+
+  async identifyExpiredResources() {
+    const expiredProjectExportsIds = await this.apiEntityManager.query(`
+      SELECT e.id FROM exports e
+        WHERE e.resource_kind = 'project' AND 
+        (AGE(NOW(), e.created_at) <  INTERVAL ${validityIntervalInHours} HOUR OR
+        e.resource_id NOT IN (SELECT id FROM published_projects));
```
hotzevzl, line 26: this should be
```suggestion
        e.resource_id IN (SELECT id FROM published_projects));
```
(again apologies for having gotten my query backwards in the task description)

hotzevzl (approved):
thanks, and sorry for the confusion about what's valid in the original task description 😢 

let's merge so you can avoid stashing the next PR on top of this - but I think this PR will need a little update as part of your next one:

- `identifyExpiredResources()`: in practice, what we want to do (and do) here is identifying _still valid_ resources, right?
- if so, let's update the method name (and `expired<Entity>ExportIds` consts, and whatever else may be named backwards within the method)
- nitpick, I always encourage folks to feel a hint of panic when looking at a query string within a template string: in this case there's no user input involved (`export-cleanup.service.ts`, L25), but I recommend to _always_ use query params by default, and only opt out and fall back to template strings if and only if we need to assemble query strings from complex logic (and even then, in most cases it may be possible to avoid template strings, and if really needing to use them, there should be clear, strict check of whatever user input may trickle into queries), or else Bobby Tables
- please check the `NOT` in the part of the query that keeps published projects

PR #1190, opened 2022-07-20
MARXAN-1484-substitute-global-throttle-guard-for-specific-guard

rubvalave:
-Changes global guard for throttling limit for endpoint-specific guard

### DIFF
--- a/api/apps/api/src/modules/authentication/authentication.controller.ts
+++ b/api/apps/api/src/modules/authentication/authentication.controller.ts
@@ -27,16 +27,16 @@ import { SignUpDto } from './dto/sign-up.dto';
 import { UserAccountValidationDTO } from './dto/user-account.validation.dto';
 import { LocalAuthGuard } from './local-auth.guard';
 import { IsMissingAclImplementation } from '@marxan-api/decorators/acl.decorator';
-import { Throttle, ThrottlerGuard } from '@nestjs/throttler';
+import { SkipThrottle, Throttle, ThrottlerGuard } from '@nestjs/throttler';
 
+@UseGuards(ThrottlerGuard)
+@Throttle(25, 60)
 @Controller('/auth')
 @ApiTags('Authentication')
 export class AuthenticationController {
   constructor(private readonly authenticationService: AuthenticationService) {}
 
   @UseGuards(LocalAuthGuard)
-  @UseGuards(ThrottlerGuard)
-  @Throttle(25, 60)
   @ApiOperation({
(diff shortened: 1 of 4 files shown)


api/apps/api/src/modules/authentication/authentication.controller.ts
```diff
@@ -27,13 +27,16 @@ import { SignUpDto } from './dto/sign-up.dto';
...
+import { Throttle, ThrottlerGuard } from '@nestjs/throttler';
 
 @Controller('/auth')
 @ApiTags('Authentication')
 export class AuthenticationController {
   constructor(private readonly authenticationService: AuthenticationService) {}
 
   @UseGuards(LocalAuthGuard)
+  @UseGuards(ThrottlerGuard)
+  @Throttle(25, 60)
```
hotzevzl, line 39: I think we can throttle most of/all the endpoints in this controller: sign-up+verification, if automated, can facilitate mass bot signup (though even at 25signups per minute, we can get 36k bots signed up per day 😢, so that's a lost battle in any case).

No real need to throttle jwt refresh and sign-out, though.

But maybe we can move the guard and the throttle setting to the controller (rather than the `sign-in` endpoint handler) and this way each endpoint handler will have a 25rpm limit.

hotzevzl: or apply the guard and throttle setting to `sign-up` too, leaving the other endpoints alone

hotzevzl (approved):
thanks for preparing this PR super quickly 🚀 

It does what it needs to do, but
- please see my inline note about maybe moving the throttle stuff to the controller level or to at least throttle sign-up
- should we throttle `PasswordRecoveryController` as well? it should not be practical to exploit its endpoints via brute force, but they're still two endpoints that can be somewhat abused at least to flood the API's db with useless garbage (well, at least `requestRecovery()`, as we generate and write a token to db for each reset request
- it's ok to keep the `ThrottlerModule` imported at `AppModule` level (so it's available for use via guard decorators in other controllers, if needed), but I'd suggest to remove the global settings (which won't apply anyway), so _I think_ that by using `    ThrottlerModule.forRoot(),` by itself in `AppModule`'s `imports` array, we'll just make the module available for use in other modules where needed, but without any predefined settings (which forces users to use the `@Throttle()` guard, which should be desirable as it makes the rate limit explicit close to the point of use)

hotzevzl:
> * so _I think_ that by using `    ThrottlerModule.forRoot(),` by itself

I meant just like this, without passing any `ttl` and `limit`

PR #1210, opened 2022-07-26
Move sparkpost email templates to TF

tiagojsag:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


infrastructure/kubernetes/modules/email-templates/main.tf
```diff
@@ -0,0 +1,40 @@
+resource "sparkpost_template" "reset-password" {
+  template_id            = "marxan-reset-password"
+  content_from_email     = "no-reply@${var.domain}"
```
hotzevzl, line 3: in some cases configuring sparkpost's domain verification (via txt record, IIRC) may not be practical or possible on the domain that hosts the application itself (because policies, because stuff) but I'd leave this as is and let whoever needs to set things up to sort this out on the DNS side 😄 

(for our own soon-to-be-decommissioned instance, for example, IIRC I don't have access to the DNS config for vizzuality.com, but someone else can help and it shouldn't be a problem in practice)

hotzevzl: I mean, I'd prefer to avoid the added complexity of yet another variable, either TF or provided via env/GH vars, unless it proves to be absolutely and provably unavoidable.

hotzevzl (approved):

infrastructure/kubernetes/modules/email-templates/main.tf
tiagojsag: https://github.com/Vizzuality/marxan-cloud/pull/1210/files#diff-54dca19b6d4dcbf23f23d41a6ffef262fd6bb895f592a520d232cc5272a495f3R61

PR #1213, opened 2022-07-27
fix: split query service current_pa calculation

angelhigueraacid:
This PR fixes the calculation of `total_area` and `current_pa` made in split query service. The idea is to have the same logic in both, copy and split query services for these columns as they should be independent from the operation kind

[Strange gap analysis values with split features](https://vizzuality.atlassian.net/browse/MARXAN-1718)


angelhigueraacid:
If we compare the made to `current_pa` calculation in  `SplitQuery` to the `current_pa` calculation made in `CopyQuery`  there is still a difference. In `CopyQuery` we intersect `protected areas` with `project.bbox`, this check has not been made in `SplitQuery` but can be added if it is considered necessary

hotzevzl (approved):

PR #1217, opened 2022-07-28
add env sample

anamontiaga:


hotzevzl:
thanks @anamontiaga 

you could avoid adding the override to `.gitignore` if you are ok with using `env.default` for the stub `.env` file (or `env.sample`), as we do for the top-level `env.default`.

I'd suggest to also document the role of the variables which are not documented in `app/README.md` yet (could be just a link to the relevant upstream documentation for variables that have a well-defined meaning in NextJS).

At the very least some comments within the sample env file (again, see `env.default`), though if we document everything this makes it easier for whoever is going to deploy the whole platform.

For example, `VERCEL_URL` can likely be ignored for non-Vercel deployments (also, is this the correct one? I only see `NEXT_PUBLIC_VERCEL_URL` mentioned in the app code). Other vars are likely only needed during development but not for production - it would help to group them somehow and document this for deployment.

thanks!

hotzevzl (approved):
@anamontiaga thanks for adding inline docs about the various variables.

In general it may be helpful to set some sensible defaults that will work for a deployment on a local dev environment (that will also serve as a minimal example of what kind of values go into these vars) - again, I don't want to weigh in with backend conventions especially at this stage where most of what we want is for things to work reliably without much additional refinement, just a recommendation from my own point of view :) - nothing to do in this PR specifically.

Ok to merge from my point of view!

PR #1221, opened 2022-07-29
allow overlaps of (gis) features in feature shapefiles [MARXAN-1731]

hotzevzl:
https://github.com/Vizzuality/marxan-cloud/pull/new/chore/api/MARXAN-1731_allow-feature-shapefiles-with-overlapping-geometries

### DIFF
--- a/api/apps/api/src/modules/projects/projects.controller.ts
+++ b/api/apps/api/src/modules/projects/projects.controller.ts
@@ -725,7 +725,9 @@ export class ProjectsController {
   ): Promise<ShapefileUploadResponse> {
     await ensureShapefileHasRequiredFiles(shapefile);
 
-    const { data } = await this.shapefileService.transformToGeoJson(shapefile);
+    const { data } = await this.shapefileService.transformToGeoJson(shapefile, {
+      allowOverlaps: true,
+    });
 
     if (!isFeatureCollection(data)) {
       throw new BadRequestException(`Only FeatureCollection is supported.`);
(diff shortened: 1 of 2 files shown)


hotzevzl:
@aagm I have prepared the base functionality of this PR, letting tests run for feedback (ok, they passed while I was composing the rest of this message 😅) and I'm now checking what you recommended, that is, that we're not going to mess up areas in feature-related calculations once we allow overlaps between (gis) features.

The critical bits I can see so far are:

- [ ] https://github.com/Vizzuality/marxan-cloud/blob/5ab87f48087d96e1ce5de1f62818e115358ec529/api/libs/puvspr-calculations/src/puvspr-calculations.service.ts#L56-L86 (this should be ok)
- [ ] https://github.com/Vizzuality/marxan-cloud/blob/5ab87f48087d96e1ce5de1f62818e115358ec529/api/apps/api/src/modules/scenarios-features/copy/copy-query.service.ts#L145-L159 (this should be ok)
- [ ] https://github.com/Vizzuality/marxan-cloud/blob/5ab87f48087d96e1ce5de1f62818e115358ec529/api/apps/api/src/modules/scenarios-features/copy/copy-query.service.ts#L173-L177 (I am not sure about this)
- [ ] https://github.com/Vizzuality/marxan-cloud/blob/5ab87f48087d96e1ce5de1f62818e115358ec529/api/apps/api/src/modules/scenarios-features/split/split-query.service.ts#L60-L62 (I am not sure about this)
- [ ] https://github.com/Vizzuality/marxan-cloud/blob/5ab87f48087d96e1ce5de1f62818e115358ec529/api/apps/api/src/modules/scenarios-features/split/split-query.service.ts#L67-L79 (I am not sure about this)

We do use/reference `features_data` elsewhere but not the areas, so unless I missed something, the last three occurrences are the ones that may need a change, though except for casually throwing in a `st_union()` in hopefully the right spot of the PostGIS call chain, I am not familiar enough with how overlapping geometries are treated to be able to autonomously figure out a robust change 😏

PR #1229, opened 2022-08-04
avoid expensive spatial intersection over features that will never be relevant to the project [MARXAN-1744]

hotzevzl:
https://vizzuality.atlassian.net/browse/MARXAN-1744

### DIFF
--- a/api/apps/api/src/modules/geo-features/geo-features.service.ts
+++ b/api/apps/api/src/modules/geo-features/geo-features.service.ts
@@ -138,24 +138,14 @@ export class GeoFeaturesService extends AppBaseService<
       (info?.params?.projectId as string) ??
       (fetchSpecification?.filter?.projectId as string);
 
-    // find over api.features may not be best, as pagination does reflect
-    // actual page/pageSize vs actual items
-
-    // making search via intersection of whole PA vs all features
-    // may not be best for performance... but ain't we doing it anyway?
-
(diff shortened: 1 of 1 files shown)


hotzevzl:
@aagm as I was adding a small filter to explicitly avoid having split features (which following Ángel's recent work on split are now "first-class" features in `(apidb).features`) in lists of features (because of many reasons that would imply wide-reaching changes across the code base), I also noticed an old tech-debt note by Kamil, and this PR is an attempt to at least address the low-hanging fruits of his notes (which I have, accordingly, removed from the debt notes, while adding more thoughts for possible future improvements).

Basically until now we were first getting the list of _all_ the distinct `featureId`s of _any_ `(geodb)features_data` that intersects the project's bbox *and then* checking which ones were either for platform-wide features or project-specific features for the project being queried.

With this proposed change, we now first check which `featureId`s may be relevant at all (so, platform-wide or project-specific) and then we limit the bbox intersection check to only the `features_data` rows that belong to these features, which may keep query complexity kinda `O(prn)` (with `prn` being the number of `features_data` rows potentially relevant to the project, irrespective of their bbox) rather than `P(pln)` (with `pln` being the number of `features_data` rows _across the whole platform, irrespective of their public status or parent project).

So basically I expect this change not to make much of a difference right now when `pln ≈ prn`, but I hope it can slow down query time collapse when `pln ≪ prn`.

PR #1258, opened 2022-11-01
Enable patch version updates on k8s

tiagojsag:
Tried to apply this, ran into the following:

> │ Error: updating Cluster: (Managed Cluster Name "marxan-tnc" / Resource Group "marxan-rg"): containerservice.ManagedClustersClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="BadRequest" Message="To enabled Auto Upgrade, agentpools and cluster kubernetes versions must be at least (lowest supported minor version - 1) for rapid and stable channel or supported for patch and node-image channel."

Need to further look into it, didn't want to invest more time without ok from the client

### DIFF
--- a/infrastructure/base/main.tf
+++ b/infrastructure/base/main.tf
@@ -71,26 +71,28 @@ module "kubernetes" {
 }
 
 module "data_node_pool" {
-  source         = "./modules/node_pool"
-  name           = "data"
-  aks_cluster_id = module.kubernetes.cluster_id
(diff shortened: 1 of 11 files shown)


hotzevzl (approved):
agree with the proposed change and the recommendation to perform patch upgrades automatically and stable updates manually, plus monitoring to make sure that we don't get surprises in any case (as mach as we can shield from Murphy's law, of course)

PR #1309, opened 2023-01-30
add missing env var set from build argument [MARXAN-1822]

hotzevzl:
### Overview

`NEXT_PUBLIC_FEATURE_FLAGS` env var was not being set/baked at container image build time

### Designs

n/a

### Testing instructions

Building images for the client app via GH actions (or simply via the Docker cli used in the relevant GH Actions YAML config) should result in the `NEXT_PUBLIC_FEATURE_FLAGS` being available within the client app's container when running.

### Feature relevant tickets

https://vizzuality.atlassian.net/browse/MARXAN-1822

---

- [ ] Update CHANGELOG file


hotzevzl:
@tiagojsag asking for your review mainly so that you are aware of the planned fix, because you may be planning to merge #1308 which has higher priority: this PR/fix can wait if you don't want to add too many moving pieces to what we're deploying this week.

PR #1310, opened 2023-01-31
Update postgres to 14

tiagojsag:
Not sure how useful a review on this is at this stage, since it's being deployed to prod as I write this, but hey, maybe you notice a us-in-6-months problem


hotzevzl (approved):
no obvious us-in-6-months problems, to my TF-untrained eyes.

great job with the parallel pg + pg14 setup and with the fine tuning of permissions on pg roles 👏🏼

PR #1317, opened 2023-03-20
(spike) run e2e test suites in parallel [MARXAN-1825]

hotzevzl:
### Overview

Validating: run e2e tests in parallel by suite

### Designs

N/A

### Testing instructions

All e2e tests should run, but vroooooommmm.

### Feature relevant tickets

https://vizzuality.atlassian.net/browse/MARXAN-1825

---

- [ ] Update CHANGELOG file


hotzevzl:
@yulia-bel @KevSanchez @alexeh I think it's time for me to make the case for this PR to be considered for merging for real at last

As discussed, run times for api and geo e2e may vary a lot depending (or so I think) on how fast GH can throw available runners our way, but with the exception of a few outliers, I've seen most test runs complete all the jobs within times comparable with the "all in one" setup we've had so far - with the benefit however that if any job fails because of transient issues we can rerun just the failed jobs, which usually complete in a shorter time than the whole set (and avoid re-running tests that have just run successfully).

Example from recent runs:
- API: https://github.com/Vizzuality/marxan-cloud/actions/runs/4934728408
- Geoprocessing: https://github.com/Vizzuality/marxan-cloud/actions/runs/4934728413

I've rebased everything on Alex' refactor of tests. There are still things to polish, but for the moment I'm happy with the results.

Next step is, as I mentioned recently, for me to get the 5 container images used in tests (2x pg, redis, api and geo) built only once and pulled from each job. I had initially thought about "baking in" the migrations already run in the pg container images, but from what I see this should not give us massive time savings, and I think it would require quite some machinery to set up, which I would happily avoid if "just" building images once can already save a few minutes per job.

yulia-bel (approved):
LGTM! Hopefully this approach will help with memory issues with node18

PR #1322, opened 2023-03-29
update data seeding [MRXN23-7] [MRXN23-8] [MRXN23-9] [MRXN23-10] [MRXN23-12]

hotzevzl:
This PR Makes Seeds Great Again, or tries to do so.

- Azure storage account and container can now be configured via env vars, rather than using an hardcoded default
- For developer convenience, a usable `.env` file for a fresh new Marxan Cloud install can be generated via `make .env`
- A workaround for an intermittent issue upstream for the WDPA data has now been removed (leaving a comment in case WDPA goes into data chaos again)
- Environment variables needed for ETL pipelines have been added to `env.default` (so that `make .env` gives a fully working `.env`)
- Assorted minor cleanups

### Designs

N/A

### Testing instructions

Follow the README to set up a new Marxan Cloud instance, or - if testing on an existing instance - restore it to a clean slate.

Start the API services (`make start-api`) and once all is up, initiate a restore of dumps via `make restore-dumps`.

After a few minutes, the data import process should finish with exit code 0 and the following tables in the geoprocessing db should have plenty of data:

```
marxan-geo-api=# select count(*) from features_data ;
  count
---------
 1734381
(1 row)

marxan-geo-api=# select count(*) from wdpa ;
 count
--------
 220754
(1 row)
```

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


alexeh (approved):
LGTM

hotzevzl:
> One thing that I would suggest is to add the possibility that docker-compose commands changes based on the doker version, from docker-compose to docker compose.

good point - we could switch outright to using only `docker compose` plugin as the standalone is being deprecated upstream, but in the meanwhile:
- 4b481f65f57d2d112f12eaf559504fc4fea30419 (my version)
- 17f60f5f755dc610d3c3e4b29d3aeab7c0216ba0 (chatgpt's version, with my addition of moving it to a mk library)

PR #1335, opened 2023-05-18
feat(Features): Adds Delete Feature for Project endpoint [MRXN23-24]

KevSanchez:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/scenarios-features/scenario-features.service.ts
+++ b/api/apps/api/src/modules/scenarios-features/scenario-features.service.ts
@@ -79,6 +79,14 @@ export class ScenarioFeaturesService extends AppBaseService<
     ];
   }
 
+  async isFeaturePresentInAnyScenario(featureId: string): Promise<boolean> {
+    const count = await this.remoteScenarioFeatures.query(
+      `select count(*) from scenario_features_data where feature_class_id in (select id from features_data where feature_id = '${featureId}')`,
+    );
+
+    return count > 0;
+  }
+
   get serializerConfig(): JSONAPISerializerConfig<ScenarioFeaturesData> {
     return {
       attributes: [
--- a/api/apps/api/src/modules/scenarios-features/scenario-features.service.ts
+++ b/api/apps/api/src/modules/scenarios-features/scenario-features.service.ts
@@ -80,11 +80,12 @@ export class ScenarioFeaturesService extends AppBaseService<
   }
 
   async isFeaturePresentInAnyScenario(featureId: string): Promise<boolean> {
-    const count = await this.remoteScenarioFeatures.query(
-      `select count(*) from scenario_features_data where feature_class_id in (select id from features_data where feature_id = '${featureId}')`,
+    const result = await this.remoteScenarioFeatures.query(
+      `select count(*) from scenario_features_data where feature_class_id in (select id from features_data where feature_id = $1)`,
+      [featureId],
     );
 
-    return count > 0;
+    return result[0].count > 0;
   }
 
   get serializerConfig(): JSONAPISerializerConfig<ScenarioFeaturesData> {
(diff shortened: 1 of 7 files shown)


api/apps/api/src/modules/scenarios-features/scenario-features.service.ts
```diff
@@ -79,6 +79,14 @@ export class ScenarioFeaturesService extends AppBaseService<
     ];
   }
 
+  async isFeaturePresentInAnyScenario(featureId: string): Promise<boolean> {
+    const count = await this.remoteScenarioFeatures.query(
+      `select count(*) from scenario_features_data where feature_class_id in (select id from features_data where feature_id = '${featureId}')`,
+    );
```
hotzevzl, line 85: a couple of things:
- I strongly advise against using template strings in queries - even if we pay close attention, it's too easy to accidentally overlook something that includes, somewhere deep up the chain of calls, some user input
```suggestion
    const count = await this.remoteScenarioFeatures.query(
      `select count(*) from scenario_features_data where feature_class_id in (select id from features_data where feature_id = $1)`,
      [ featureId ]
    );
```

api/apps/api/src/modules/scenarios-features/scenario-features.service.ts
```diff
@@ -79,6 +79,14 @@ export class ScenarioFeaturesService extends AppBaseService<
...
+
+    return count > 0;
```
hotzevzl, line 87: secondly, although in theory the query above looks legit (_BUT_ I need some time to think through the implications of split features on this query, which I can only do tomorrow morning unfortunately), in practice you end up boycotting the query result in this check here, because `count` is not what you think it will be, and `count > 0` will always be `true`

If you inspect this, the return of `query()` will be something like this:

```
[ { count: '3' } ]
```

As a corollary of this, I think you should also review the test that checks that deletion of in-use features is prevented, as this bug should be caught there, and if the test passes then... I think the test it's not correct.

Try creating a project/scenario, uploading some features (fun time with bots?!) and link them to a scenario, and then delete these via the new API endpoint.

hotzevzl (changes requested):

hotzevzl:
@KevSanchez thank you! A couple of cosmetic notes, and two substantive (three if you count my doubt about the test I mention inline) notes about the check for features that are in use.

api/apps/api/test/project/delete-project-feature.e2e-spec.ts
```diff
@@ -0,0 +1,267 @@
...
+      featureId,
+    );
+    await fixtures.ThenDeleteWasForbidden(
+      result,
+      featureId,
+      `Feature with id ${featureId}, for project with id ${viewOnlyProjectId}, cannot be deleted`,
+    );
+  });
+
+  test('should delete the name of a feature, when permitted', async () => {
```
hotzevzl, line 117: leftover from test copied over from the previous suite, probably?
```suggestion
  test('should delete a feature, when permitted', async () => {
```

api/apps/api/test/project/delete-project-feature.e2e-spec.ts
```diff
@@ -0,0 +1,267 @@
...
+    const result = await fixtures.WhenDeletingFeatureForProject(
+      viewOnlyProjectId,
+      featureId,
+    );
+    await fixtures.ThenDeleteWasForbidden(
+      result,
+      featureId,
+      `Feature with id ${featureId}, for project with id ${viewOnlyProjectId}, cannot be deleted`,
+    );
+  });
```
hotzevzl, line 98: ok, still a quick glance but I think the reason why tests don't catch the issue that I flagged in my review about linking of features to scenarios not being correctly checked is that here we're testing again that a feature cannot be deleted if the user only has read-only access to a project (the following test, basically) - at least that's what the naming of fixtures used here suggests - I haven't followed the full flow

hotzevzl (approved):

PR #1337, opened 2023-05-24
update scenario planning units inclusion processor+ar suggestions [MRXN23-78]

hotzevzl:
My suggestions would not fit in the margin of the original PR, so I created this branch instead with just a couple of commits with my suggestions.

### DIFF
--- a/api/apps/geoprocessing/src/modules/scenario-planning-units-inclusion/scenario-planning-units-inclusion-processor.ts
+++ b/api/apps/geoprocessing/src/modules/scenario-planning-units-inclusion/scenario-planning-units-inclusion-processor.ts
@@ -1,6 +1,6 @@
 import { Injectable } from '@nestjs/common';
 import { InjectDataSource, InjectRepository } from '@nestjs/typeorm';
-import { DataSource, In, Repository } from 'typeorm';
+import { DataSource, EntityManager, In, Repository } from 'typeorm';
 import { MultiPolygon, Polygon } from 'geojson';
 import { flatMap, intersection } from 'lodash';
 import { Job } from 'bullmq';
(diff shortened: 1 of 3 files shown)


api/apps/geoprocessing/src/modules/scenario-planning-units-inclusion/scenario-planning-units-inclusion-processor.ts
```diff
@@ -124,38 +122,19 @@ export class ScenarioPlanningUnitsInclusionProcessor
     const uniquePuIdsToExclude = new Set(puIdsToExclude);
     const uniquePuIdsToMakeAvailable = new Set(puIdsToMakeAvailable);
 
-    const doInclusionAndExclusionIntersect =
-      intersection([...uniquePuIdsToInclude], [...uniquePuIdsToExclude])
+    const doClaimsIntersect =
+      intersection([...uniquePuIdsToInclude], [...uniquePuIdsToExclude], [...uniquePuIdsToMakeAvailable])
```
yulia-bel, line 126: There seems to be an issue here, since when we are passing 3 arrays to intersection, it will return the part that exists in all 3 of them. So, for example, if include and exclude arrays have intersecting pus, but make available doesn't have same part - it will return empty array and no error will be thrown..
I suggest keeping same 3 intersection checks for each pair and unify error message like this:
``` if (
      doInclusionAndExclusionIntersect ||
      doMakeAvailableAndExclusionIntersect ||
      doMakeAvailableAndInclusionIntersect
    ) {
      throw new Error(
        'Contrasting claims to include, exclude or make available some of the planning units have been made: please check your selections.',
      );
    }```

PR #1338, opened 2023-05-26
docs(Features): High Level design for Feature tagging

KevSanchez:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/docs/features/tagging/brief.md
+++ b/docs/features/tagging/brief.md
@@ -9,7 +9,7 @@ useful for the users, so that they may be able to identify/group features and do
 
 ## Requirements
 
-Whenever an user uploads a feature, they must be able to add a tag (although it shouldn't be a requirement when creating
+Whenever a user uploads a feature, they must be able to add a tag (although it shouldn't be a requirement when creating
 the feature). There will be one tag per feature. The user must also be able to change the tag of a feature at later
 point when editing the feature. There's an exception to this: System wide features (features with no project) are
 considered to be non-taggable, however features obtained by splitting from System wide features will become effectively
@@ -29,13 +29,13 @@ Bulk operations are not required to be implemented on the API side.
 
 ## Use cases
 
-This are the main use cases that are to be expected:
+These are the main use cases that are to be expected:
 
 - tagging a feature (when uploading, or as an isolated operation)
-- update/change a tag in a feature
-- remove the tag of a feature
-- get all features for a given project and tag
-- remove a tag in a project (thus removing the tags from the affected features)
+- updating/changing a tag in a feature
+- removing the tag of a feature
+- getting all features for a given project and tag
+- removing a tag in a project (thus removing the tags from the affected features)
 - renaming a tag in a project
-- query tags for a project with partial match on the tag name (by default, tags containing `lepr`, other matching styles
-  can be considered in the implementation)
+- querying tags for a project with partial match on the tag name (by default, tags containing a given token such as, for
+  example, `lepr`; other matching styles can be considered in the implementation)
--- a/docs/features/tagging/brief.md
+++ b/docs/features/tagging/brief.md
@@ -1,0 +1,41 @@
+# Feature Tagging (Brief)
+
+This document provides a brief overview of the feature tagging system in Marxan
+
+## Aim
+
+The main objective is to let the user tag features, be it when they upload them or at a later point in time. This can be
+useful for the users, so that they may be able to identify/group features and do any required operations based on that.
+
+## Requirements
+
+Whenever an user uploads a feature, they must be able to add a tag (although it shouldn't be a requirement when creating
+the feature). There will be one tag per feature. The user must also be able to change the tag of a feature at later
+point when editing the feature. There's an exception to this: System wide features (features with no project) are
+considered to be non-taggable, however features obtained by splitting from System wide features will become effectively
+regular features that can be tagged.
+
+Each Project will have its own isolated set of tags; for example, tag `leprechaun` on Project A would be completely
+independent from the same tag on project B. Related to his, all operations related to tags (such as renaming, removing),
+will be applied on a given project, and its related tags; when removing tag `leprechaun` from project A, the same tag
+from project B will remain unaffected. Tags can only be modified in the inventory panel.
+
+Tags should be restricted to a maximum number of characters (this amount can be configurable), and should not contain
+special characters such as new lines. Misspellings and variations of the tag should be handled by the users themselves.
+However, even though no capitalization scheme is enforced (the user can use tag in whichever capitalization they want),
+**capitalization uniqueness is**; so if there's a tag `Cat`, tagging a feature with `cat` or `CAT`, the system will
+internally tag with the already existent capitalization of `Cat`.
+Bulk operations are not required to be implemented on the API side.
+
+## Use cases
+
+This are the main use cases that are to be expected:
+
+- tagging a feature (when uploading, or as an isolated operation)
+- update/change a tag in a feature
+- remove the tag of a feature
+- get all features for a given project and tag
+- remove a tag in a project (thus removing the tags from the affected features)
+- renaming a tag in a project
+- query tags for a project with partial match on the tag name (by default, tags containing `lepr`, other matching styles
+  can be considered in the implementation)
--- a/docs/features/tagging/high-level-design.md
+++ b/docs/features/tagging/high-level-design.md
@@ -1,0 +1,151 @@
+# Feature Tagging (High-Level Design)
+
+The selected solution for this feature is to implement a new table `feature_tags` on the API DB that will hold, for each
+row, a tag for a given combination of feature and project id like this:
+
+`| rowId (VARCHAR) | projectId (VARCHAR) | featureId (VARCHAR) | tag (VARCHAR) |`
+
+where:
+
+- **rowId** is a generated UUID PK
+- **projectId** is a foreign key referencing the PK of the Projects table with ON DELETE CASCADE set. Cannot be NULL.
+- **featureId** is a foreign key referencing the PK of the Features table with ON DELETE CASCADE set. Cannot be NULL.
+- **tag** contains the tag string itself, **case sensitive**, and cannot be NULL or empty. If a feature doesn't have
+  any tags for a given project, then no rows associated to it should be present.
+- There will be an **UNIQUE INDEX** on the combination of **`projectId` and `featureId`**. This will limit the number of
+  tags to **1 tag per feature**.
+- There will be an **UNIQUE INDEX** on the combination of **`projectId` and `LOWER(tag)`** in order to enforce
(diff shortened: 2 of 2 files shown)


docs/features/tagging/high-level-design.md
```diff
@@ -0,0 +1,151 @@
...
+  tags to **1 tag per feature**.
+- There will be an **UNIQUE INDEX** on the combination of **`projectId` and `LOWER(tag)`** in order to enforce
+  **capitalization uniqueness** of the tag names
+
+On the DB side, the size the `tag` is unbounded, but is bounded at the service level via validation by a configurable
+amount of
+max characters.
+
+Once the last feature with a given tag is removed or untagged, that tag won't be available to be selected as a tag
+anymore. Tags will be considered removed when there are no rows for a given `projectId` and `featureId` combination
```
hotzevzl, line 25: maybe for extra clarity?

```suggestion
anymore. Tags will be considered removed from a project when there are no rows for a given `projectId` and `featureId` combination
```

docs/features/tagging/high-level-design.md
```diff
@@ -0,0 +1,151 @@
...
+with a given tag name.
+
+**Capitalization uniqueness should be enforced at the service level**. When tagging/renaming a tag on a feature, a case
+insensitive check must be made first to see if there's an already existing tag, regardless of capitalization, and use
+that tag instead of the one provided by the user. For example, if an user tags a Feature with `Mogwai`, and other
```
hotzevzl, line 30:
```suggestion
that tag instead of the one provided by the user. For example, if a user tags a Feature with `Mogwai`, and other
```

docs/features/tagging/high-level-design.md
```diff
@@ -0,0 +1,151 @@
...
+features for the same project are already tagged with `mogwai` instead, `mogwai` should be used as the new tag. However
+this capitalization uniqueness **should not be applied tag renames for a given project** as the user might want to
```
hotzevzl, line 32:
```suggestion
this capitalization uniqueness **should not be applied when renaming tags for a given project** as the user might want to
```

docs/features/tagging/high-level-design.md
```diff
@@ -0,0 +1,151 @@
...
+Some other solutions considered were JSONB Arrays or an Array of text values, as an extra column on the `features`
+table, containing all the tags in plain text form. Although these are better performant solutions in some use cases (tag
+cloud generation, tag combination querying...), given the scope, budget and data volume, a more classic solution based
+on an external table with JOINs is enough for this case.
+
+More info on different tagging implementation approaches [here](http://www.databasesoup.com/2015/01/tag-all-things.html)
+
+These are some of the main criteria used for deciding on the external table solution:
+
+- **All tag managing operations are done per project**: this limits quite a bit the number rows that queries must
```
hotzevzl, line 44:
```suggestion
- **All tag managing operations are done per project**: this limits quite a bit the number of rows that queries must
```

docs/features/tagging/high-level-design.md
```diff
@@ -0,0 +1,151 @@
...
+- **The original `features` table remains the same**
+- **Budget**
+
+## Workflow
+
+There are two distinct parts to the tagging system.
+
+### Feature tagging operations
+
+These endpoints deal with operating tags for an individual feature.
```
hotzevzl, line 61:
```suggestion
These endpoints deal with handling tags for an individual feature.
```

docs/features/tagging/high-level-design.md
```diff
@@ -0,0 +1,151 @@
...
+
+- `PATCH /api/v1/geo-features/:featureId/tags`
+
+```typescript
+Payload
+{
+	projectId: string
```
hotzevzl, line 68: we can decide at implementation time, but I think we can safely leave out `projectId` here because we would need to look it up in any case, as "parent" project of the feature the user wants to edit, as we need the "real" `projectId` anyway in order to process ACLs (we can't rely on a user-supplied `projectId` for this).

docs/features/tagging/brief.md
```diff
@@ -0,0 +1,41 @@
...
+
+Each Project will have its own isolated set of tags; for example, tag `leprechaun` on Project A would be completely
+independent from the same tag on project B. Related to his, all operations related to tags (such as renaming, removing),
+will be applied on a given project, and its related tags; when removing tag `leprechaun` from project A, the same tag
+from project B will remain unaffected. Tags can only be modified in the inventory panel.
+
+Tags should be restricted to a maximum number of characters (this amount can be configurable), and should not contain
+special characters such as new lines. Misspellings and variations of the tag should be handled by the users themselves.
+However, even though no capitalization scheme is enforced (the user can use tag in whichever capitalization they want),
+**capitalization uniqueness is**; so if there's a tag `Cat`, tagging a feature with `cat` or `CAT`, the system will
```
hotzevzl, line 26: tiny and opinionated detail (feel free to ignore) - maybe we could label this "consistent capitalization" instead?

docs/features/tagging/brief.md
```diff
@@ -0,0 +1,41 @@
...
+This document provides a brief overview of the feature tagging system in Marxan
+
+## Aim
+
+The main objective is to let the user tag features, be it when they upload them or at a later point in time. This can be
+useful for the users, so that they may be able to identify/group features and do any required operations based on that.
+
+## Requirements
+
+Whenever an user uploads a feature, they must be able to add a tag (although it shouldn't be a requirement when creating
```
hotzevzl, line 12:
```suggestion
Whenever a user uploads a feature, they must be able to add a tag (although it shouldn't be a requirement when creating
```

docs/features/tagging/high-level-design.md
```diff
@@ -0,0 +1,151 @@
...
+Removes the tag for the given `featureId`. This means removing the corresponding row on the `feature_tags` table.
+
+## Tag Managing
+
+These endpoints deal with tag managing operations on a per Project level.
+
+- `GET /api/v1/projects/:projectId/tags?tag=someName`
+
+For the given `projectId`, returns a list of all the DISTINCT tags from the Project's Features that have a tag that
+partially matches with `tag`. By default, the matching is `containing`, but other options could be considered.
```
hotzevzl, line 104:
```suggestion
For the given `projectId`, returns a list of all the DISTINCT tags from the Project's Features that
partially matche the `tag` query parameter. By default, the matching is `containing`, but other options could be considered.
```

hotzevzl: I think we want the `tag` query param to be optional too, right? so that users can use `GET /api/v1/projects/:projectId/tags` to get the list of all the tags currently available within the given project.

docs/features/tagging/high-level-design.md
```diff
@@ -0,0 +1,151 @@
...
+No pagination required, since the possible number of tags per Project will be in the low tens.
+
+- `DELETE /api/v1/projects/:projectId/tags/`
+
+```typescript
+Payload
+{
+	tag: string;
+}
+```
```
hotzevzl, line 117: A payload on a DELETE may be risky: although it should work _in practice_, we don't fully control the whole HTTP stack in production/staging and as such... I would discourage this.

I'd either use the uuid of the tag to delete (which is a bit of an extra pain to get and use, but allows to limit the URL to ASCII) or the name of the tag (maybe urlencoded to avoid having to trust that all layers of the HTTP stack correctly handle unicode in URLs?)

hotzevzl: For context, see https://stackoverflow.com/a/299696. ChatGPT also invites to be cautious 😬

hotzevzl (approved):
Thanks! Basically all good to me besides some spelling or (opinionated: feel free to ignore) naming suggestions.

The only core concern is with the reliance on a body for the DELETE endpoint for tags, which... should work, but I'd recommend to avoid. We do use this elsewhere at least in another project, where the use predates the latest RFC's clarifications/updates - I don't remember if we do use it in this project - in which case, we should consider removing it, but that is a fully distinct story 😄

docs/features/tagging/high-level-design.md
KevSanchez: Yeah, thinking back on it, I totally agree.
However, I think using the uuid of the tag wouldn't be correct, because that identifies a single instance of a given tag for that project. If you have multiple features with the same tag, using the uuid it's going to delete the tag for one feature.
Using the tag text as a query parameter, urlencoded as  you say, would most likely be the way to go.

docs/features/tagging/high-level-design.md
KevSanchez: You mean doing a quick load of the feature metadata to retrieve its projectId right? Yeah, that could work too, and shouldn't be much of an overhead. We'll see when we get to it.

docs/features/tagging/high-level-design.md
hotzevzl: of course, you are very right. perfect as in your latest commit 👍🏼

hotzevzl (approved):

docs/features/tagging/high-level-design.md
```diff
@@ -0,0 +1,147 @@
...
+
+- `GET /api/v1/projects/:projectId/tags?tag=someName`
+
+For the given `projectId`, returns a list of all the DISTINCT tags from the Project's Features that partially matches
+the `tag` query parameter. By default, the matching is `containing`, but other options could be considered.
+
+The `tag` query parameter may be optional, so that if it is empty, all the DISTINCT tags for the given `projectId` are
+returned.
+
+Ordered alphabetically by default.
```
andresgnlez, line 108: Will be possible other sorting values? Like `most-recent` or similar. Per design, the default view when the user is about to type a tag is displaying the most recent tags introduced, so it would be ideal having this use case as well.

image

docs/features/tagging/high-level-design.md
```diff
@@ -0,0 +1,170 @@
+# Feature Tagging (High-Level Design)
+
+The selected solution for this feature is to implement a new table `feature_tags` on the API DB that will hold, for each
+row, a tag for a given combination of feature and project id like this:
+
+`| rowId (uuid) | projectId (uuid) | featureId (uuid) | tag (VARCHAR) | last_applied (timestamp) |`
```
hotzevzl, line 6: from the rest of the text, I suppose you meant

```suggestion
`| rowId (uuid) | projectId (uuid) | featureId (uuid) | tag (VARCHAR) | last_updated (timestamp) |`
```

It's just a tiny implementation detail, in any case (and maybe at implementation time I'd suggest to stick to the same name we have for similar columns already, that is `last_modified_at timestamp without timezone default now()`)

hotzevzl (approved):

docs/features/tagging/high-level-design.md
KevSanchez: Good catch, I'll change it later.

PR #1340, opened 2023-06-06
Add new API endpoint for SF comparison maps

yulia-bel:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


api/apps/api/src/modules/projects/projects.service.ts
```diff
@@ -759,6 +774,55 @@ export class ProjectsService {
...
   }
 
+  async getScenarioFrequencyComparisonMap(
+    scenarioIdA: string,
+    scenarioIdB: string,
+    userId: string,
+    configForWebshot: WebshotSFComparisonMapPdfConfig,
+  ): Promise<
+    Either<
+      | typeof forbiddenError
```
hotzevzl, line 784: sorry, I had missed this. We should also check that the current user (which we capture in the controller endpoint via `RequestWithAuthenticatedUser`) has a role on the parent project of the scenario, or throw a `forbiddenError`.

This is clearly my fault - I haven't clearly specified this, and on the other hand we should do this kind of automatically (alongside a few other things) when creating new functionality, so I've created a task for myself to document these things, for them to be available as a checklist for all of us: https://vizzuality.atlassian.net/browse/MRXN23-202

hotzevzl: sorry, only getting to this now 😓 - I think it's perfectly fine to use `canViewProject()`

I don't understand your mention of project owner, though: any user with any role on a project should be able to download the comparison maps.

yulia-bel: yes, sorry - misread the functionality of that :grimacing:
ok, thanks, will merge it now

PR #1341, opened 2023-06-08
feat(Feature Tags): Adds PATCH endpoint to update tags on features

KevSanchez:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


KevSanchez:
Would it be necessary to check whether the feature on the PATCH call is a custom feature or not, and prevent tagging if non custom?
Custom features are always associated to a project, and the DTO cannot have an empty projectId, so that situation should never be possible, right?

KevSanchez:
One last question, we had agreed for now not to modify the `last_modified_at` field of the equivalent tag rows on the same project, right?

api/apps/api/src/modules/geo-feature-tags/dto/tag-info.dto.ts
```diff
@@ -0,0 +1,16 @@
...
+import { IsValidTagNameValidator } from '@marxan-api/modules/geo-feature-tags/validators/is-valid-tag-name.custom.validator';
+
+const tagMaxlength = AppConfig.get<number>('marxan.maxTagLength');
+
+export class TagInfoDTO {
+  @ApiProperty()
+  @IsNotEmpty()
+  @Validate(IsValidTagNameValidator)
+  @MaxLength(tagMaxlength, {
+    message: `tagName length should not be longer than ${tagMaxlength} characters`,
```
hotzevzl, line 13: very tiny, mostly style - in case frontend would like to pass on an error message from the API, I suggest to make these as user-meaningful as possible (for example, "A tag should not be longer than ... characters").

I see no harm in trying to make errors/warnings as simple and meaningful as possible for end users even if they end up not being displayed. When triaging issues, the error details in responses should give us enough information to get to the implementation details without having to surface these through error messages.

hotzevzl: And yes, I know we don't have universally good examples across this application on this front - something for me to add to documentation on project conventions, I guess 😬

api/apps/api/src/modules/geo-feature-tags/geo-feature-tags.service.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export class GeoFeatureTagsService {
+  constructor(
+    @InjectRepository(GeoFeatureTag)
+    private readonly geoFeatureTagsRepo: Repository<GeoFeatureTag>,
+    @InjectRepository(GeoFeature)
+    private readonly geoFeaturesRepo: Repository<GeoFeature>,
+  ) {}
+
+  async updateTagForFeature(
```
hotzevzl, line 20: tiny style suggestion: we do often provide functions that allow to set or update something, where specifically we don't care if this something was already set or not - in these cases, I recommend to use a `setOrUpdateSomething` kind of name to make this immediately clear when reading the source code.

api/apps/api/src/modules/geo-feature-tags/geo-feature-tags.service.ts
```diff
@@ -0,0 +1,53 @@
...
+    projectId: string,
+    featureId: string,
+    tag: string,
+  ): Promise<
```
hotzevzl, line 24: apologies about not having specified this explicitly, but we should always pass in the user (from the controller's `RequestWithAuthenticatedUser`) _and_ do all the usual permissions checks - you can mostly copy over the code you've already created in `GeoFeaturesService.updateFeatureForProject()`, for example.

api/apps/api/src/modules/geo-feature-tags/geo-feature-tags.service.ts
```diff
@@ -0,0 +1,53 @@
...
+    const geoFeature = await this.geoFeaturesRepo.findOne({
+      where: { id: featureId },
+    });
+
+    if (!geoFeature) {
+      return left(featureNotFound);
+    }
+    if (geoFeature.projectId !== projectId) {
+      return left(featureProjectNotRelated);
+    }
```
hotzevzl, line 36: here we're using the user-supplied `projectId` without even checking that this is the id of an actual project - but instead of validating this indirectly by comparing the feature's actual `projectId` with the `projectId` supplied by the user, I would suggest to simplify like this

```suggestion
    const geoFeature = await this.geoFeaturesRepo.findOne({
      where: { id: featureId, projectId },
    });

    if (!geoFeature) {
      return left(featureNotFound);
    }
```

and forget about the `projectId` - what we need to know here is whether the given feature exists within the given project, and I think the suggested check should suffice.

api/apps/api/src/modules/geo-features/geo-features.controller.ts
```diff
@@ -52,6 +71,38 @@ export class GeoFeaturesController {
...
+
+    if (isLeft(result)) {
+      switch (result.left) {
+        case featureNotFound:
+          throw new NotFoundException(`Feature with id ${featureId} not found`);
+        case featureProjectNotRelated:
+          throw new NotFoundException(
+            `Feature with id ${featureId} and Project with id ${dto.projectId} are not related`,
+          );
+      }
```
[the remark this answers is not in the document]
hotzevzl: this would include any new permissions-related errors, for which we may already have suitable cases in `mapAclDomainToHttpError()`

hotzevzl:
> Would it be necessary to check whether the feature on the PATCH call is a custom feature or not, and prevent tagging if non custom? Custom features are always associated to a project, and the DTO cannot have an empty projectId, so that situation should never be possible, right?

See my inline notes on this.

> One last question, we had agreed for now not to modify the `last_modified_at` field of the equivalent tag rows on the same project, right?

yep, I think we can avoid this for the time being. My understanding is that we're not implementing MRU/LRU sorting of tags, so this won't really be needed.

api/apps/api/test/geo-features/geo-feature-tags.e2e-spec.ts
```diff
@@ -0,0 +1,354 @@
...
+import { Organization } from '@marxan-api/modules/organizations/organization.api.entity';
+import { FixtureType } from '@marxan/utils/tests/fixture-type';
+import { AppConfig } from '@marxan-api/utils/config.utils';
+import { v4 } from 'uuid';
+
+let fixtures: FixtureType<typeof getFixtures>;
+
+const tagMaxlength = AppConfig.get<number>('marxan.maxTagLength');
+
+describe('GeoFeatureTag PATCH (e2e)', () => {
```
hotzevzl, line 22: I forgot - as this is a new test folder within `/api/apps/api/test`, we also need to add this to the job matrix in `.github/workflows/tests-api-e2e.yml`, otherwise these tests will never be executed.

I think I should figure out a pre-push hook to check that we don't forget this when creating new folders within the root test folder for either api or geoprocessing.

what I mean, in practice is something like this:

```
diff --git a/.github/workflows/tests-api-e2e.yml b/.github/workflows/tests-api-e2e.yml
index 285aa8a51..fb9451765 100644
--- a/.github/workflows/tests-api-e2e.yml
+++ b/.github/workflows/tests-api-e2e.yml
@@ -25,6 +25,7 @@ jobs:
           - 'calculate-pu-protection-level'
           - 'cloning'
           - 'files-repository'
+          - 'geo-features'
           - 'implicit-permissions'
           - 'integration'
           - 'legacy-project-import'
```

api/apps/api/src/modules/geo-feature-tags/dto/update-geo-feature-tag.dto.ts
```diff
@@ -0,0 +1,16 @@
...
+export class UpdateGeoFeatureTagDTO {
+  @ApiProperty()
+  @IsUUID()
+  projectId!: string;
+
+  @ApiProperty()
+  @IsNotEmpty()
+  @ValidateNested({ each: true })
+  @Type(() => TagInfoDTO)
+  tagInfo!: TagInfoDTO;
```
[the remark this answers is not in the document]
KevSanchez: This I was wasn't 100% convinced on, but it come from one comment in our early conversations about the high level design, where we could reuse the DTO, so that it wouldn't be needed to copy/paste all the validations in any new DTOs that require tag information. For example the message for the length validation would get duplicated but I guess we can turn it into a constant/function?

PR #1345, opened 2023-06-14
feat(Project Tags): Adds DELETE endpoint to remove a tag from a project

KevSanchez:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/geo-feature-tags/geo-feature-tags.service.ts
+++ b/api/apps/api/src/modules/geo-feature-tags/geo-feature-tags.service.ts
@@ -1,10 +1,15 @@
 import { InjectDataSource, InjectRepository } from '@nestjs/typeorm';
 import { GeoFeatureTag } from '@marxan-api/modules/geo-feature-tags/geo-feature-tag.api.entity';
-import { DataSource, Repository } from 'typeorm';
+import { DataSource, ILike, Repository } from 'typeorm';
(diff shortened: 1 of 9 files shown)


api/apps/api/src/modules/geo-feature-tags/geo-feature-tags.service.ts
```diff
@@ -109,4 +116,30 @@ export class GeoFeatureTagsService {
 
     return right(true);
   }
+
+  async deleteTagForProject(
+    userId: string,
+    projectId: string,
+    tag: string,
+  ): Promise<Either<typeof projectNotEditable | typeof projectNotFound, any>> {
```
hotzevzl, line 124: tiny: I think you can avoid this `any` in the right side of the `Either`

hotzevzl (approved):
Nice touch to only test the sad path (computer says no) in the ACL-related tests. I'm going to add this recommendation in the BE dev workflow docs 🚀 (basically all we should be checking there is that if no role, no worky, I think - the happy (ACL-wise) path should be implicit in the functional e2e tests)

api/apps/api/src/modules/geo-feature-tags/geo-feature-tags.service.ts
KevSanchez: you mean as in using boolean/true/void/whatever? if so, yeah, I can modify that on another commit

PR #1346, opened 2023-06-16
[api] Clear PU status by category

yulia-bel:
### Overview

Adding possibility for the user to clear selected kind of previously set statuses: Locked-in, Locked-out or Made available.

Process is the following:
 - Endpoint receives scenario id and kind of status to be cleared
 - Service finds all the PU ids that were set by user for each status kind 
 - Removes the kind selected to be cleared from the claims to be sent to status change processor
 - Processor resets all the statuses and ads only received claims. Since no claims for the selected to be cleared statuses will be received - PUs that had this status will be set to:
  available with setbyUser=false OR 
  locked-in with setByUser=false with protectedByDefault=true in case of PUs belonging to a protected area

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units.service.ts
+++ b/api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units.service.ts
@@ -32,7 +32,7 @@ export class ScenarioPlanningUnitsService {
     scenarioId: string,
     lockStatus: LockStatus,
   ): Promise<ScenariosPlanningUnitGeoEntity[]> {
-    return this.puRepo.find({
+    return await this.puRepo.find({
       where: {
         scenarioId,
         lockStatus,
--- a/api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units.service.ts
+++ b/api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units.service.ts
@@ -28,6 +28,18 @@ export class ScenarioPlanningUnitsService {
     });
   }
 
+  async getByStatus(
+    scenarioId: string,
+    lockStatus: LockStatus,
+  ): Promise<ScenariosPlanningUnitGeoEntity[]> {
+    return this.puRepo.find({
+      where: {
+        scenarioId,
+        lockStatus,
+      },
+    });
+  }
+
   async resetLockStatus(scenarioId: string): Promise<void> {
     return await this.entityManager.transaction(async (manager) => {
       await manager.update(
--- a/api/apps/api/src/modules/scenarios/scenarios.controller.ts
+++ b/api/apps/api/src/modules/scenarios/scenarios.controller.ts
@@ -108,6 +108,7 @@ import { forbiddenError } from '../access-control';
 import { scenarioNotFound } from '../blm/values/blm-repos';
 import { RequestScenarioCloneResponseDto } from './dto/scenario-clone.dto';
 import { ensureShapefileHasRequiredFiles } from '@marxan-api/utils/file-uploads.utils';
+import { LockStatus } from '@marxan/scenarios-planning-unit';
 import { WebshotPdfReportConfig } from '@marxan/webshot/webshot.dto';
 
 const basePath = `${apiGlobalPrefixes.v1}/scenarios`;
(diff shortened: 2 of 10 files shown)


api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units.service.ts
```diff
@@ -28,6 +28,18 @@ export class ScenarioPlanningUnitsService {
     });
   }
 
+  async getByStatus(
+    scenarioId: string,
+    lockStatus: LockStatus,
+  ): Promise<ScenariosPlanningUnitGeoEntity[]> {
+    return this.puRepo.find({
```
hotzevzl, line 35: I usually recommend to use `return await` _always_ even when not strictly needed, mostly for the reasons put forward in this comment: https://stackoverflow.com/questions/38708550/difference-between-return-await-promise-and-return-promise#comment128897650_70198397

In practice, linters should take care of guiding us and letting us not have to keep all these details in mind. As long as the reasons for using either arrangement are clear, I have no strong opinions though.

api/apps/api/src/modules/scenarios/scenarios.service.ts
```diff
@@ -447,6 +482,50 @@ export class ScenariosService {
       },
     };
   }
+
+  private async mapCurrentPuStatusesAndClearRequested(
+    scenarioId: string,
+    kindToClear: string,
```
hotzevzl, line 488: I think you could type this more strictly as
```suggestion
    kindToClear: LockStatus,
```

api/apps/api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -524,6 +525,25 @@ export class ScenariosController {
     return AsyncJobDto.forScenario().asJsonApiMetadata();
   }
 
+  @ApiTags(asyncJobTag)
+  @ApiOkResponse()
+  @Delete(':id/planning-units/status/:kind')
+  async clearPlanningUnitsStatus(
+    @Param('id', ParseUUIDPipe) id: string,
+    @Param('kind') kind: string,
```
hotzevzl, line 533: you could (should!) enforce that `kind` is one of the vetted `LockStatus` enum values. If it's not any of these, we can fail fast and return a 400 directly here (ideally via class-validator)

hotzevzl (changes requested):
thanks @yulia-bel! Mostly all good, and well done for already doing things "the Marxan way" (or one of the several Marxan ways 😬 , but whatever).

only real recommendation is to enforce already at controller level that the kind to clear is only one of the ones we can expect and can process (and then narrow the typing of `clearLockStatus()` accordingly, too.

Plus a bunch of naming suggestions/typo checks, feel free to take your decision (on naming suggestions, not on typos 😛).

yulia-bel:
@hotzevzl thanks for the comments and suggestions - all fixed, also added one more check to geoprocessing test to cover 2 clears (include and exclude)

api/apps/api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -524,6 +525,25 @@ export class ScenariosController {
     return AsyncJobDto.forScenario().asJsonApiMetadata();
   }
 
+  @ApiTags(asyncJobTag)
+  @ApiOkResponse()
+  @Delete(':id/planning-units/status/:kind')
+  async clearPlanningUnitsStatus(
+    @Param('id', ParseUUIDPipe) id: string,
+    @Param('kind') kind: LockStatus,
```
hotzevzl, line 533: mind you, this will still not prevent users from requesting `DELETE /api/v1/scenarios/:scenarioId/planning-units/status/potato`: this will likely result in a no-op behind the scenes (the `scenario.planningUnitsInclusion` flow finishes without errors, and I think it just does nothing). Not a big deal since there is no error and it's unlikely to result in an exploit (hard to know without more careful analysis), and since almost all the use of this will be via the client app, which should always send meaningful values.

yulia-bel: That's true, missed the fact that it's not the body with dto.. Thanks for pointing out - I added a small fix of validation and a test for it

hotzevzl (approved):

PR #1359, opened 2023-07-03
feat(Scenarios): New Id for Scenarios wihtin Projects

KevSanchez:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


hotzevzl (commented):
thanks @KevSanchez, I think this is all that is needed.

One thing I would ask you to check is whether we'd need to plugin in your new creation function in the legacy projects import flow - there we create a single project with a single scenario in it, but then users could add new scenarios. We should make sure the only scenario imported is correctly assigned its numeric id.

Also, in scenario piece exporters and importers, we should handle this new column.

And ideally we should assign numeric ids to scenarios retroactively via a migration for scenarios already in a "live" instance (staging and production).

api/apps/api/src/modules/scenarios/scenario.api.entity.ts
```diff
@@ -94,6 +94,10 @@ export class Scenario extends TimeUserEntityMetadata {
   @Column('uuid', { name: 'project_id' })
   projectId!: string;
 
+  @Column('integer', { name: 'project_scenario_id' })
+  @IsOptional()
+  projectScenarioId?: number;
```
hotzevzl, line 99: this should be not null, at least once we've run a migration to give a number to each of the scenarios already in an instance at the time we add this new column (so I think it should be created as null, and then altered to not null, but the entity should have this as always defined)

api/apps/api/src/modules/scenarios/scenario.api.entity.ts
```diff
@@ -94,6 +94,10 @@ export class Scenario extends TimeUserEntityMetadata {
   @Column('uuid', { name: 'project_id' })
   projectId!: string;
 
+  @Column('integer', { name: 'project_scenario_id' })
```
hotzevzl, line 97: I think you may have forgotten to add a migration to this PR's commits

besides adding the column to the table (I'd suggest to use a name that may make it clearer what it's for, as we deal with so many kinds of ids... maybe `project_scenario_numeric_id`?), I recommend to ensure that it is `not null` and unique within a given project it (`alter table scenarios add constraint unique_scenario_numid_per_project unique (project_id, row_count);`)

api/apps/api/src/modules/scenarios/scenarios-crud.service.ts
```diff
@@ -218,4 +221,29 @@ export class ScenariosCrudService extends AppBaseService<
...
+        await this.repository.findOneOrFail({
+          where: { id: scenario.id },
+        }),
+      );
+    } catch (e) {
+      await this.repository.delete(scenario.id);
+      return left(scenarioNotCreated);
+    }
+  }
 }
```
hotzevzl, line 249: something like this should also work _and_ would offload some logic to postgresql, so for example in the case of cloned projects we should not need to handle `project_scenario_numeric_id` in piece exporters and piece importers, as long as we're ok with numeric ids potentially not being consistent between clones of the same project, but I think that's fine because if users delete some scenarios from a cloned projects and then create new ones, in some cases numids may be "recycled" as we don't keep track (nor should, in my view) of already-assigned numids.

When importing a project and its scenarios, numids would be automatically assigned on insert.

So far we only expect these numids to be used to associate best and summed solutions to individual scenarios, so I don't think that having inconsistent numids between clones should be a problem.

```
CREATE OR REPLACE FUNCTION next_numid_for_scenario_in_project()
RETURNS TRIGGER AS $$
DECLARE
  numid integer;
BEGIN
    numid := (
        SELECT COALESCE(MAX(project_scenario_numeric_id), 0) + 1
        FROM scenarios
        WHERE project_id = NEW.project_id
    );
    NEW.project_scenario_numeric_id := numid;
    RAISE NOTICE 'Setting numid of new scenario % to %', NEW.id, numid;
    RETURN NEW;
END;
$$ LANGUAGE plpgsql;

CREATE TRIGGER set_numid_for_scenario_in_project
BEFORE INSERT ON scenarios
FOR EACH ROW
EXECUTE FUNCTION next_numid_for_scenario_in_project();
```
If going trigger-happy, I think `BEFORE INSERT` should be enough - combined with a non null constraint and unique constraint, there should never be a need to update a scenario's numid when a scenario is updated (e.g. its name is changed)

api/apps/api/src/modules/scenarios/scenario.api.entity.ts
hotzevzl: and that max is 999, but I think you have already taken care of this in the migration, given one of the tests below - just add the migration to the PR 😉

hotzevzl (commented):
sorry, more comments after a fresh look.

I'd normally lean on leaving applications out of handling values that should be automatically assigned: whether this is the case here is debatable, I've left some notes inline, I think it is but in reality we don't fully know and likely noone does have a final take on this.

api/apps/api/test/project-scenarios.e2e-spec.ts
```diff
@@ -23,40 +26,72 @@ let fixtures: FixtureType<typeof getFixtures>;
 
 beforeEach(async () => {
   fixtures = await getFixtures();
-}, 12_000);
+}, 20000000);
```
hotzevzl, line 29: I assume these are typos - 20M milliseconds is ~5.5hrs 🤔

api/apps/api/test/project-scenarios.e2e-spec.ts
```diff
@@ -223,13 +261,39 @@ async function getFixtures() {
...
+        .set('Authorization', `Bearer ${ownerToken}`)
+        .send(minimalCreateScenarioDTO);
+
+      if (grabId) {
+        scenarioId = response.body.data.id;
+      }
+      return response;
+    },
+
+    WhenCreatingAScenarioWithOverMaxProjectScenarioId: async () => {
```
hotzevzl, line 296: is this accidentally incomplete? there's no explicit handling of project_scenario_id, and from what I can see the function is not even used in any test

api/apps/api/src/modules/scenarios/scenarios.service.ts
```diff
@@ -295,7 +297,14 @@ export class ScenariosService {
...
       return left(projectNotReady);
 
-    const scenario = await this.crudService.create(validatedMetadata, info);
+    let scenario;
+    try {
+      scenario = await this.crudService.create(validatedMetadata, info);
+      scenario = await this.crudService.getById(scenario.id);
+    } catch (e) {
+      return left(scenarioNotCreated);
+    }
```
hotzevzl, line 306: Very minor, but... I am not sure about the rationale for this: why do we expect to handle exceptions when creating the scenario in a more surgical way, so to speak, than before?

In case the creation fails (it's the 1000th scenario in the project, for example), then `create()` should fail outright and this would trigger the global exception filter in the previous implementation where we were not using a `try` block... I think that should be enough (especially as we don't provide additional information about the _why_ of a possible failure here - our error message says that the scenario could not be created within the given project, but we don't give a reason (not saying we should - even discriminating between violation of different constraints, from `pg` errors, is kinda clunky and overkill IMO)

Also, shouldn't we get the scenario back from the call to `create()`, without having to `getById()` separately?

I think we can let things fail here, and as long as we do have a valid scenario returned by `create()` then the rest of the method can continue (otherwise, if `scenario` is undefined we should `return left(scenario)` or if an exception is thrown, either catch it - as your implementation does - or just let it bubble up).

hotzevzl (approved):

api/apps/api/test/project-scenarios.e2e-spec.ts
KevSanchez: Yes and no. I accidentally left them in, but because I need them locally, since the tests will always time out on my machine hahaha

api/apps/api/src/modules/scenarios/scenarios.service.ts
KevSanchez: Yeah, I don't like it honestly, but for whatever reason, the create method wasn't returning the full entity, and in the response I would always get a projectScenarioId undefined. This would make subsequent tests that check for projectScenarioId fail.

Maybe it's not that big of a deal? Maybe instead of checking the response from the create endpoint, I could simply GET the scenario, and check the projectScenarioId there, but it still irks me that I couldn't get the full info. I left it like this because I thought that if it fails on the creation because of the MAX 999 id, it would be nicer to receive a more user friendly message instead of a 500.

The save method not returning the full entity seems like it was a bug bug/missing feature/whatever in typeorm, but it doesn't look like it ever got resolved
https://github.com/typeorm/typeorm/pull/5680

api/apps/api/test/project-scenarios.e2e-spec.ts
KevSanchez: Not used, forgot to remove after reconsidering the approach for that test

api/apps/api/src/modules/scenarios/scenarios.service.ts
hotzevzl: ok, yes - I understand your rationale for this now, thanks.

The discussion past this comment in this thread is particularly insightful: https://github.com/typeorm/typeorm/issues/3490#issuecomment-585630457

Basically db `default`s and values set via trigger would not be set.

We could do what TypeORM doesn't do by implementing `BaseService.extendCreateResult()` for `ScenariosService`, something like

```
  async extendCreateResult(entity: Scenario, _createModel: CreateScenarioDTO): Promise<Scenario> {
    return await this.crudService.getById(entity.id);
  }
```

but pragmatically, it's just moving the `getById` in a different place, so I'm fully ok with keeping your implementation - maybe add a comment to clarify that we `create()` and then `getById()` in order to have the full entity back because some properties are set via trigger or defaults, and let's move on with life 😀

PR #1360, opened 2023-07-03
[api] Fix setByUser filter bug when changing status by kind

yulia-bel:
### Overview

- New method added to correctly find the PUs that has have been set by user and NOT coincide with status kind to be clear (basically adding filter 
- New separate method for finding the above mentioned PUs with status 'available' - in this case we need to use Query Builder, because in DB available lock status is NULL, and with the latest update of Typeorm repository find method with where condition translates to 'WHERE scenarioPuData.lockin_status =NULL' which leads to corrupted results
- Test preconditions are expanded to have more PUs with both SetByUser property values of true and false, expected results updates

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units.service.ts
+++ b/api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units.service.ts
@@ -41,6 +41,7 @@ export class ScenarioPlanningUnitsService {
     });
   }
 
+  // @todo: Review the possibility to use @IsNull() in getByStatusSetByUser instead of having this method
   async getAvailablePUsSetByUser(
     scenarioId: string,
   ): Promise<ScenariosPlanningUnitGeoEntity[]> {
@@ -48,7 +49,7 @@ export class ScenarioPlanningUnitsService {
       .createQueryBuilder('scenarioPuData')
       .where('scenarioPuData.scenario_id = :scenarioId', { scenarioId })
       .andWhere('scenarioPuData.lockin_status IS NULL')
-      .andWhere('scenarioPuData.lock_status_set_by_user = false')
+      .andWhere('scenarioPuData.lock_status_set_by_user = true')
       .getMany();
   }
 
--- a/api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units.service.ts
+++ b/api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units.service.ts
@@ -28,7 +28,7 @@ export class ScenarioPlanningUnitsService {
     });
   }
 
-  async getByStatus(
+  async getByStatusSetByUser(
     scenarioId: string,
     lockStatus: LockStatus,
   ): Promise<ScenariosPlanningUnitGeoEntity[]> {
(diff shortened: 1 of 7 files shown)


api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units.service.ts
```diff
@@ -28,18 +28,30 @@ export class ScenarioPlanningUnitsService {
...
 
-  async getByStatus(
+  async getByStatusSetByUser(
     scenarioId: string,
     lockStatus: LockStatus,
   ): Promise<ScenariosPlanningUnitGeoEntity[]> {
     return await this.puRepo.find({
       where: {
         scenarioId,
         lockStatus,
```
hotzevzl, line 38: would it not work (sorry, I'm the kind of knolling OCD person 🤷🏼 ) to do something like
```suggestion
        lockStatus: [LockStatus.LockedIn, LockStatus.LockedOut].includes(lockStatus) ? lockStatus : IsNull()
```
here and avoid using the new `getAvailablePUsSetByUser()` method? I don't remember what the TypeORM breaking change we faced at some point with null-related queries was: I think it was something related to `is null`, but I am not sure, maybe this suggestion is nonsense.

If it could actually work, I would still wrap it in a tiny method that does the `LockStatus`-to-db-values mapping, and I think we even have something like this, but perhaps just in terms of class-transformer in the entity 🤔

hotzevzl (approved):

api/apps/api/src/modules/scenarios/planning-units/scenario-planning-units.service.ts
yulia-bel: According to docs, the IsNull() should work in fin options, however it keeps getting translated to '= NULL',
even avoiding ternary operator and leaving just `status: IsNull()` leads to this.. out of curiosity I also left just one where option (lockStatus: IsNull() ), still same result..
```
async getByStatusSetByUser(
    scenarioId: string,
    lockStatus: LockStatus,
  ): Promise<ScenariosPlanningUnitGeoEntity[]> {
    return await this.puRepo.find({
      where: {
        scenarioId,
        lockStatus: IsNull(),
        setByUser: true,
      },
    });
  }
```
`query: SELECT "ScenariosPlanningUnitGeoEntity"."id" AS "ScenariosPlanningUnitGeoEntity_id", "ScenariosPlanningUnitGeoEntity"."project_pu_id" AS "ScenariosPlanningUnitGeoEntity_project_pu_id", "ScenariosPlanningUnitGeoEntity"."scenario_id" AS "ScenariosPlanningUnitGeoEntity_scenario_id", "ScenariosPlanningUnitGeoEntity"."lockin_status" AS "ScenariosPlanningUnitGeoEntity_lockin_status", "ScenariosPlanningUnitGeoEntity"."lock_status_set_by_user" AS "ScenariosPlanningUnitGeoEntity_lock_status_set_by_user", "ScenariosPlanningUnitGeoEntity"."xloc" AS "ScenariosPlanningUnitGeoEntity_xloc", "ScenariosPlanningUnitGeoEntity"."yloc" AS "ScenariosPlanningUnitGeoEntity_yloc", "ScenariosPlanningUnitGeoEntity"."protected_area" AS "ScenariosPlanningUnitGeoEntity_protected_area", "ScenariosPlanningUnitGeoEntity"."protected_by_default" AS "ScenariosPlanningUnitGeoEntity_protected_by_default", "ScenariosPlanningUnitGeoEntity"."feature_list" AS "ScenariosPlanningUnitGeoEntity_feature_list" FROM "scenarios_pu_data" "ScenariosPlanningUnitGeoEntity" WHERE ("ScenariosPlanningUnitGeoEntity"."scenario_id" = $1 **AND "ScenariosPlanningUnitGeoEntity"."lockin_status" = $2** AND "ScenariosPlanningUnitGeoEntity"."lock_status_set_by_user" = $3) -- PARAMETERS: ["32cedb9f-1a9a-4278-84f2-a79f28454fd4",null,true]`

PR #1375, opened 2023-07-13
feat(Marxan): Adds new service to generate Project Summaries after a run

KevSanchez:
Adds a new OutputProjectSummariesService that will generate a series of CSV files with summary info for all the scenarios runs in a given project, and store a zip with those files on the DB. This will be triggered after completing a Marxan run, on the API side at the end of the whole process.

### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/projects/output-project-summaries/output-project-summaries.service.ts
+++ b/api/apps/api/src/modules/projects/output-project-summaries/output-project-summaries.service.ts
@@ -1,0 +1,291 @@
+import { InjectRepository } from '@nestjs/typeorm';
+import { Repository } from 'typeorm';
+import { Scenario } from '@marxan-api/modules/scenarios/scenario.api.entity';
+import { BestSolutionDataService } from '@marxan-api/modules/projects/output-project-summaries/solution-data/best-solution-data.service';
+import { SummedSolutionDataService } from '@marxan-api/modules/projects/output-project-summaries/solution-data/summed-solution-data.service';
+import { ProjectsPuEntity } from '@marxan-jobs/planning-unit-geometry';
+import { CsvFormatterStream, FormatterRow, write } from 'fast-csv';
+import { execSync } from 'child_process';
+import { createWriteStream, existsSync, readFileSync } from 'fs';
+import { pipeline } from 'stream/promises';
+import { join as joinPath } from 'path';
+import { Injectable, Logger } from '@nestjs/common';
+import { OutputProjectSummaryApiEntity } from '@marxan/output-project-summaries';
+import { mkdir, rm } from 'fs/promises';
+import { DbConnections } from '@marxan-api/ormconfig.connections';
+import { AppConfig } from '@marxan-api/utils/config.utils';
+
+export const outputProjectSummaryFolder = 'output-project-summary/';
+export const outputProjectSummaryFilename = 'summary.csv';
+export const outputProjectSummaryScenariosFilename = 'scenarios.csv';
+
+@Injectable()
+export class OutputProjectSummariesService {
+  private readonly logger = new Logger(OutputProjectSummariesService.name);
+
+  constructor(
+    private readonly bestSolutionsDataService: BestSolutionDataService,
+    private readonly summedSolutionsDataService: SummedSolutionDataService,
+    @InjectRepository(Scenario)
+    private readonly scenarioRepo: Repository<Scenario>,
+    @InjectRepository(ProjectsPuEntity, DbConnections.geoprocessingDB)
+    private readonly projectsPU: Repository<ProjectsPuEntity>,
+    @InjectRepository(OutputProjectSummaryApiEntity)
+    private readonly outputSummaryRepo: Repository<OutputProjectSummaryApiEntity>,
+  ) {}
+
+  async saveSummaryForProjectOfScenario(scenarioId: string): Promise<void> {
+    const { projectId } = await this.scenarioRepo.findOneOrFail({
+      where: { id: scenarioId },
+    });
+
+    const planningUnits = (await this.projectsPU.find({ where: { projectId } }))
+      .map((pu) => pu.puid)
+      .sort();
+
+    const scenarios = await this.scenarioRepo.find({
+      select: { id: true, projectScenarioId: true, name: true },
+      where: { projectId },
+    });
+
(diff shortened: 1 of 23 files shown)


api/apps/api/src/modules/projects/output-project-summaries/output-project-summaries.service.ts
```diff
@@ -0,0 +1,291 @@
...
+    const bestSolutionData = await this.getBestSolutionDataForScenarios(
+      scenarios,
+    );
+    const summedSolutionData = await this.getSummedSolutionDataForScenarios(
+      scenarios,
+    );
+
+    const workingDirectory = OutputProjectSummariesService.getProjectTempFolder(
+      projectId,
+    );
```
hotzevzl, line 60: I think we discussed that this may be highly unlikely, but... wouldn't it be safer to use a unique directory each time this code is run?

hotzevzl: if so, I would still do it in a separate "refinement" PR: for the moment I'd say this is ok, in the interest of moving the PR on, in any case.

hotzevzl: also, it's not just a case of "if several of these project summaries are compiled more or less at the same time for a single project", but also (tin foil hat on), if anywhere else in the code someone adds _something_ later on that uses the same path structure for something completely unrelated, then boom.

tl;dr there are very good reasons for `mktemp` to be an essential tool ;)

api/apps/api/src/modules/projects/output-project-summaries/output-project-summaries.service.ts
```diff
@@ -0,0 +1,291 @@
...
+    headers.push('puid');
+    for (const projectScenarioId of projectScenarioIds) {
+      const scenarioId = this.formatProjectScenarioId(projectScenarioId);
+      headers.push(`S${scenarioId}_best`);
+      headers.push(`S${scenarioId}_ssoln`);
+    }
+
+    const csvData = [];
+    csvData.push(headers);
+    for (const planningUnit of planningUnits.sort()) {
```
hotzevzl, line 116: same as above - we don't want a lexicographical sorting but a numeric sorting here

api/apps/geoprocessing/test/steps/given-scenario-pu-data-exists.ts
```diff
@@ -187,3 +188,40 @@ export const GivenScenarioPuData = async (
     };
   });
 };
+
+export const GivenProjectPuData = async (
```
hotzevzl, line 192: have you checked if we don't have these elsewhere? looking at what these fixtures do rings some kind of bells... but I'm not sure :/

hotzevzl (approved):
thanks @KevSanchez , I haven't _run_ this but overall it should do what it should.

just check my notes about the use of `.sort()` (unless lexicographical sorting is what you had in mind, but I don't think this should be the case).

api/apps/api/src/modules/projects/output-project-summaries/output-project-summaries.service.ts
```diff
@@ -0,0 +1,291 @@
...
+
+  private arrangeSummaryCSV(
+    planningUnits: number[],
+    scenarios: Scenario[],
+    bestSolutionData: ProjectScenarioBestSolutionMapping,
+    summedSolutionData: ProjectScenarioSummedSolutionMapping,
+  ): CsvFormatterStream<FormatterRow, FormatterRow> {
+    const projectScenarioIds = scenarios
+      .map((scenario) => scenario.projectScenarioId)
+      .sort();
```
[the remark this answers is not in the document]
KevSanchez: I assumed that having the type number[] it would apply a sorting algorithm that makes sense for that, but yeah, you're right; this will fail for larger planningUnitIds

api/apps/api/src/modules/projects/output-project-summaries/output-project-summaries.service.ts
KevSanchez: Hmm, I had the temp folders to be based on the projectId so that it could be more easily identified in case of debugging and looking at the files in staging/production, but you got a good point there.
I'll do it on a separate PR then.

api/apps/api/src/modules/projects/output-project-summaries/output-project-summaries.service.ts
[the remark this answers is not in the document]
hotzevzl: yep, this being Yavascript, WAT is always around the corner: https://www.destroyallsoftware.com/talks/wat

to be fair, once the TS is transpiled to JS, types are erased so the `Array.prototype.sort()` function is necessarily (?) going to be dumber than we'd wish 🤷🏼 

another option would be to use _true_ JS typed arrays such as `UInt32Array`, for which the `.sort()` method would do a proper numeric sorting, but your fix is idiomatic and doesn't rely on creating new typed arrays out of primitive arrays.

hotzevzl:
@KevSanchez sorry I had missed this yesterday:
https://github.com/Vizzuality/marxan-cloud/actions/runs/5586804904/jobs/10211303525?pr=1375#step:3:4125

we'll also need this:

```diff
diff --git a/api/api.Dockerfile b/api/api.Dockerfile
index 53ae8f93c..add17526e 100644
--- a/api/api.Dockerfile
+++ b/api/api.Dockerfile
@@ -5,7 +5,7 @@ ENV NAME marxan-api
 ENV USER $NAME
 ENV APP_HOME /opt/$NAME

-RUN apk add --no-cache bash
+RUN apk add --no-cache bash zip
 RUN addgroup $USER && adduser -s /bin/bash -D -G $USER $USER

 WORKDIR $APP_HOME
```

api/apps/geoprocessing/test/steps/given-scenario-pu-data-exists.ts
KevSanchez: I didn't find anything equivalent at first glance, but I did notice there was some redundancy to the way I refactored the functions so that it would affect other tests suites that were already using the combined scenario + project Pu function. Now it should look better

PR #1376, opened 2023-07-17
Fix protectedByDefault over-ride for makeAvailable claims

yulia-bel:
### Overview

When PUs are in locked-in by default status because of being considered protected, if I make some of these “available” and I then clear the “available” status, the PUs that were explicitly set as available earlier on are now not locked in again (the PU fill-color stays blue but they are not coloured with the pale blue-green stroke on their edges).

This is a bug in api applyClaims() method that resets protectedByDefault to false when applying makeAvailable claims.

PR removes protectedByDefault reset and adds a new test covering the use case

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/geoprocessing/test/integration/planning-unit-inclusion/planning-unit-inclusion.e2e-spec.ts
+++ b/api/apps/geoprocessing/test/integration/planning-unit-inclusion/planning-unit-inclusion.e2e-spec.ts
@@ -361,5 +361,57 @@ describe(`when planning units exist for a scenario`, () => {
 
       expect(lockedOutPUsAfterClearingExcluded).toEqual([]);
     }, 10000);
+
+    it(`updates protected by default PUs once made available back to locked-in after clearing makeAvailable selection`, async () => {
+      /**
+       * Running the processor with no claims to make sure protected by default PUs
+       * are marked as locked in and value of lock status set by user is false
+       **/
+
+      await sut.process(({
+        data: {
+          scenarioId: world.scenarioId,
+        },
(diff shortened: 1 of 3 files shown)


api/apps/geoprocessing/test/integration/planning-unit-inclusion/planning-unit-inclusion.e2e-spec.ts
```diff
@@ -361,5 +361,57 @@ describe(`when planning units exist for a scenario`, () => {
...
+
+      /**
+       * Making all protectedByDefault PUs available
+       **/
+      await sut.process(({
+        data: {
+          scenarioId: world.scenarioId,
+          makeAvailable: {
+            pu: lockedInPUs,
+          },
```
hotzevzl, line 391: _don't change anything in response to this comment_ (the test is good enough like this) - just a general comment: to try to rule out any behaviours related to _something else_, I'd generally recommend to apply "under test" statuses to a _subset_ of a full set

in this case, since we only want to make sure that if we make N PUs available out of M PUs that are locked in, then locked in PUs will be M-N and that PUs with user claims will be N, we could use something like

```suggestion
          makeAvailable: {
            pu: lockedInPUs.slice(0,2),
          },
```

and then check that 3 PUs have manually-set claims and that locked-in PUs are `lockedInPUs.length - 3` (using meaningful constants rather than magic numbers planted in the source, but just to illustrate).

hotzevzl (approved):

PR #1378, opened 2023-07-17
(feat): run scenario

anamontiaga:
### Designs

[Figma](https://www.figma.com/file/0xD4gR9pCqU7NROnlSW6ck/Marxan-Layout---Redesign?node-id=521%3A58009&mode=dev)

### Testing instructions

- It should not appear if at least we have not selected a feature.
- Once the scenario is running, the button will disappear until the run is over and it will become visible again.

### Feature relevant tickets

[MRXN23-236](https://vizzuality.atlassian.net/browse/MRXN23-236)

[MRXN23-236]: https://vizzuality.atlassian.net/browse/MRXN23-236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/layout/project/navigation/index.tsx
+++ b/app/layout/project/navigation/index.tsx
@@ -6,6 +6,9 @@ import { useRouter } from 'next/router';
 
 import { TippyProps } from '@tippyjs/react/headless';
 
+import { useRunScenario } from 'hooks/scenarios';
+import { useToasts } from 'hooks/toast';
+
 import Icon from 'components/icon';
 import { Popover, PopoverContent, PopoverTrigger } from 'components/popover';
 import Tooltip from 'components/tooltip';
@@ -55,6 +58,8 @@ export const Navigation = (): JSX.Element => {
   const isProjectRoute = route.startsWith('/projects/[pid]');
   const isScenarioRoute = route.startsWith('/projects/[pid]/scenarios/');
 
+  const { addToast } = useToasts();
+
   const [submenuState, setSubmenuState] = useState<{ [key in NavigationTreeCategories]: boolean }>({
     user: false,
     inventory: isProjectRoute && NAVIGATION_TREE.inventory.includes(tab),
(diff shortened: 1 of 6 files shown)


app/layout/project/navigation/index.tsx
```diff
@@ -55,19 +58,24 @@ export const Navigation = (): JSX.Element => {
...
+  const { addToast } = useToasts();
+
   const [submenuState, setSubmenuState] = useState<{ [key in NavigationTreeCategories]: boolean }>({
     user: false,
     inventory: isProjectRoute && NAVIGATION_TREE.inventory.includes(tab),
     gridSetup: isScenarioRoute && NAVIGATION_TREE.gridSetup.includes(tab),
     solutions: isScenarioRoute && NAVIGATION_TREE.solutions.includes(tab),
     advancedSettings: isScenarioRoute && NAVIGATION_TREE.advancedSettings.includes(tab),
   });
+  const [scenarioRunning, setScenarioRunning] = useState(false);
```
andresgnlez, line 70: Can you please test if using the `status` property of the mutation helps to avoid using a state here? https://tanstack.com/query/v4/docs/react/reference/useMutation

Ideally, `status.isLoading` should do the same.

app/layout/project/navigation/index.tsx
```diff
@@ -88,9 +96,41 @@ export const Navigation = (): JSX.Element => {
...
+            'run-start',
+            <>
+              <h2 className="font-medium">Success!</h2>
+              <p className="text-sm">Run started</p>
+            </>,
+            {
+              level: 'success',
+            }
+          );
+          console.info('Scenario runned succesfully', s);
```
andresgnlez, line 116: Also, I know this is a copy and paste from the _Parameters_ tab, but let's try to avoid console things.

app/layout/project/navigation/index.tsx
```diff
@@ -88,9 +96,41 @@ export const Navigation = (): JSX.Element => {
   }, []);
 
   const handleRunScenario = useCallback(() => {
-    // todo: define run scenario button behaviour
-    console.log('handleRunScenario', sid);
-  }, [sid]);
+    setScenarioRunning(true);
+
+    runScenarioMutation.mutate(
+      { id: `${sid}` },
```
andresgnlez, line 102: `sid` is already a string, so there shouldn't be a need to recast it.

PR #1379, opened 2023-07-17
[MRXN23-120]: fixes PU count with protected area set

andresgnlez:
### Overview

This PR attempts to fix the issue reported [here](https://vizzuality.atlassian.net/browse/MRXN23-120?focusedCommentId=25215) where, once a user created a scenario and sets a protected area, the locked-in PU of the protected area were reflected in the summary in the panel. This lead to display of the _Clear_ button but actually it does nothing as it's trying to revert to the original state of the PUs when it is already there.

Before:
image

The solution is filtering the, in this case, locked-in PUs (_included_ ones for humans) by the property `setByUser` as the protected area setup is not run by the user (well, it is, but the user does not set explicitly the status of the PU) so we can assume there's nothing to clear as there isn't anything set by the user yet.

This also implies those PUs won't be set with the same styles as if the user sets them (which I think is fine).

After: 
image

Note: this filtering also applies now to locked-out and available PUs.

### Designs

–

### Testing instructions

### Feature relevant tickets

---

- [x] Update CHANGELOG file

### DIFF
--- a/app/hooks/scenarios/index.ts
+++ b/app/hooks/scenarios/index.ts
@@ -740,8 +740,6 @@ export function useScenarioPU(
       excluded: ScenarioPlanningUnit['id'][];
       included: ScenarioPlanningUnit['id'][];
       available: ScenarioPlanningUnit['id'][];
-      // includedDefault: ScenarioPlanningUnit['id'][];
-      // excludedDefault: ScenarioPlanningUnit['id'][];
     }
   >
 ) {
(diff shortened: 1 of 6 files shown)


hotzevzl (approved):
works as expected, thanks!

🙌🏼 for the cleanup of unused code too 😃

andresgnlez:
@hotzevzl I am including in this PR as well displaying always the summary of locks, in case you want to review and merge.

PR #1380, opened 2023-07-18
feat(Project): Reworks Project/Scenario piece importers/exporters to …

KevSanchez:
…include output summary

### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


api/apps/geoprocessing/src/import/pieces-importers/scenario-metadata.piece-importer.ts
```diff
@@ -79,6 +79,7 @@ export class ScenarioMetadataPieceImporter implements ImportPieceProcessor {
         status: values.status,
         created_by: ownerId,
         solutions_are_locked: values.solutionsAreLocked,
+        project_scenario_id: values.projectScenarioId,
```
hotzevzl, line 82: I think this may end up... doing nothing? As in, the trigger for setting these `projectScenarioId`s should always "win", whatever we throw at the `INSERT` statement for this column.

Moreover, I think it doesn't matter if a scenario ends up having a numeric id in the "source" project and a different one in cloned copies of the parent project, so I would not even think about a way to strictly enforce the same numbering.

Also, if this piece importer is run as part of a scenario cloning operation (rather than scenario creation within the cloning of a project), we would not want to have _the same_ value for `projectScenarioId` in the cloned scenario, supposing we would be setting this directly, bypassing the trigger somehow in a different implementation.

TL;DR to avoid confusion and to keep this as simple as possible, I'd remove the handling of `projectScenarioId` in piece exporters/importers, altogether.

hotzevzl (approved):
substantially 👍🏼 , but please review my inline note about `projectScenarioId`s

api/apps/geoprocessing/src/import/pieces-importers/scenario-metadata.piece-importer.ts
KevSanchez: Good point about the cloning, there. I was concerned mostly just about maintaining consistency with what's in the output summary, because those maybe used in other systems, and the projectScenarioId is the only way to correlate them to the Scenarios on marxan, but then again, you're right; we should worry only about consistency and simplicity within the system.

PR #1390, opened 2023-07-24
feature(GeoFeatures): Adds scenario usage count to the find all resul…

KevSanchez:
…t of GeoFeatures

### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


api/apps/api/src/modules/geo-features/geo-features.service.ts
```diff
@@ -682,4 +698,49 @@ export class GeoFeaturesService extends AppBaseService<
...
+    const scenarioUsages: {
+      id: string;
+      usage: number;
+    }[] = await this.geoEntityManager
+      .createQueryBuilder()
+      .select('api_feature_id', 'id')
+      .addSelect('COUNT(DISTINCT sfd.scenario_id)', 'usage')
+      .from('scenario_features_data', 'sfd')
+      .where('sfd.api_feature_id IN (:...featureIds)', { featureIds })
+      .groupBy('api_feature_id')
```
hotzevzl, line 716: I think we need an
```suggestion
      .andWhere('sfd.scenario_id IN (:...projectScenarioIds)', { projectScenarioIds })
      .groupBy('api_feature_id')
```
(of course, providing the list of `projectScenarioIds` as argument to the function

the reason for this is - we have this bloody platform-wide feature, and although _the one in use right now_ in the platform is not meant to be used as-is (it should be split, which would create project-specific/custom features), we should account for users deciding to use it as-is, as well as possible future platform-wide features that may not be meant to be split.

Otherwise every user will see a count of _all the scenarios in the whole platform_ where each platform-wide feature is in use, which is not what we want.

I suggest to handle this by limiting the count of scenarios to those in the current project, so that we keep a uniform logic for these counts, even if in practice this information is not going to be used for the decision to allow to delete platform-wide features, because every feature for which `isCustom === false` cannot be deleted in any case. But we can still tell users in how many scenarios the feature is used (may me more useful in the far future, if platform-wide features that can be used as-is are ever added).

hotzevzl: I am not sure if it's worth adding a test for this - maybe it is.

I'd add arrange things by creating two projects and at least a scenario in each, then a platform-wide feature (`project_id = NULL`), and then add this feature in both scenarios (one for each project), and check that the usage count is 1 for one of the two projects.

api/apps/api/src/modules/geo-features/geo-features.service.ts
```diff
@@ -682,4 +698,49 @@ export class GeoFeaturesService extends AppBaseService<
...
+    feature: GeoFeature,
+  ): Promise<GeoFeature> {
+    const [usage]: {
+      count: number;
+    }[] = await this.geoEntityManager
+      .createQueryBuilder()
+      .select('COUNT(DISTINCT sfd.scenario_id', 'count')
+      .from('scenario_features_data', 'sfd')
+      .where('sfd.api_feature_id = featureId', { featureId: feature.id })
+      .execute();
```
hotzevzl, line 739: likewise as above for the `extendFindAll` case

PR #1412, opened 2023-08-01
use current queryRunner rather than a free-floating data source when checking if a migration can rely on PostgreSQL v13 or later [MRXN23-273]

hotzevzl:
### Testing instructions

(careful: this will destroy all data in the instance where the PR is checked!)

`make clean-slate && make start-api` should rebuild and start the API and geoprocessing services correctly (without this fix, the services would fail to start with errors related to the use of `PostgresqlUtils.version13Plus()` in the first migration that uses this function).

### Feature relevant tickets

https://vizzuality.atlassian.net/browse/MRXN23-273

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/migrations/api/1610395720000-AddSupportForAuthentication.ts
+++ b/api/apps/api/src/migrations/api/1610395720000-AddSupportForAuthentication.ts
@@ -6,7 +6,7 @@ export class AddSupportForAuthentication1610395720000
   implements MigrationInterface {
   async up(queryRunner: QueryRunner): Promise<any> {
     // Only CREATEDB privilege required in 13+ rather than SUPERUSER (ht @agnessa)
-    if (await PostgreSQLUtils.version13Plus()) {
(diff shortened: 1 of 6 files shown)


hotzevzl:
@yulia-bel could you please have a look at my mid-afternoon brain explosion leading to this PR when you have a chance?

IIRC you also run into this issue recently when setting up an instance in containers on your laptop - I don't know why this doesn't cause any issues every time we basically spin up a full new MaPP cluster in compose in CI and run all the migrations there, whereas it does explode miserably when upping a local dev environment (well, in practice it's because we run migrations explicitly in CI whereas we do run them automatically as part of service startup in local envs, though how exactly this makes the former run without issues is not something I've explored in detail, but whatever).

yulia-bel:
@hotzevzl yes, it happened when I re-created databases in my local, now I tested with make `clean-slate` and the `make start-api` - everything works and no connection errors!

yulia-bel (approved):

PR #1413, opened 2023-08-01
[api] improve insert performance when persisting features data from CSV uploads [MRXN23-256]

yulia-bel:
### Overview

Adding logs and small performance improvement for bulk insert query - according to tests and logs analysis the fastest way for now is to set CTE used for features_data inserts as NOT MATERIALIZED and using chunks of 1000 values.

Some postgres analyse results:
Screenshot from 2023-08-01 16-19-45

Possible further solutions - inserting data with async jobs, refactoring geometries handling for features data uploaded via csv.

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/geo-features/import/features-amounts-upload.service.ts
+++ b/api/apps/api/src/modules/geo-features/import/features-amounts-upload.service.ts
@@ -51,7 +51,9 @@ export class FeatureAmountUploadService {
 
     let newFeaturesFromCsvUpload;
     try {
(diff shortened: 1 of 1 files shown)


hotzevzl (approved):
while we think about a more resilient (as in, given the many moving parts) solution, this does what it says on the tin.

not too happy about the way this does not scale (as discussed, it's likely because we'd need a different query setup altogether), but again, very much good enough for now.

PR #1415, opened 2023-08-01
[MRXN23-264] Breadcrumbs

anamontiaga:
### Overview

New layout breadcrumbs component:

Screenshot 2023-08-01 at 18 34 04

### Designs

[Designs](https://www.figma.com/file/0xD4gR9pCqU7NROnlSW6ck/Marxan-Layout---Redesign?node-id=1453%3A14822&mode=dev)

### Testing instructions

Access dashboard, project and scenario through the app and check that the page you are on is blue in the breadcrumbs and also you can navigate with them.

### Feature relevant tickets

[JIRA task](https://vizzuality.atlassian.net/browse/MRXN23-264)


app/layout/project/navigation/breadcrumbs/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { cn } from 'utils/cn';
+
+export const Breadcrumbs = ({ className }: { className?: string }): JSX.Element => {
+  const { pathname, push, query } = useRouter();
+  const { pid, sid } = query as { pid: string; sid: string };
+  const projectQuery = useProject(pid);
+  const scenarioQuery = useScenario(sid);
+
+  return (
+    <nav className="flex font-sans text-xs text-white" aria-label="Breadcrumb">
```
andresgnlez, line 17: Why do you need a `font-sans` class at this point? You should have already inherited the font styles.

app/layout/project/navigation/breadcrumbs/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+      <ol role="list" className="flex items-center space-x-2">
```
andresgnlez, line 18: No need of `role="list"`, semantically, `ol` is already a list so no need to redeclare it.

app/layout/project/navigation/breadcrumbs/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+            onClick={() => push('/projects')}
+            className={cn({
+              'flex h-6 items-center overflow-hidden hover:text-primary-500 focus:outline-none':
+                true,
+              'font-semibold text-primary-500': pathname === '/projects',
+              [className]: !!className,
+            })}
+          >
+            Dashboard
+          </button>
```
andresgnlez, line 30: Why not just set a `<Link />` ? 🤔

app/layout/project/navigation/breadcrumbs/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+        </li>
+        {pid && (
+          <li>
```
andresgnlez, line 33: I would include the right arrow inside the `<li>` element, it's not that important to give it its own space in the list. Also, why are you using a raw SVG?

app/layout/project/navigation/breadcrumbs/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+    <nav className="flex font-sans text-xs text-white" aria-label="Breadcrumb">
+      <ol role="list" className="flex items-center space-x-2">
+        <li>
+          <button
+            onClick={() => push('/projects')}
+            className={cn({
+              'flex h-6 items-center overflow-hidden hover:text-primary-500 focus:outline-none':
+                true,
+              'font-semibold text-primary-500': pathname === '/projects',
+              [className]: !!className,
```
andresgnlez, line 26: It's weird the `className` the component receives affects exclusively this very element, I would expect to apply it on line 17 so I can adapt the breadcrumbs to other contexts if needed but not here.

app/layout/project/navigation/breadcrumbs/index.tsx
```diff
@@ -0,0 +1,93 @@
+import React from 'react';
```
andresgnlez, line 1: Any reason to import React here?

PR #1420, opened 2023-08-03
upgrade NestJS to v10 [MRXN23-283]

hotzevzl:
### Testing instructions

API and Geoprocessing services should Just Work (tm) as before 😬 😫 

### Feature relevant tickets

https://vizzuality.atlassian.net/browse/MRXN23-283

---

- [ ] Update CHANGELOG file


yulia-bel:
All my basic workflow seem to work as usual (create project with grid, upload csv with features, create and run scenario), so should be fine

api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@marxan-cloud/api",
-  "version": "0.4.0",
+  "version": "2.0.35",
```
alexeh, line 3: now this is a bump

alexeh (approved):
Noice! As mentioned, the only issue I had upgrading ALL THE THINGS to v10 was with the commander, which we are not really using it, so... in CI we trust, LGTM!

api/package.json
hotzevzl: more like... nobody ever cared to bump our own version as part of our minimalistic "release" "process" so far 😬 ... we've been on v1.2.x for a while during the previous project phase for a while, then we had a stray v1.3.0.5 and then I lost my patience and took the BDFL decision to just call this phase v2.x 😬

PR #1423, opened 2023-08-04
return JSON:API representation of new feature when successfully creating a feature from shapefile [MRXN23-289 ]

hotzevzl:
### Testing instructions

`POST /api/v1/projects/:projectId/features/shapefile` should return a JSON:API representation of the uploaded feature from shapefile, upon success.

### Feature relevant tickets

https://vizzuality.atlassian.net/browse/MRXN23-289

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/projects/projects.controller.ts
+++ b/api/apps/api/src/modules/projects/projects.controller.ts
@@ -24,6 +24,7 @@ import {
 import { projectResource, ProjectResultSingular } from './project.api.entity';
 import {
   ApiBearerAuth,
+  ApiBody,
   ApiConsumes,
   ApiCreatedResponse,
   ApiForbiddenResponse,
@@ -729,8 +730,12 @@ export class ProjectsController {
   @ApiOperation({
     description: `Upload shapefiles of species or bioregional features.`,
   })
-  @ApiOkResponse({ type: ShapefileUploadResponse })
+  @ApiOkResponse({ type: GeoFeature })
   @ApiTags(inlineJobTag)
+  @ApiBody({
+    description: 'Shapefile to upload',
+    type: UploadShapefileDTO,
+  })
   @Post(`:id/features/shapefile`)
   @GeometryFileInterceptor(GeometryKind.ComplexWithProperties)
   async uploadFeatures(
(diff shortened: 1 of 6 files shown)


api/apps/api/src/modules/projects/projects.controller.ts
```diff
@@ -736,7 +737,7 @@ export class ProjectsController {
     @Param('id') projectId: string,
     @UploadedFile() shapefile: Express.Multer.File,
     @Body() body: UploadShapefileDTO,
-  ): Promise<ShapefileUploadResponse> {
+  ): Promise<GeoFeature> {
```
yulia-bel, line 740: shall we change @ApiOkResponse({ type: ShapefileUploadResponse }) -> @ApiOkResponse({ type: GeoFeature }) ?

hotzevzl: you're right, done in 53a901080da06bda7c29dbe234a4a870529fcaf8

I should probably put forward a more structured suggestion for Swagger documentation - we've been dealing with this in a bit of a copy-and-paste way without much consideration for the actual use 😢

hotzevzl:
> Worked as expected! The only small note - seems like in swagger the body is not described correctly (it has just 'file' there, while 'name' is required in body as well).. This is strange, because according to docs everything is set correctly. I tried using @ApiBody() tag - it shows body structure as it is in UploadShapefileDTO, but doesn't show 'file' then.. which is weird Anyways, it is just a small detail, but I have flashbacks from conversations with FE xD

I think 53a901080da06bda7c29dbe234a4a870529fcaf8 likewise fixes this, according to what the latest docs say (https://docs.nestjs.com/openapi/operations#file-upload)

Screenshot 2023-08-08 at 18-17-16 Swagger UI

hotzevzl:
@yulia-bel I'm going to merge this since you had pre-approved it and I think your note is fully addressed (https://github.com/Vizzuality/marxan-cloud/pull/1423#issuecomment-1669930151).

Let's keep in mind that we could and definitely should deal with endpoint documentation with much more attention, cleaning old cruft up as we go - thanks for paying attention to this in your review and making me realise about the need to up our game on this.

PR #1428, opened 2023-08-07
use validateSync() (to avoid temporal dead zone?) [MRXN23-292]

hotzevzl:
Uploading CSV files of features from puvspr data is _always_ failing since... some days ago, because by the time we'd call `resolve(featureAmount)` (`on('end')`), in `featureAmountCsvParser()`, `featureAmount` would be null.

I haven't been able to check why the original implementation, which we didn't touch since it was done, has always been working as expected, and then suddenly stopped working, all the time, except it's working just fine when running in CI: I guess this should be best done by `git bisect`ing things, which could take some time.

In the meanwhile, the proposed fix in this PR does solve the problem that sprung up.

### Testing instructions

Uploading features from a puvspr CSV sample file should lead to features being correctly created (please use a valid features.csv file so that we don't have to take into account validation errors).

On the current (and recent) `develop`, the upload would fail with a db error such as

```
query failed: SELECT DISTINCT "features"."feature_class_name" FROM "features" "features" WHERE "features"."project_id" = $1 AND "features"."feature_class_name" IN () -- PARAMETERS: ["<id of the project the features are being uploaded to>"]
```

and in turn this fails because the `featureClassName`s that should be in the `IN()` part of the query are none - no data is returned when resolving the promise returned by `featureAmountCsvParser()`.

### Feature relevant tickets

https://vizzuality.atlassian.net/browse/MRXN23-292

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/geo-features/import/csv.parser.ts
+++ b/api/apps/api/src/modules/geo-features/import/csv.parser.ts
@@ -1,7 +1,7 @@
 import { FeatureAmountCSVDto } from '@marxan-api/modules/geo-features/dto/feature-amount-csv.dto';
-import { Stream } from 'stream';
+import { Readable } from 'stream';
 import { parseStream } from 'fast-csv';
-import { validateOrReject } from 'class-validator';
+import { validateSync } from 'class-validator';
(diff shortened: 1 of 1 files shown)


hotzevzl:
merging this as discussed during the sprint review meeting today

PR #1431, opened 2023-08-08
only mark scenario as run once it finished running successfully [MRXN23-280]

hotzevzl:
Only mark a scenario as `ranAtLeastOnce: true` after it has run fully and successfully.

### Testing instructions

Create new scenario, set it to run with a high enough number of iteration to give some time to check if the `Scenario.ranAtLeastOnce` flag is flipped from the initial default (`false`) to `true` while the Marxan solver is running.

Without this PR, the property should be flipped almost instantly after pressing the Run button (or triggering a marxan run via the API directly). With this PR, the property should only be flipped to `true` only after Marxan has run fully and successfully (so, it should stay `false` if the Marxan run is interrupted in flight or fails for whatever reason.

### Feature relevant tickets

https://vizzuality.atlassian.net/browse/MRXN23-280

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/scenarios/marxan-run/run.handler.ts
+++ b/api/apps/api/src/modules/scenarios/marxan-run/run.handler.ts
@@ -39,9 +39,6 @@ export class RunHandler {
       scenarioId: scenario.id,
       assets,
     });
-    await this.scenarios.update(scenario.id, {
-      ranAtLeastOnce: true,
-    });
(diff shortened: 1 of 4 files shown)


hotzevzl:
@yulia-bel don't forget a `yarn format` once done with everything here 😉

hotzevzl:
@yulia-bel thanks! let's merge 🚀

PR #1435, opened 2023-08-10
New ProjectCostSurfaceController [MRXN23-92]

alexeh:

- Adds a new ProjectCostSurfaceControler
- Moves all cost surface related operations from ScenarioController to ProjectCostSurfaceControler
- Updates tests / fixtures

Following the API conventions in the scope of this task, no functionality is added, besides creating new endpoints.
Access control is handled at service level since 49ea0d52c98c530cc6eb3250711b3476a51ba6d8 as opposed to how is done in different controlers, the only auth stuff being used is the built-in module with some custom decorators.

I also realized that there are no tests specifically hitting the moved endpoints, however acl methods being used in said endpoints are tested hitting another endpoints

This being said I think it would be redundant to add more tests hitting those new endpoints

### Testing instructions

CI happy we happy

### Feature relevant tickets

1. [MARXN-23-92](https://vizzuality.atlassian.net/browse/MRXN23-92)

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/projects/project-cost-surface.controller.ts
+++ b/api/apps/api/src/modules/projects/project-cost-surface.controller.ts
@@ -1,0 +1,132 @@
+import {
+  Controller,
+  forwardRef,
+  Get,
+  Header,
+  Inject,
+  Param,
+  ParseUUIDPipe,
+  Post,
+  Req,
+  Res,
+  UploadedFile,
+  UseGuards,
+} from '@nestjs/common';
+import { ApiOkResponse, ApiOperation, ApiTags } from '@nestjs/swagger';
+import { projectResource } from '@marxan-api/modules/projects/project.api.entity';
+import { apiGlobalPrefixes } from '@marxan-api/api.config';
+import { ApiConsumesShapefile } from '@marxan-api/decorators/shapefile.decorator';
+import {
+  GeometryFileInterceptor,
+  GeometryKind,
+} from '@marxan-api/decorators/file-interceptors.decorator';
+import { asyncJobTag } from '@marxan-api/dto/async-job-tag';
+import { RequestWithAuthenticatedUser } from '@marxan-api/app.controller';
+import {
+  AsyncJobDto,
+  JsonApiAsyncJobMeta,
+} from '@marxan-api/dto/async-job.dto';
+import { ensureShapefileHasRequiredFiles } from '@marxan-api/utils/file-uploads.utils';
+import { isLeft } from 'fp-ts/Either';
+import { mapAclDomainToHttpError } from '@marxan-api/utils/acl.utils';
+import { scenarioResource } from '@marxan-api/modules/scenarios/scenario.api.entity';
+import { ScenariosService } from '@marxan-api/modules/scenarios/scenarios.service';
+import { CostRangeDto } from '@marxan-api/modules/scenarios/dto/cost-range.dto';
+import { plainToClass } from 'class-transformer';
+import { Response } from 'express';
+import { marxanRunFiles } from '@marxan-api/modules/scenarios/scenarios.controller';
+import { JwtAuthGuard } from '@marxan-api/guards/jwt-auth.guard';
+import { ImplementsAcl } from '@marxan-api/decorators/acl.decorator';
+
+@ApiTags(projectResource.className)
+@Controller(`${apiGlobalPrefixes.v1}/projects`)
+export class ProjectCostSurfaceController {
+  constructor(
+    @Inject(forwardRef(() => ScenariosService))
+    public readonly scenarioService: ScenariosService,
+  ) {}
+
+  @ImplementsAcl()
+  @UseGuards(JwtAuthGuard)
+  @ApiConsumesShapefile({ withGeoJsonResponse: false })
+  @GeometryFileInterceptor(GeometryKind.ComplexWithProperties)
+  @ApiTags(asyncJobTag)
+  @Post(`:scenarioId/cost-surface/shapefile`)
+  async processCostSurfaceShapefile(
+    @Param('scenarioId') scenarioId: string,
+    @Req() req: RequestWithAuthenticatedUser,
+    @UploadedFile() file: Express.Multer.File,
+  ): Promise<JsonApiAsyncJobMeta> {
+    await ensureShapefileHasRequiredFiles(file);
+
+    const result = await this.scenarioService.processCostSurfaceShapefile(
+      scenarioId,
+      req.user.id,
+      file,
+    );
+
+    if (isLeft(result)) {
+      throw mapAclDomainToHttpError(result.left, {
+        scenarioId,
+        userId: req.user.id,
+        resourceType: scenarioResource.name.plural,
+      });
+    }
+    return AsyncJobDto.forScenario().asJsonApiMetadata();
+  }
+
+  @ImplementsAcl()
+  @UseGuards(JwtAuthGuard)
+  @Get(`:scenarioId/cost-surface`)
+  @ApiOkResponse({ type: CostRangeDto })
+  async getCostRange(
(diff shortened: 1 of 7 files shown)


api/apps/api/test/business-critical/marxan-run/fixtures.ts
```diff
@@ -42,7 +42,7 @@ export const getFixtures = async () => {
     },
     GivenCostSurfaceTemplateFilled: async () => {
       await request(app.getHttpServer())
-        .get(`/api/v1/scenarios/${scenario}/cost-surface/shapefile-template`)
+        .get(`/api/v1/projects/${scenario}/cost-surface/shapefile-template`)
```
alexeh, line 45: Tests where this precondition is used are skipped, plus the endpoint is gone as well, so dunno if we should remove it altogether

api/apps/api/src/modules/projects/project-cost-surface.controller.ts
```diff
@@ -0,0 +1,132 @@
...
+    );
+
+    if (isLeft(result)) {
+      throw mapAclDomainToHttpError(result.left, {
+        scenarioId,
+        userId: req.user.id,
+        resourceType: scenarioResource.name.plural,
+      });
+    }
+  }
```
hotzevzl, line 131: this controller handler had plenty of naming confusion: in practice it should stay in `ScenariosController` and should  be renamed to something like `GetPuDatFile()`; its Swagger decorators are also fully wrong (likely copypasta from somewhere else and we didn't catch this in review)

the confusion goes on through the rabbit hole: `ScenariosController.getCostSurfaceCsv()` is likewise more like `getPuDatCsv()` and `InputFilesService.readCostSurface()` again is more like `getPuDatContent()`

(also, `InputFilesService.get<whatever>DatContent()` should be async, and they should await the async functions that they call, but well... the rabbit hole must end somewhere)

api/apps/api/src/modules/projects/project-cost-surface.controller.ts
```diff
@@ -0,0 +1,132 @@
...
+    );
+    if (isLeft(result)) {
+      throw mapAclDomainToHttpError(result.left, {
+        scenarioId,
+        userId: req.user.id,
+        resourceType: scenarioResource.name.plural,
+      });
+    }
+    return plainToClass<CostRangeDto, CostRangeDto>(CostRangeDto, result.right);
+  }
```
hotzevzl, line 98: we'll likely end up changing this - the URL was kinda wonky all along (it should be something like `@Get(':scenarioId/cost-surface/cost-range')`) but I think in the id26.2 implementation we could _store_ min and max for each cost surface at upload/creation time (in `(apidb)cost_surfaces`) and just return that, no need to do even an index scan (I hope it is currently an index scan, not a table scan, I haven't checked right now the indexes we may rely on).

Frontend will need the `[min, max]` range both for individual cost surfaces (we use this to set the gradient colours) and for the active cost surface for each scenario, which could be done transparently in the API by using the same code path we'd use to retrieve the range of a cost surface, for the cost surface that is active in the given scenario.

TL;DR we'll need a "cost-range" endpoint both in this `ProjectCostSurfaceController` and in the `ScenariosController`, for the two use cases in the previous sentence 😓

api/apps/api/src/modules/projects/project-cost-surface.controller.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    if (isLeft(result)) {
+      throw mapAclDomainToHttpError(result.left, {
+        scenarioId,
+        userId: req.user.id,
+        resourceType: scenarioResource.name.plural,
+      });
+    }
+    return AsyncJobDto.forScenario().asJsonApiMetadata();
+  }
```
hotzevzl, line 76: for this handler, we'll need to coordinate with frontend and accept a `projectId` instead, and then move the `processCostSurfaceShapefile()` method to a different service (moving it out of `ScenariosService`, basically, as it won't belong there anymore).

If we want to break the fewest things possible, we may want to keep this method and its current code path in `ScenariosController` while we complete the basic workflow for id26.2, so that the current FE UI can still allow users to upload cost surfaces directly to a scenario, and then we (including FE UI) can switch over to the "upload to project, link to scenario" workflow once ready

api/apps/api/test/business-critical/marxan-run/fixtures.ts
hotzevzl: tbh I'd remove this function and its use in the skipped `Marxan run` test

api/apps/api/test/scenario-input-files/scenario-cost-surface/scenario-cost-surface.fixtures.ts
```diff
@@ -152,24 +152,24 @@ export const getFixtures = async () => {
...
         await request(app.getHttpServer())
-          .get(`/api/v1/scenarios/${scenarioId}/marxan/dat/pu.dat`)
+          .get(`/api/v1/projects/${scenarioId}/marxan/dat/pu.dat`)
           .set('Authorization', `Bearer ${viewerToken}`)
       ).text,
     WhenGettingMarxanDataAsUserNotInScenario: async () =>
       await request(app.getHttpServer())
-        .get(`/api/v1/scenarios/${scenarioId}/marxan/dat/pu.dat`)
+        .get(`/api/v1/projects/${scenarioId}/marxan/dat/pu.dat`)
         .set('Authorization', `Bearer ${userNotInScenarioToken}`),
```
hotzevzl, line 173: this will need to stay as it was 

eason: https://github.com/Vizzuality/marxan-cloud/pull/1435/files#r1289957576, with apologies for the naming confusion that leads to think quite naturally that this is related to cost surfaces, while it isn't (_in practice_ the data that is served through this endpoint _is_ a cost surface, but that's just an accident)

hotzevzl (commented):
besides any changes you may still want to add here, and after addressing the notes inline about endpoints that should stay (or _also_ stay) in `ScenariosController`, I would then merge this PR, and then keep building on it - no need to hold it back until more functionality is ready, especially if we can avoid breaking existing user flows that rely on the current endpoints (hence the suggestion to _also_ keep `processCostSurfaceShapefile()` in `ScenariosController` for the time being, just marking it as deprecated there maybe)

api/apps/api/src/modules/projects/project-cost-surface.controller.ts
alexeh: Right, I totally missed that in the scope of this task. I was planning to handle MRXN23-89, 23-89 and 23-90 as they felt kind of isolated enough and a good starting point.

I have decided to rollback and deprecate the endpoint in ScenariosController to not disturb any current functionality at then tmp throw a not implemented exception in the new controller. I hope this will adjust better to the scope until I get more understanding of the whole epic.

api/apps/api/src/modules/projects/project-cost-surface.controller.ts
alexeh: Again, I totally missed the big picture. The id26.2 you mention covers a lot of stuff that I am yet to fully understand. I am not sure how to proceed with this specific suggestion.

api/apps/api/src/modules/projects/project-cost-surface.controller.ts
alexeh: Right, rolling this back again as I considered to be a cost-surface related operation, as the task suggest to move. 
Thanks

api/apps/api/test/scenario-input-files/scenario-cost-surface/scenario-cost-surface.fixtures.ts
alexeh: addressed in 09d815d9091d397ad86d4a5bb7d369abbe29f6e2

PR #1436, opened 2023-08-10
tweak tolerance for st_removerepeatedpoints() [MRXN23-281]

hotzevzl:
For full context, see the Jira issue linked below

### Testing instructions

Tiles of features that include "very small" regular polygons should display correctly: for example, for features created from an hexagonal grid when uploading features from shapefiles, the polygons that make up the feature should show as filled hexagons, not as triangles that join only some of the vertices of the hexagons.

### Feature relevant tickets

https://vizzuality.atlassian.net/browse/MRXN23-281

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/geoprocessing/src/modules/features/features.service.ts
+++ b/api/apps/geoprocessing/src/modules/features/features.service.ts
@@ -73,7 +73,7 @@ export class FeatureService {
   ): Promise<Buffer> {
     const { z, x, y, id } = tileSpecification;
     const attributes = 'feature_id, properties';
-    const table = `(select ST_RemoveRepeatedPoints((st_dump(the_geom)).geom, 0.1) as the_geom, properties, feature_id from "${this.featuresRepository.metadata.tableName}")`;
+    const table = `(select ST_RemoveRepeatedPoints((st_dump(the_geom)).geom, 0.01) as the_geom, properties, feature_id from "${this.featuresRepository.metadata.tableName}")`;
     const customQuery = this.buildFeaturesWhereQuery(id, bbox);
     return this.tileService.getTile({
       z,
--- a/api/apps/geoprocessing/src/modules/features/features.service.ts
+++ b/api/apps/geoprocessing/src/modules/features/features.service.ts
@@ -1,7 +1,7 @@
 import { Injectable, Logger, Inject } from '@nestjs/common';
 import { TileService } from '@marxan-geoprocessing/modules/tile/tile.service';
-import { InjectRepository } from '@nestjs/typeorm';
-import { Repository } from 'typeorm';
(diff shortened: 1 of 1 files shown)


api/apps/geoprocessing/src/modules/features/features.service.ts
```diff
@@ -73,7 +73,7 @@ export class FeatureService {
   ): Promise<Buffer> {
     const { z, x, y, id } = tileSpecification;
     const attributes = 'feature_id, properties';
-    const table = `(select ST_RemoveRepeatedPoints((st_dump(the_geom)).geom, 0.1) as the_geom, properties, feature_id from "${this.featuresRepository.metadata.tableName}")`;
+    const table = `(select ST_RemoveRepeatedPoints((st_dump(the_geom)).geom, 0.01) as the_geom, properties, feature_id from "${this.featuresRepository.metadata.tableName}")`;
```
hotzevzl, line 76: I am not really happy about this - more of a hack than a fix.

This brutal lowering of the tolerance to `ST_RemoveRepeatedPoints()` _does_ make features display properly, when they are made up, for examples, of the small polygons of the TNC Okavango or Rwanda training projects; however, of course, make things much worse for features from complex shapefiles where there may be very close points that we could safely remove without changing much of the overall visual rendering.

Some things we could do:
- add a db query (to the apidb) before starting assembling the main query for each tile request, in `FeatureService.findTile()`, and lower the tolerance only for features that are "legacy/from-puvspr"

this could be ok - it's a very simple select, though it's one extra db roundtrip to the apidb for each tile for each feature; still, making a decision would not be so straightforward - grids can be of any size and shape, and I'm not sure how best we'd make a decision about the tolerance (I guess if we wanted the features to always match the underlying grid, we could fully remove the removal of repeated points in this case, but if a user uploads a very complex grid with indeed many very close points, we end up sending so much more data into these tiles)

- make the tolerance dependent on... something, such as the area of the polygon, or the length of its perimeter...

this would still be an very loose proxy, and moreover instead of a plain select we'd do spatial calculations (area) on the fly, with the added complication that we use st_dump on the geometries here

Also, as noted in the Jira story, once we decouple feature amounts from `(geodb)features_data` in id31.3, we could (and definitely should, I think) st_union all the polygons that make up each from-puvspr-data feature, in which case this whole issue may be come a bit less relevant - though of course, skipping points along the outer edge of the `ST_Union`ed polygons may still introduce artifacts, perhaps just not as blatant as the one that led us to take care of this.

hotzevzl: I ended up discriminating between legacy and non-legacy features, as PoC, but @aagm's input here will be very useful, as always!

hotzevzl:
closing this PoC: better implementation in #1491

PR #1443, opened 2023-08-17
[api] add endpoint to upload protected area shapefile for a project [MRXN23-267]

yulia-bel:
### Overview

- adds endpoint in `ProjectsController` to upload protected area shapefile for projectId
- adds service for adding protected area in `ProjectsService`
- adds `protected-area`module to `projects`
- add new event types and migration to upload new event types to DB
- update tests

Uploading protected area shapefile for scenario functionality alreade removed as part of this PR.

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


hotzevzl (approved):
thanks @yulia-bel! at a glance all looks good.

Ideally we should check that protected area data and metadata is fully and correctly created. But this is not something we did for the scenario-level approach until now, so while it would be an extra safety net for possible regressions, I'm also ok with not going out of our way on this (we'd need to poll for job status, and once the job succeeds, check that a corresponding record is created in `(geodb)wdpa` and maybe that its `hash` matches a know-good hash for whatever sample protected area shapefile we upload.

PR #1454, opened 2023-08-24
Rollback ScenariosController.getCostRange() as deprecated, fix param bug

yulia-bel:
## Rollback ScenariosController.getCostRange() as deprecated, fix param bug

After some investigation of the missing endpoint, the conclusions and sequence of events are the following:
- Following endpoints have been removed from ScenariosController and added to ProjectCostSurfaceController: processCostSurfaceShapefile, getCostRange, getScenarioCostSurface
- Then following endpoints were roll back as deprecated: processCostSurfaceShapefile, getScenarioCostSurface rolled back and renamed to getPuDatFile
- So getCostRange() hasn't been rolled back as deprecated to ScenariosController

In order for FE to be able to work with 'old' endpoints, they need to be the following (implemented in this PR) :
`GET /api/v1/scenarios/{scenarioId}/cost-surface`- to get cost surface range
`POST /api/v1/scenarios/{scenarioId}/cost-surface/shapefile` - to upload shapefil of cost surface

Regarding the bug we have now in marxan23 - for some reason, specifically in thos endpoint @Param() decorator is not working correctly and not extracting `scenarioId' param from query.. However, replacing it with `req.params.scenarioId`works fine

### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/scenarios/scenarios.controller.ts
+++ b/api/apps/api/src/modules/scenarios/scenarios.controller.ts
@@ -108,6 +108,8 @@ import { RequestScenarioCloneResponseDto } from './dto/scenario-clone.dto';
 import { ensureShapefileHasRequiredFiles } from '@marxan-api/utils/file-uploads.utils';
 import { WebshotPdfReportConfig } from '@marxan/webshot/webshot.dto';
 import { ClearLockStatusParams } from '@marxan-api/modules/scenarios/dto/clear-lock-status-param.dto';
+import { CostRangeDto } from "@marxan-api/modules/scenarios/dto/cost-range.dto";
+import { plainToClass } from "class-transformer";
 
 const basePath = `${apiGlobalPrefixes.v1}/scenarios`;
 const solutionsSubPath = `:id/marxan/solutions`;
@@ -399,7 +401,7 @@ export class ScenariosController {
   @ApiConsumesShapefile({ withGeoJsonResponse: false })
   @GeometryFileInterceptor(GeometryKind.ComplexWithProperties)
   @ApiTags(asyncJobTag)
-  @Get(`:scenarioId/cost-surface/shapefile`)
+  @Post(`:scenarioId/cost-surface/shapefile`)
   async processCostSurfaceShapefile(
     @Param(':scenarioId') scenarioId: string,
     @Req() req: RequestWithAuthenticatedUser,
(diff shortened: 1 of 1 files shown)


api/apps/api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -399,7 +401,7 @@ export class ScenariosController {
   @ApiConsumesShapefile({ withGeoJsonResponse: false })
   @GeometryFileInterceptor(GeometryKind.ComplexWithProperties)
   @ApiTags(asyncJobTag)
-  @Get(`:scenarioId/cost-surface/shapefile`)
+  @Post(`:scenarioId/cost-surface/shapefile`)
```
alexeh, line 404: Hey, I am not sure if this is correct. @andresgnlez mentioned that this should be a GET endpoint, I rolled it back in 7d12f1dbf03a6a7e89f64eca19ce9eed246af1dc

We mentioned that in future implementations (`ProjectCostSurfaceController`) since this has to receive a name, we could probably set it as POST, however since this is a deprecated endpoint, probably should be GET

api/apps/api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -408,21 +410,43 @@ export class ScenariosController {
     await ensureShapefileHasRequiredFiles(file);
 
     const result = await this.service.processCostSurfaceShapefile(
-      scenarioId,
+      req.params.scenarioId,
```
alexeh, line 413: This is weird. `@Param()` decorator does not extract que property from the `request.param` object?
Probably worth checking why, as this is widely used across the API.

Otherwise we should type it in the request, and remove the decorator altogether?

api/apps/api/src/modules/scenarios/scenarios.controller.ts
```diff
@@ -408,21 +410,43 @@ export class ScenariosController {
...
+    if (isLeft(result)) {
+      throw mapAclDomainToHttpError(result.left, {
+        scenarioId,
+        userId: req.user.id,
+        resourceType: scenarioResource.name.plural,
+      });
+    }
+    return plainToClass<CostRangeDto, CostRangeDto>(CostRangeDto, result.right);
+  }
+
```
alexeh, line 449: This is totally my fault, as it was removed in f1e3d8de931e6b3f4b2410ff8ef3731736ab7c34 as part of:
[MARXN-23-92](https://vizzuality.atlassian.net/browse/MRXN23-92)
Where it was required to remove all cost-surface related operations to the new controller. I assumed it was a safe operation. 

Thanks!

PS: It's double-weird that in the previous endpoint the `@Param()` decorator seems to not be working, and we expect here to do it. Let's discuss it later

api/apps/api/src/modules/scenarios/scenarios.controller.ts
yulia-bel: I agree - that's exactly what it should do and what it is doing in multiple other endpoints with identical implementation, so I checked many times, tried to remove other decorators (thought maybe some is overriding it), but it just kept as undefined.. 
Maybe it is worth leaving a todo note, to try to solve this mystery in the future, because it doesn't make sence

api/apps/api/src/modules/scenarios/scenarios.controller.ts
yulia-bel: there was a bit of confusion here, because @andresgnlez reported that the endpoint of `GET :id/cost-surface` was returning 404 - he meant the one without shapefile in the end and the one that return cost surface range (hat was removed and not rolled back).

PR #1460, opened 2023-08-28
feat(CostSurface): Adds new endpoint to update a Cost Surface [MRXN23-29]

KevSanchez:
For now, only the name is updated

### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


api/apps/api/src/modules/cost-surface/cost-surface.service.ts
```diff
@@ -52,4 +60,47 @@ export class CostSurfaceService {
...
+      ))
+    ) {
+      return left(projectNotEditable);
+    }
+
+    const nameCount = await this.costSurfaceRepository.count({
+      where: { name: dto.name },
+    });
+    if (nameCount > 0) {
+      return left(costSurfaceNameAlreadyExistsForProject);
```
alexeh, line 90: if we throw `costSurfaceNameAlreadyExistsForProject` shouldn't we filter by project as well?
plus, maybe for being consistent through the code, I would use `find()` and then check if the result is truthy or not, but just my two cents

KevSanchez: You're right, completely missed this. Thanks.

PR #1494, opened 2023-09-08
feat(CostSurface): scenario-cost surface relationship and migration of existing cost data [MRXN23-453]

KevSanchez:
…tomatic creation

This commit addresses the creation by default of cost surfaces when creating new Projects, and using the Project's default cost surface when creating a new Scenario. It also includes many changes to accomodate for the NOT NULL relationship.

### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


KevSanchez:
I've managed to get everything working except for the Scenario Piece Importer, which brings up an important topic to discuss; that being, what to do with cost surfaces when importing (probably exporting too) projects and scenarios, as I'm not _that_ familiar with all the internal steps involved, and how inconsistent data might be by just inserting default cost surfaces with no real corresponding data on the geoprocessing side.
@hotzevzl @alexeh

api/apps/api/src/migrations/api/1694192071502-AddScenarioToCostSurfaceAPIRelationship.ts
```diff
@@ -0,0 +1,16 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddScenarioToCostSurfaceAPIRelationship1694192071502
+  implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+        ALTER TABLE scenarios ADD COLUMN cost_surface_id uuid NOT NULL REFERENCES cost_surfaces(id);
```
hotzevzl, line 7: I think we'll have to make this nullable here, then within this same migration file:

- create as many default cost surfaces as there are project in a given MaPP instance, each linked to its project
- create as many cost surfaces as there are scenarios and link each of these to its scenario
- _copy_ all the `scenarios_pu_cost_data` over from each scenario's set of `scenarios_pu_cost_data` to the relevant cost surface data
- and finally make the `scenarios.cost_surface_id` fk not nullable

Alternatively, go for any of the Python or bash/psql based data migration paths you've mentioned, in which case this migration should "just" add the new column as nullable, then data migrations need to be done manually (do we _really want_ to do this, though?) and only then deploy a new "release" that includes a migration that makes the new column not null.

hotzevzl (commented):
thanks @KevSanchez!

besides my minor comment about possibly simplifying the name of default cost surfaces (tests may need to be updated too), the main point here is _whether_ to handle with data migrations (for all existing scenarios in any given instance) within this same PR, and if so, how to do this.

Personally, I'd favour a distinct PR that only includes a migration file that adds the new `scenarios.cost_surface_id` column, migrates data and finalises the column as not null. I'd leave all the rest of this PR alone here, depending on that other earlier PR.

Let's discuss this on Monday.

Apologies for taking so long on this.

hotzevzl:
> I've managed to get everything working except for the Scenario Piece Importer, which brings up an important topic to discuss; that being, what to do with cost surfaces when importing (probably exporting too) projects and scenarios, as I'm not _that_ familiar with all the internal steps involved, and how inconsistent data might be by just inserting default cost surfaces with no real corresponding data on the geoprocessing side. @hotzevzl @alexeh

@KevSanchez this is a bit of a tricky network of interdependencies (cc @yulia-bel because you're working on piece exporters/importers).

I suggest to tackle things in this order:

- finish the implementation in this PR (but see my inline note about splitting off of this the changes required to migrate existing cost data to the new data model we introduced in id26.2), avoiding to export and import the link between scenarios and cost surfaces for this step
- implement MRXN23-111 and MRXN23-112
- only then, implement MRXN23-19

In terms of what we discussed in our mob session this week:

- `ExportResourcePiecesAdapter` is where sequencing of export components (=instances of pieces) is set; here is where we need to add the sequencing of something like `ExportComponent.newOne(id, ClonePiece.ProjectCostSurfaces)` (probably we can simplify and treat the default cost surface like just any other project cost surface - I don't think we need to export and import this separately); the other side of this (import) is handled in `ImportResourcePiecesAdapter`
- in terms of how to handle the "re-linking" of cost surfaces to projects and scenarios on import, once we have new `id`s (UUIDs) for each cost surface that has been imported for a given project via its piece importer, an example of how this is done for an existing flow (importing project features and associating them with the `id` generated on import to the "new" parent project as well as to scenarios via scenario/features specifications), I suggest to have a look at `ScenarioFeaturesSpecificationPieceImporter.getFeaturesIdByNameAndProject()`; basically we need to rely on some kind of unique (single column or compound) identifier that is _not_ the `id`/UUID, and then, use that as "link" between the entities

For example, when exporting features, we export their name as is (this is a fragment of the `custom-features.json` file for a project):

```
{"features":[{"feature_class_name":"Chimpanzee", [...]
```

And when exporting feature specifications for scenarios, we establish the link in the exported data (in `features-specification.json` for a scenario of that same exported project):

```
[{"draft":true,"raw":{"status":"draft","features":[{"kind":"plain","featureId":"Chimpanzee/project", [...]
```

Here the `/project` suffix is added (and then parsed and used on import) to differentiate between the just-imported Chimpanzee custom feature and any possible identically named platform-wide features, which are not exported nor imported but assumed as already existing - noting this as an extra step needed for features, but for cost surfaces we won't need this!

In `ScenarioFeaturesSpecificationPieceImporter.getFeaturesIdByNameAndProject()` we then look up, given the name expected for a feature that is used in a scenario specification, its `id` (UUID) _as was generated for the newly imported feature_, and this is then used to link things up.

I feel this should be a brief piece of documentation in the repository. Happy to help next week - do have a look at the pieces of code from the pointers above before if you haven't seen them in detail already, for baseline context about this.

api/apps/api/src/migrations/api/1694192071502-AddScenarioToCostSurfaceAPIRelationship.ts
alexeh: Hullo @hotzevzl! We discussed a couple of ways to manage this. 
Here is a ugly py script that should be able to handle this and just reflect the database changes with the above migration (adding required IF NOT EXISTS statements...)

https://github.com/Vizzuality/marxan-cloud/pull/1501

Let's discuss it later

api/apps/api/src/modules/cost-surface/cost-surface.service.ts
```diff
@@ -107,4 +124,8 @@ export class CostSurfaceService {
 
     return right(costSurface);
   }
+
+  static defaultCostSurfaceName(projectName?: string): string {
+    return `${projectName ? projectName + ' - ' : ''}Default Cost Surface`;
+  }
```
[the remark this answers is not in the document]
KevSanchez: Another possibility would be having `Default cost surface for project {projectId}`, but then again, all operations on cost surfaces are done at Project level, so you're not really to ever going to have more than one "Default Cost Surface" on a GET list, so, yeah, what you say is probably the most ideal.

PR #1503, opened 2023-09-18
[api] Add cost surface piece exporter and importer [MRXN23-111] [MRXN23-19] [MRXN23-112] [MRXN23-20]

yulia-bel:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/geoprocessing/src/export/pieces-exporters/project-cost-surfaces.piece-exporter.ts
+++ b/api/apps/geoprocessing/src/export/pieces-exporters/project-cost-surfaces.piece-exporter.ts
@@ -33,7 +33,7 @@ type ProjectCostSurfacesSelectResult = {
 type CostSurfaceDataSelectResult = {
   cost_surface_id: string;
   cost: number;
-  project_pu_id: number;
+  projects_pu_id: number;
 };
 
 @Injectable()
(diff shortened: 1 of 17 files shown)


KevSanchez (approved):

api/apps/geoprocessing/src/export/pieces-exporters/project-cost-surfaces.piece-exporter.ts
```diff
@@ -0,0 +1,140 @@
...
+        data: costSurfaceData
+          .filter(
+            (data: CostSurfaceDataSelectResult) => data.cost_surface_id === id,
+          )
+          .map(({ cost_surface_id, projects_pu_id, ...data }) => {
+            const puid = projectPusMap[projects_pu_id];
+            return { puid, ...data };
+          }),
+      })),
+    };
```
hotzevzl, line 96: just a note on style - my recommendation for these kinds of "augmentations" is to use named functions to make what we're doing via array operations (filter, map, etc.), spreads and so on clearer.

In practice here we're combining a lookup of cost surface costs with the setter part of a lens, using lower-level primitives: which is ok and works, but makes even a tiny snippet of code like this less legible than something like `const fileContent: ProjectCostSurfacesContent = addCostDataToCostSurfaces(costSurfaces, costSurfaceCostData)`

but no need to do anything here! (and as always, these kinds of style comments can be highly opinable)

hotzevzl (commented):
@yulia-bel thanks, and apologies for taking so long to review this - I've been mostly down the rabbit hole of trying to make the functionality of this feature to work _in practice_ in a variety of contexts (projects with or without scenarios, with run scenarios, etc.) and as you know this has been a royal headache in its own right because of the moving pieces in related PRs 😬 

As it stands, the code itself looks fine, but:

- please rebase it on the latest `develop` and consider what to do with the check that I highlighted inline in my review, as after rebasing we'd be "guaranteed" (except if any bugs, etc) that by the time we clone a project, this would have at least the default cost surface
- check the functionality after rebasing - so far I have only been able to get cloning to work with a very specific combination of factors

hotzevzl (approved):

PR #1506, opened 2023-09-19
[api] Update GET project protected areas list endpoint [MRXN23-451]

yulia-bel:
### Overview

Response structure updated - new property 'name' added. For custom PAs it will contain 'fullName', for global - 'iucnCategory'.

Global PAs are grouped by category.

Sorting and filtering by 'name' property enabled.

Tests updated

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/decorators/json-api-parameters.decorator.ts
+++ b/api/apps/api/src/decorators/json-api-parameters.decorator.ts
@@ -76,6 +76,33 @@ const disablePaginationQueryParam = ApiQuery({
   required: false,
 });
 
+const protectedAreasFilterQueryParam = ApiQuery({
+  name: 'filter',
+  example: 'filter[name]=II&filter[name]=customAreaName',
+  description: `An array of filters for name property`,
+  type: String,
+  isArray: true,
+  required: false,
+});
+
+const protectedAreasSortQueryParam = ApiQuery({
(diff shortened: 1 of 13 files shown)


hotzevzl (commented):
thanks @yulia-bel. Your PR does what we need, but looking at it in details made me feel more uncomfortable as I kept thinking about how we're bending nestjs-base-service... so well, besides a long tirade about my own irrelevant views on clean code 😅 , I ended up proposing to simplify this, to avoid using nestjs-base-service "just" to get filter/sort/partial match (I actually forgot to explicitly mention partial match in the proposed implementation, but that's another `Array.filter()` away), and overall to simplify how we handle this.

I'd be happy as well if we wanted to keep your current implementation, with some cleanup for the benefit of our future selves as per my inline notes, and ideally stripping unneeded (and IMO misleading) properties from the response, but then... see my whole rationale for proposing to lightly hijack the current implementation of lists of scenario-level protected areas 😬

hotzevzl (approved):
looks good, thanks! thanks for putting up with my changes of heart and well done for cleaning up the code (for some definition of clean aligning with my definition of clean, but still 😬)

let's merge and deploy, please!

PR #1512, opened 2023-09-20
added coalesce to avoid issue when property does not exist

aagm:
Hotfix that adds a coalesce to prevent a side effect of concatenating a null value with a json.


hotzevzl (approved):
good catch 🙇🏼

PR #1537, opened 2023-10-06
avoid error on export if no scenario has ever been run [MRXN23-473]

hotzevzl:
See https://vizzuality.atlassian.net/browse/MRXN23-473


yulia-bel (approved):
Tested locally by duplicating empty project (no scenarios), all works well

PR #1539, opened 2023-10-06
[api] Cost surfaces tiles endpoint [MRXN23-118]

yulia-bel:
### Overview

New endpoint in ProjectCostSurfaceController to retrieve Cost Surface tiles.
Implementation is similar to the endpoint for tiles for Geo Features

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/projects/project-cost-surface.controller.ts
+++ b/api/apps/api/src/modules/projects/project-cost-surface.controller.ts
@@ -261,7 +261,6 @@ export class ProjectCostSurfaceController {
 
   @ImplementsAcl()
   @UseGuards(JwtAuthGuard)
-  @TilesOpenApi()
   @ApiOperation({
     description: 'Get cost surface tiles.',
   })
--- a/api/apps/api/src/modules/projects/project-cost-surface.controller.ts
+++ b/api/apps/api/src/modules/projects/project-cost-surface.controller.ts
@@ -1,17 +1,20 @@
 import {
   Body,
   Controller,
+  ForbiddenException,
   Delete,
   forwardRef,
   Get,
   Inject,
-  Param,
+  Param, ParseIntPipe,
+  ParseUUIDPipe,
   Patch,
   Post,
   Req,
+  Res,
   UploadedFile,
-  UseGuards,
-} from '@nestjs/common';
+  UseGuards
+} from "@nestjs/common";
 import {
   ApiForbiddenResponse,
   ApiOkResponse,
@@ -45,6 +48,10 @@ import {
 import { ensureShapefileHasRequiredFiles } from '@marxan-api/utils/file-uploads.utils';
 import { UpdateCostSurfaceDto } from '@marxan-api/modules/cost-surface/dto/update-cost-surface.dto';
 import { CostSurfaceSerializer } from '@marxan-api/modules/cost-surface/dto/cost-surface.serializer';
+import { JSONAPICostSurface } from '@marxan-api/modules/cost-surface/cost-surface.api.entity';
+import { TilesOpenApi } from '@marxan/tiles';
+import { Response } from 'express';
+import { ProxyService } from '@marxan-api/modules/proxy/proxy.service';
 import {
   costSurfaceResource,
   CostSurfaceResult,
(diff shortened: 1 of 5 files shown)


hotzevzl (approved):
thanks @yulia-bel, overall all is good and it works as expected as far as I could check.

Besides adding OpenAPI docs decorators to the main API (because that's what FE devs use as reference - in theory we could even drop openapi docs from the geoprocessing service altogether!), my only other comment is about something happening even with the newest simplification algorithm backported from Alicia's recent implementation - I'm getting simplification artifacts for a tile such as 6/34/34 for an Okavango test project: the end result looks kind of neat, but it's not what users will expect:

pattern

api/apps/geoprocessing/src/modules/cost-surface/cost-surface.service.ts
```diff
@@ -0,0 +1,88 @@
...
+   *   within the project grid
+   */
+  public findTile(
+    tileSpecification: CostSurfaceTileRequest,
+    bbox?: BBox,
+  ): Promise<Buffer> {
+    const { z, x, y, costSurfaceId } = tileSpecification;
+    const simplificationLevel = 360 / (Math.pow(2, z + 1) * 100);
+    const attributes = 'cost';
+    const table = `(SELECT ST_RemoveRepeatedPoints((st_dump(the_geom)).geom, ${simplificationLevel}) AS the_geom,
```
hotzevzl, line 71: about my earlier comment on simplification artifacts - maybe we could avoid applying simplification altogether here? what do you think @aagm?

in a way, we're dealing with one single set of geometries, those of a project's grid, and so I think this is less crucial than when dealing with features, where each can have their own intricated and extensive geometries. In the case of regular (hex/square) grids, geometries will end up being quite simple overall (though we may have very many, in some cases); in the case of user-uploaded grids, things can indeed be more complicated, though.

hotzevzl:
@yulia-bel if you end up doing more changes (bbox, openapi docs, whatnot), please give a final cleanup (unused imports or parameters) - no need to touch these if you're not going to make any other changes (so that we don't have to wait for CI to run again - or if you do cosmetic cleanups only, let's consider the latest good CI run as good enough as in, we can assume that cosmetic changes won't break anything 😬)

api/apps/api/src/modules/projects/project-cost-surface.controller.ts
```diff
@@ -263,7 +263,6 @@ export class ProjectCostSurfaceController {
 
   @ImplementsAcl()
   @UseGuards(JwtAuthGuard)
-  @TilesOpenApi()
```
hotzevzl: @yulia-bel apologies, I had completely missed that this decorator was in place (and I had actually completely forgotten about it even existing). This was correct, we had all the params documented via this decorator. I just meant that we would not necessarily need OpenAPI docs for the _geoprocessing_ endpoint, but here all was correct - sorry!

The only glitch here is that `@TilesOpenApi()` includes the `@ApiQuery` decorator for `bbox`, which we may not want here. If we wanted to _optionally exclude_ the bbox param (so that we don't have to touch any of the existing uses of the `@TilesOpenApi()` decorator), we could change the decorator with something like the patch below

```patch
diff --git a/api/libs/tiles/src/tiles-api-doc.ts b/api/libs/tiles/src/tiles-api-doc.ts
index 7beb84368..965d0ca5f 100644
--- a/api/libs/tiles/src/tiles-api-doc.ts
+++ b/api/libs/tiles/src/tiles-api-doc.ts
@@ -7,7 +7,16 @@ import {
   ApiUnauthorizedResponse,
 } from '@nestjs/swagger';

-export const TilesOpenApi = () =>
+export const TilesOpenApi = (excludeBbox: boolean) => {
+  const bboxDecorator = excludeBbox ? [] : [
+      ApiQuery({
+        name: 'bbox',
+        description: 'Bounding box of the project [xMin, xMax, yMin, yMax]',
+        type: [Number],
+        required: false,
+        example: [-1, 40, 1, 42],
+      }),
+  ];
   applyDecorators(
     ...[
       ApiParam({
@@ -35,13 +44,6 @@ export const TilesOpenApi = () =>
         required: false,
         example: 'e5c3b978-908c-49d3-b1e3-89727e9f999c',
       }),
-      ApiQuery({
-        name: 'bbox',
-        description: 'Bounding box of the project [xMin, xMax, yMin, yMax]',
-        type: [Number],
-        required: false,
-        example: [-1, 40, 1, 42],
-      }),
       ApiOkResponse({
         description: 'Binary protobuffer mvt tile',
         type: String,
@@ -49,4 +51,6 @@ export const TilesOpenApi = () =>
       ApiBadRequestResponse(),
       ApiUnauthorizedResponse(),
     ],
-  );
+    ...bboxDecorator
+    );
+  };
```

PR #1541, opened 2023-10-09
use `cost-surfaces` as part of path for cost surface-related endpoints (rather than `cost-surface`) [MRXN23-479] [MRXN23-480]

hotzevzl:
see https://vizzuality.atlassian.net/browse/MRXN23-479


api/apps/api/src/modules/projects/projects.protected-areas.controller.ts
```diff
@@ -88,4 +97,37 @@ export class ProjectProtectedAreasController {
...
+  @GeometryFileInterceptor(GeometryKind.Complex)
+  @ApiTags(asyncJobTag)
+  @Post(':id/protected-areas/shapefile')
+  async shapefileForProtectedArea(
+    @Param('id') projectId: string,
+    @UploadedFile() file: Express.Multer.File,
+    @Req() req: RequestWithAuthenticatedUser,
+    @Body() dto: UploadShapefileDto,
+  ): Promise<JsonApiAsyncJobMeta> {
+    const outcome = await this.projectsService.addProtectedAreaFor(
```
yulia-bel, line 115: just a question/suggestion - maybe move the addProtedtedAreaFor() to ProjectProtectedAreasService.. its basically just doing the file validation and acl though and then passes to AddProtectedAreaService, so maybe not worth it.. but being inspired by this nice cleanup, feels a bit like there are too much services involved in project protected areas

hotzevzl: yep, it does make sense, but then we enter the rabbit hole of cleaning up ProjectsService which is another sprawling chaos, and... I'd rather stop at the controller layer for this PR :grimacing:

PR #1551, opened 2023-10-16
[api] Add range of amounts (min, max) for feature per PU [MRXN23-444]

yulia-bel:
### Overview

This PR adds 2 new columns to (apiDB).features table to store min and and amount values.
As agreed before, service (in GeoFeaturesServices) will retrieve values from (geoDB)puvspr_calculations table by featureId and save them to (apiDB).features table. the serviced will be reused for all feature types..

Tests updated accordingly and to be reviewed when implementing [https://vizzuality.atlassian.net/browse/MRXN23-460](url) (currently null values are being saved)

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file


hotzevzl (commented):
thanks @yulia-bel! just some performance notes to make sure things work smoothly irrespective(ish) of how many `puvspr_records` are in a given instance, for how many distinct `feature_id`s, and of how many features we want to set min/max for at once (through the new `GeoFeaturesService.saveAmountRangeForFeature()` method)

api/apps/api/src/modules/geo-features/import/features-amounts-upload.service.ts
```diff
@@ -107,6 +117,12 @@ export class FeatureAmountUploadService {
         apiQueryRunner.manager,
       );
 
+      this.logger.log(`Saving min and max amounts for new features...`);
+
+      await this.geoFeaturesService.saveAmountRangeForFeatures(
+        newFeaturesFromCsvUpload.map((feature) => feature.id),
+      );
```
hotzevzl, line 124: hm, this won't work - I had missed this in my initial review but as I've been checking this locally, I realised that since we are inserting puvspr data in a transaction, and `saveAmountRangeForFeatures()` doesn't use the entity manager of the ongoing transaction, when trying to get min/max within `saveAmountRangeForFeatures()` we will always get an empty result, as the entity manager within this function won't be aware of the transaction.

We could either move these three lines past the committing of the transactions, or allow to pass in an optional entity manager, so that if this is available it is used, otherwise we fall back to `this.geoEntityManager` injected into `GeoFeaturesService`. I wouldn't make the entity manager a compulsory argument, as we may not need in every context where we call this function (or if we do, we can make it compulsory when we consolidate all the `min/max` + `puvspr_calculations` bits in the leftover tasks for id0.

hotzevzl: actually we'd need to pass in both entity managers from the ongoing transactions, so that we can either commit or roll back everything at once, including the min/max amounts in the apidb (there may be caveats here, but probably can be ignored in practice for this specific use case)

hotzevzl (approved):

PR #1554, opened 2023-10-17
avoid relinking planning areas already used for a project [MRXNM-31]

hotzevzl:
https://vizzuality.atlassian.net/browse/MRXNM-31

### DIFF
--- a/api/apps/api/test/projects/projects.fixtures.ts
+++ b/api/apps/api/test/projects/projects.fixtures.ts
@@ -5,12 +5,12 @@ import { CloningFilesRepository } from '@marxan/cloning-files-repository';
 import { ComponentId, ComponentLocation } from '@marxan/cloning/domain';
 import { HttpStatus } from '@nestjs/common';
 import { CommandBus, CqrsModule } from '@nestjs/cqrs';
-import { getRepositoryToken } from '@nestjs/typeorm';
+import { TypeOrmModule, getEntityManagerToken, getRepositoryToken } from '@nestjs/typeorm';
 import { isLeft } from 'fp-ts/lib/Either';
 import { Readable } from 'stream';
 import * as request from 'supertest';
-import { Repository } from 'typeorm';
-import { validate, version } from 'uuid';
+import { EntityManager, Repository } from 'typeorm';
+import { v4, validate, version } from 'uuid';
 import { ApiEventsService } from '../../src/modules/api-events';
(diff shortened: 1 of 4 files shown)


api/apps/api/test/projects/projects.fixtures.ts
```diff
@@ -547,3 +640,78 @@ export const getFixtures = async () => {
...
+      name: `${
+        nameForCostSurface ? nameForCostSurface + ' - ' : ''
+      }Default Cost Surface`,
+      projectId: projectId,
+      min: 0,
+      max: 0,
+      isDefault: true,
+    })
+    .execute();
+}
```
KevSanchez, line 717: Is there any special reason why you're not importing these functions instead? Not a big deal, but it could easier to refactor down the line importing them; having a copy might instill doubtfulness as to whether these are the exact same functions or they defined separately because of any possible issue.

hotzevzl: yep, it's ugly. I'd have to move these fixtures to somewhere under `api/libs` though, as the original code is in `geoprocessing` and here I'm using it in `api`, and when we build the OCI images for `api` test runners in CI we delete the whole tree of `geoprocessing`.

Indeed I thought about tidying up some common functions used in tests somewhere, but given that there is a lot of duplication and kinduplication (kind of duplication but not quite, with minor divergences here and there, and so on) already in our tests monster, I thought I'd just give up at this stage.

Do you have suggestions on how to handle this without much moving around of stuff?

KevSanchez: Oh, yeah... the api/geo split. Yeah, I see what you mean now.
I agree that other than having a module on libs with common methods for testing on both apis, would be the way to go, but probably not worthy right now. To do it properly, would require scouring the tests quite a bit.
So, for now, doing it like this and tagging it as debt (maybe with some of indicator like "candidate for common tests utils grouping" or something) should be enough.

hotzevzl: agree. I've added a bit more detail to the `@debt` comments, and also added these to the geoprocessing functions from which I copied the api ones, so that the debt is clear on both sides.

hotzevzl: acf7ff46a26c9506ac7c130c0371e7793434e192

KevSanchez (approved):

PR #1559, opened 2023-10-25
Copy amount per PU for shapefiles and legacy [MRXN23-466] [MRXN23-467]

KevSanchez:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/geoprocessing/src/legacy-project-import/legacy-piece-importers/features.legacy-piece-importer.ts
+++ b/api/apps/geoprocessing/src/legacy-project-import/legacy-piece-importers/features.legacy-piece-importer.ts
@@ -30,6 +30,7 @@ import {
   PuvsprDatReader,
 } from './file-readers/puvspr-dat.reader';
 import { SpecDatReader, SpecDatRow } from './file-readers/spec-dat.reader';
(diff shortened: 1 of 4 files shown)


api/apps/geoprocessing/src/legacy-project-import/legacy-piece-importers/features.legacy-piece-importer.ts
```diff
@@ -355,4 +382,39 @@ export class FeaturesLegacyProjectPieceImporter
...
+        update features set
+           amount_min = minmax.min,
+           amount_max = minmax.max
+        from (
+        values
+            ${minMaxSqlValueStringForFeatures}
+        ) as minmax(feature_id, min, max)
+        where features.id = minmax.feature_id;`;
+    await apiEntityManager.query(query);
+  }
```
hotzevzl, line 419: ok to keep this in this PR - in a later one, I suggest to use a single instance of this code, after adding a parameter to allow to pass in an apiEntityManager and removing the reliance on any entity managers injected in the parent class, so that we can remove duplicated code... though if moving this to an existing lib (for example `libs/geofeature-calculations`) turns out to be even just slightly more than banal, duplication be it and let's move on with life.

hotzevzl (approved):

PR #1568, opened 2023-10-31
Add feature amount calculation to feature shapefile upload [MRXN-468]

KevSanchez:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/geo-features/geo-features.service.ts
+++ b/api/apps/api/src/modules/geo-features/geo-features.service.ts
@@ -53,7 +53,7 @@ import {
   FeatureAmountsPerPlanningUnitService,
 } from '@marxan/feature-amounts-per-planning-unit';
 import { ComputeFeatureAmountPerPlanningUnit } from '@marxan/feature-amounts-per-planning-unit/feature-amounts-per-planning-units.service';
-import { CHUNK_SIZE_FOR_BATCH_GEODB_OPERATIONS } from '@marxan-geoprocessing/utils/chunk-size-for-batch-geodb-operations';
+import { CHUNK_SIZE_FOR_BATCH_APIDB_OPERATIONS } from '@marxan-api/utils/chunk-size-for-batch-apidb-operations';
 
 const geoFeatureFilterKeyNames = [
   'featureClassName',
@@ -431,10 +431,14 @@ export class GeoFeaturesService extends AppBaseService<
         computedFeatureAmounts,
       );
 
+      await this.saveAmountRangeForFeatures(
+        apiQueryRunner.manager,
+        geoQueryRunner.manager,
+        [geoFeature.id],
+      );
+
       await apiQueryRunner.commitTransaction();
       await geoQueryRunner.commitTransaction();
-
-      await this.saveAmountRangeForFeatures([geoFeature.id]);
     } catch (err) {
       await apiQueryRunner.rollbackTransaction();
       await geoQueryRunner.rollbackTransaction();
@@ -468,7 +472,7 @@ export class GeoFeaturesService extends AppBaseService<
         amount,
         projectPuId,
       })),
-      { chunk: CHUNK_SIZE_FOR_BATCH_GEODB_OPERATIONS },
+      { chunk: CHUNK_SIZE_FOR_BATCH_APIDB_OPERATIONS },
     );
   }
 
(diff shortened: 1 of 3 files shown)


api/apps/api/src/modules/geo-features/geo-features.service.ts
```diff
@@ -431,6 +452,24 @@ export class GeoFeaturesService extends AppBaseService<
...
+    );
+    await repo.save(
+      featureAmounts.map(({ amount, projectPuId, featureId }) => ({
+        projectId,
+        featureId,
+        amount,
+        projectPuId,
+      })),
+    );
+  }
```
hotzevzl, line 471: I think this may break under pressure, aka when inserting a high enough number of PU amounts at once so that the combined number of query parameters goes above PostgreSQL's limit for this.


```suggestion
// add to imports
import { CHUNK_SIZE_FOR_BATCH_GEODB_OPERATIONS } from '@marxan-geoprocessing/utils/chunk-size-for-batch-geodb-operations';
//and here
    await repo.save(
      featureAmounts.map(({ amount, projectPuId, featureId }) => ({
        projectId,
        featureId,
        amount,
        projectPuId,
      })),
      {
        chunk: CHUNK_SIZE_FOR_BATCH_GEODB_OPERATIONS,
      }
    );
  }
```

api/libs/feature-amounts-per-planning-unit/src/feature-amounts-per-planning-units.service.ts
```diff
@@ -41,48 +41,51 @@ export class FeatureAmountsPerPlanningUnitService {
...
     /**
      * @TODO further performance savings: limiting scans to planning_units_geom
      * by partition (we need to get the grid shape from the parent project); use
      * && operator instead of st_intersects() for bbox-based calculation of
      * intersections.
      */
+    geoEntityManager = geoEntityManager
+      ? geoEntityManager
+      : this.geoEntityManager;
+
```
hotzevzl, line 56: I think you could simplify this:

```suggestion
    geoEntityManager: EntityManager = this.geoEntityManager,
  ): Promise<ComputeFeatureAmountPerPlanningUnit[]> {
    /**
     * @TODO further performance savings: limiting scans to planning_units_geom
     * by partition (we need to get the grid shape from the parent project); use
     * && operator instead of st_intersects() for bbox-based calculation of
     * intersections.
     */

```

hotzevzl (commented):
thanks @KevSanchez! I'm still trying out the functionality in my local dev environment, but in the meanwhile, please see a couple of notes inline.

api/apps/api/src/modules/geo-features/geo-features.service.ts
hotzevzl: 🤔 I thought I had posted a follow-up comment via my IDE but I think I may have messed that up - basically it was about my ill-informed advice to use `CHUNK_SIZE_FOR_BATCH_GEODB_OPERATIONS` as this is defined in the geoprocessing service and so it will not be importable directly in the api service. I had some random rants about semantics in my vanished comment, but basically tl;dr please let's use `CHUNK_SIZE_FOR_BATCH_APIDB_OPERATIONS` here without moving anything else around (moving constants to libs, etc.)

hotzevzl:
@KevSanchez works beautifully for me - please have a look at the leftover inline comments including my own backpedaling on the suggestion to use `CHUNK_SIZE_FOR_BATCH_GEODB_OPERATIONS` and then we can merge ❤️

api/apps/api/src/modules/geo-features/geo-features.service.ts
```diff
@@ -854,6 +893,10 @@ export class GeoFeaturesService extends AppBaseService<
       .groupBy('fappu.feature_id')
       .getRawMany();
 
+    if (minAndMaxAmountsForFeatures.length === 0) {
+      return;
+    }
```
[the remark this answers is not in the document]
KevSanchez: Yeah I agree, it would be better to throw error, and catch it later. 
The only issue in this case is that by the time this function executes, where it gets executed, the changes relater to feature amounts have already been committed (geo-features.service.ts line 437 and features-amount-upload.service.ts line 116), and catching the error won't do much more than logging it (which already better than nothing, indeed)

hotzevzl (approved):
let's merge once all is green, thanks! 🚀

PR #1569, opened 2023-10-31
fix(OutputSummary): Fixes bug generating output summary with not correct run at least once criteria [MRXN-474]

KevSanchez:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/test/projects/project-summaries/output-project-summaries.e2e-spec.ts
+++ b/api/apps/api/test/projects/project-summaries/output-project-summaries.e2e-spec.ts
@@ -110,16 +110,14 @@ describe('when generating output summary metadata for a project', () => {
   it('should generate zip files', async () => {
     // This test will arrange 2 scenarios, with 5 runs, and 5 Planning Units
     const projectId = await fixtures.GivenProjectExistsOnAPI('project');
-    const scenario1Id =
-      await fixtures.GivenScenarioWithAtLeastOneRunExistsOnAPI(
-        'scenario1Name',
-        projectId,
-      );
-    const scenario2Id =
-      await fixtures.GivenScenarioWithAtLeastOneRunExistsOnAPI(
-        'scenario2Name',
-        projectId,
-      );
+    const scenario1Id = await fixtures.GivenScenario(
+      'scenario1Name',
+      projectId,
+    );
+    const scenario2Id = await fixtures.GivenScenario(
+      'scenario2Name',
+      projectId,
+    );
 
     const scenarioIdToNameMapping = [
       {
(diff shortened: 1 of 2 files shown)


api/apps/api/test/projects/project-summaries/output-project-summaries.e2e-spec.ts
```diff
@@ -398,10 +399,13 @@ const getFixtures = async () => {
...
       return projectId;
     },
-    GivenScenarioWithAtLeastOneRunExistsOnAPI: async (
+    GivenScenarioRanAtLeastOnceExistsOnAPI: async (
       name: string,
       projectId: string,
+      numberOfRuns?: number,
     ) => {
+      // Even though numberOfRuns shouldn't be null, it can happen very sporadically because of MRXN-29
+      // so ranAtLeastOnce must be used as criteria to select the scenarios for output summary
```
hotzevzl, line 408: 😬 I think my explanation in MRXN23-474 was indeed confusing: `numberOfRuns` _has nothing to do_ with what we are checking here: `numberOfRuns` is a Marxan setting, that we use to instruct the Marxan solver about how many iterations to use when running its solver flow.

It does not tell us "how many times a user has run a scenario successfully" - we do record this information only through API events, but we don't really expose it in any way through the API or use it internally. It is only a Marxan setting, but does not describe any state of things.

The fix in the present PR is not needed because `numberOfRuns` number may be zero or undefined: it is needed because it should never be used to infer whether a scenario has ever run successfully; the only thing we should ever check is that if a scenario's `ranAtLeastOnce` property is true, then it should be included in the summary.

I recommend to remove this comment altogether, and in terms of tests, I suggest to remove the handling of `numberOfRuns` from `GivenScenarioRanAtLeastOnceExistsOnAPI()`, and adjust fixtures below to only deal with `ranAtLeastOnce`.

tl;dr we should not be concerned at all with `numberOfRuns` in this PR: setting it explicitly to zero should not even be something we ever do in tests or in reality, unless we want to test that Marxan will still default to its default number of runs if it is supplied a number that doesn't make sense (but this would be testing Marxan, not our app, so I'd forget about this altogether 😅)


```suggestion
```

hotzevzl (changes requested):
thanks @KevSanchez 

the fix is ok, and thanks for improving `ThenOutputCSVDoesNotContainScenarioWith()` to be able to handle more complex test scenarios.

I've left some inline notes about possible confusion (my bad in the explanation in the Jira story, sorry!) stemming from the use of `numberOfRuns` in test fixtures - see the notes, but tl;dr is that I recommend to not be bothered with `numberOfRuns` in these tests at all

api/apps/api/test/projects/project-summaries/output-project-summaries.e2e-spec.ts
KevSanchez: Ahhh I see what you mean, I wasn't aware of that property being a setting per se. Yeah, I agree, better to not leave that info on there to avoid further confusion.

hotzevzl (approved):
great, thanks!

PR #1577, opened 2023-11-14
remove deprecated scenario-level endpoints [MRXN23-506]

hotzevzl:
https://vizzuality.atlassian.net/browse/MRXN23-506


KevSanchez:
Wondering if it would be worth taking the opportunity and remove the ScenarioService methods used that are not relevant anymore, `processCostSurfaceShapefile` and `getCostRange`. The latter seems to be easy enough, as it just goes 2 function levels deep, but the former, involves commands, events and most surely jobs, it can turn into a bit of a rabbit hole.

andresgnlez:
@hotzevzl @KevSanchez  https://github.com/Vizzuality/marxan-cloud/pull/1634 should unlock this.

hotzevzl:
> Wondering if it would be worth taking the opportunity and remove the ScenarioService methods used that are not relevant anymore, `processCostSurfaceShapefile` and `getCostRange`. The latter seems to be easy enough, as it just goes 2 function levels deep, but the former, involves commands, events and most surely jobs, it can turn into a bit of a rabbit hole.

excellent point, and silly me - good catch.

I think I've removed all that could be removed now. Waiting for tests to run, and doing a functional check in a dev environment in the meanwhile 🤞🏼

hotzevzl:
ok, all tests passed and functional QA was ok - I'll merge

PR #1578, opened 2023-11-14
prevent db deadlocks while updating feature tags [MRXN23-316]

hotzevzl:
https://vizzuality.atlassian.net/browse/MRXN23-316

### DIFF
--- a/api/apps/api/src/modules/geo-feature-tags/geo-feature-tags.service.ts
+++ b/api/apps/api/src/modules/geo-feature-tags/geo-feature-tags.service.ts
@@ -160,20 +160,27 @@ export class GeoFeatureTagsService {
     await apiQueryRunner.startTransaction();
 
     try {
-      const previousTag = await this.geoFeatureTagsRepo.findOne({
+      const previousTag = await apiQueryRunner.manager.findOne(GeoFeatureTag, {
         where: { projectId, featureId },
       });
       if (previousTag) {
-        await this.geoFeatureTagsRepo.delete(previousTag.id);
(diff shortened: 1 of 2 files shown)


hotzevzl:
@KevSanchez so, I've been unable to consistently reproduce the db transaction festival in my local dev environment, but I'm actually able to reproduce this almost always on marxan23 with the current `develop` code, whereas when checking out this WIP branch (still on m23), I have not been able to trigger the deadlock ever.

Not that I've actually spent hours doing this, but combining my initial guess of where the issue may have been, with the changes I'm suggesting in this PR as well as the behaviour just noted above about reproducing this (or not) on marxan23, I'd be inclined to vaguely trust that the mix of in-transaction and out-of-transaction db ops in the current implementation _may_ be at least part of the observed behaviour.

Given how jumpy the behaviour is, I am not sure how to test that the proposed fix is indeed a fix and that the observed problem is indeed a problem: I think the problem manifests itself depending on timing of sequential operations, so even writing a test that for example creates 100 features and then tries to bulk edit their tags all at once may occasionally still pass without the changes in this PR depending on... things (sequence of requests, load on CPU and IO for the node where the tests are running, etc.).

Maybe still worth it through brute force? I am not sure. Especially at this stage I'd be inclined to be satisfied with a cross-check of the changes from your end.

KevSanchez:
I think it's best to do as you suggest, as the nature of the error was really erratic for me too; I could bulk edit a couple of tags several times in a row without issue, and other times, even right after spinning the API up.
I noticed that the previous code, by not really doing the operations inside the transaction block might explain this. In any case, it makes much more sense the way you're suggesting now.

KevSanchez (approved):

PR #1612, opened 2023-12-14
report min/max amounts in square km for features from shapefile [MRXN23-528]

hotzevzl:
https://vizzuality.atlassian.net/browse/MRXN23-528


KevSanchez:
Everything OK, added the rounding logic as per the comments on the JIRA issue

KevSanchez (approved):

PR #1632, opened 2024-01-18
adds upcoming changes banner to homepage (cherry-picked change for MaPP v1 production environment) [MRXN23-560]

hotzevzl:
### Overview

Adds notice banner for upcoming changes, for MaPP v1 (currently only used in the TNC production environment)

### Designs

N/A

### Testing instructions

Banner should be shown in production when enabled via feature flag. Nothing else should break 🙏🏼 .

### Feature relevant tickets

https://vizzuality.atlassian.net/browse/MRXN23-560

---

- [ ] Update CHANGELOG file

### DIFF
--- a/app/layout/upcoming-changes/component.tsx
+++ b/app/layout/upcoming-changes/component.tsx
@@ -1,0 +1,44 @@
+import { useCallback } from 'react';
+
+import { useCookies } from 'react-cookie';
+
+import { useFeatureFlags } from 'hooks/feature-flags';
+
+import Button from 'components/button';
+import Wrapper from 'layout/wrapper';
+
+export const UpcomingChangesBanner = (): JSX.Element => {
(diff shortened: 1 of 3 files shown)


hotzevzl:
@andresgnlez this is what I think we need in order to add the notice banner for the upcoming MaPP release to the production instance - could you please cross-check that I didn't do anytyhing silly here?

The vercel deploy triggered as part of this PR's actions is failing because we are feeding the Vercel build the old and obsolete Node version from our `package.json` - though this should not be a problem in production as we just keep running what we have there in terms of Node runtime version.

andresgnlez:
@hotzevzl yes, this is perfect, thanks Andrea!

And yes, I received the fail notification from Vercel, we should probably shut down that instance as it is not being used, AFAIK, but that's a battle for another day :D

andresgnlez (approved):

PR #1633, opened 2024-01-19
fix(tags): Tags are now trimmed down of white spaces before updating them [MRXN23-567]

KevSanchez:
### Overview

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/api/apps/api/src/modules/geo-feature-tags/dto/update-geo-feature-tag.dto.ts
+++ b/api/apps/api/src/modules/geo-feature-tags/dto/update-geo-feature-tag.dto.ts
@@ -1,6 +1,7 @@
 import { ApiProperty } from '@nestjs/swagger';
-import { IsNotEmpty, MaxLength, Validate } from 'class-validator';
+import { IsNotEmpty, IsString, MaxLength, Validate } from 'class-validator';
 import { IsValidTagNameValidator } from '@marxan-api/modules/geo-feature-tags/validators/is-valid-tag-name.custom.validator';
+import { Transform } from 'class-transformer';
 
 export const tagMaxlength = 30;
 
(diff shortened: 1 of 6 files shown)


hotzevzl (commented):
@KevSanchez thank you!

The strategy is good. A couple of notes:

- should we also decorate `UploadShapefileDTO`? (I think this one DTO would be the only one missing)
- there could be some paranoia-level edge cases where this may fail: since we don't (but probably should?) use `@IsString()` decorators on these properties, if a client bug (which is where the paranoia bit kicks in: for the time being I think we would not be expecting anything else than our web app and our bots, which don't even deal with tags, as clients for this API) gets these properties in as numbers in the request payloads, `value.trim()` would fail with `value.trim is not a function`.

I think we can ignore this in practice, but if you think that adding the type-enforcing decorator is worth it, please go ahead.

KevSanchez:
> @KevSanchez thank you!
> 
> The strategy is good. A couple of notes:
> 
>     * should we also decorate `UploadShapefileDTO`? (I think this one DTO would be the only one missing)
> 
>     * there could be some paranoia-level edge cases where this may fail: since we don't (but probably should?) use `@IsString()` decorators on these properties, if a client bug (which is where the paranoia bit kicks in: for the time being I think we would not be expecting anything else than our web app and our bots, which don't even deal with tags, as clients for this API) gets these properties in as numbers in the request payloads, `value.trim()` would fail with `value.trim is not a function`.
> 
> 
> I think we can ignore this in practice, but if you think that adding the type-enforcing decorator is worth it, please go ahead.

Good catch, it's been a while, and that one slipped by.
As for enforcing the type safety, if @IsString() is not there I think for consistency we should include, we pretty much have type enforcement all across aboard. I highly doubt they are going to use tags based solely on numbers, right?

hotzevzl:
> [...]
> As for enforcing the type safety, if @IsString() is not there I think for consistency we should include, we pretty much have type enforcement all across aboard. I highly doubt they are going to use tags based solely on numbers, right?

who knows, but on the other hand, even if so, the FE would likely supply these as numbers-in-string - but yes, adding a validation decorator would help to align this with much of our other DTOs.

PR #1652, opened 2024-02-20
[MRXN23-531]: invalidates cost-surface fetch a second later of creation

andresgnlez:
### Overview

This PR delays the query invalidation of cost surfaces up to one second so it avoids falling in limbo when the API processes the cost surface between the POST callback and the first `/status` polling. If the API takes more than that second into processing the cost-surface, the `/status` will in charge of telling the application when the query invalidation should occur.

For future reference:

Screenshot 2024-02-20 at 16-56-43 Excalidraw — Collaborative whiteboarding made easy

### Designs

### Testing instructions

### Feature relevant tickets

https://vizzuality.atlassian.net/browse/MRXN23-531

---

- [ ] Update CHANGELOG file

### DIFF
--- a/app/hooks/cost-surface/index.ts
+++ b/app/hooks/cost-surface/index.ts
@@ -1,4 +1,4 @@
-import { useQuery, QueryObserverOptions, useMutation } from 'react-query';
+import { useQuery, QueryObserverOptions, useMutation, useQueryClient } from 'react-query';
 
 import { useSession } from 'next-auth/react';
 
@@ -86,6 +86,7 @@ export function useEditProjectCostSurface() {
 
 export function useUploadProjectCostSurface() {
   const { data: session } = useSession();
+  const queryClient = useQueryClient();
 
   const uploadProjectCostSurface = ({ id, data }: { id: CostSurface['id']; data: FormData }) => {
     return UPLOADS.request({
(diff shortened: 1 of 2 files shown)


hotzevzl:
@andresgnlez thanks for all your patient explanations off-band today.

All looks good to me, and I can reliably get cost surfaces to display once the async upload job has completed, with this change applied.

What puzzles me a bit is that:
- on marxan23.vercel.app (so, without this change applied until now), I can't trigger the behaviour whereby a cost surface is not added to the sidebar and legend once its async upload job is done, _even with super-simple grids_ where the limbo situation will very often happen - basically there the undesirable behaviour doesn't seem to happen _already_ 🤷🏼 
- using the same `develop` branch as marxan23.vercel.app but running it locally, and with the marxan23 API as backend, I cannot trigger the undesirable behaviour either 🤷🏼 
- same as above (`develop` frontend running locally) but using `api.marxanplanning.org` as backend, I can consistently trigger the undesirable behaviour

So I guess there's some PEBKAC somewhere on my end 🤷🏼 but nevertheless, since local app with this PR's branch and api.marxanplanning.org as backend _always avoids the undesirable behaviour_, I'd say that we can consider this fixed. Why on earth a different backend would make any difference is 🤯 to me, but I also think that trying to understand this at this stage is not a battle worth fighting. In case this issue raises its head again somehow, then hopefully we'll remember what I observed while checking this PR in case that could help to shed any light. But hopefully we won't need any of this once merging this PR - which I'm going to do in a few moments.

thanks @andresgnlez!

hotzevzl (approved):

PR #1662, opened 2024-03-11
[MRXN23-604] fixes target/spf values after editing one of them

andresgnlez:
### Overview

This PR fixes an issue were the features listed were defaulted to the app's default values (for target and spf) when one of the features was edited by selecting it (through the checkbox) and clicking on the edit button in the floating menu.

The PR also displays the current values for target and spf when the user edits their values instead of defaulting to the app's default values. If more than one features are selected to edit, then the app's default values will display though.

image

### Designs

### Testing instructions

### Feature relevant tickets

https://vizzuality.atlassian.net/browse/MRXN23-604

---

- [ ] Update CHANGELOG file

### DIFF
--- a/app/layout/project/sidebar/scenario/grid-setup/features/target-spf/all-targets/index.tsx
+++ b/app/layout/project/sidebar/scenario/grid-setup/features/target-spf/all-targets/index.tsx
@@ -34,13 +34,16 @@ const AllTargetsSelector = ({
             type="number"
             min={0}
             max={100}
+            step={0.01}
             defaultValue={values.target}
             value={values.target}
-            // disabled={!editable}
             onChange={({ target: { value: inputValue } }) => {
+              const numericValue = Number(inputValue);
+              if (numericValue < 0 || numericValue > 100) return;
+
               setValues((prevValues) => ({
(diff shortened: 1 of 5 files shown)


hotzevzl:
@andresgnlez thanks - I didn't really get the magic behind the change, but it does what it should do - values stay as they are unless they are touched/changed.

I really love the solution you devised to clarify bulk edits! This is _vastly_ better than the original one based on focus events, and way clearer.

A couple of minor thing about the bulk edit popup:
- the target % should only allow decimal numeric values between 0 and 100 inclusive
- spf does not need to be capped at 1; in fact the Marxan manual is silent on min/max for this, so I think in practice users may want to enter any numeric value (decimal or integer) as they see fit - the value is used as a multiplier so I think it should not be zero, and I don't see when a negative spf may make sense, but I'd still just constrain this to any numeric value, decimal or integer

Other than this, all looks good to me, thanks, and well done for the big leap forward on the bulk edit UX! 🙌🏼

andresgnlez:
@hotzevzl noted! I added the validations to the fields. I'll be merging this!

PR #1667, opened 2024-03-15
pass through gap data (target/met) values with rounding to 4 decimal digits [MRXN23-601]

hotzevzl:
https://vizzuality.atlassian.net/browse/MRXN23-601

Screenshot 2024-03-15 at 11-28-26 Scenarios Edit test 1 - MRXN23-601 1_x
Screenshot 2024-03-15 at 11-28-02 Scenarios Edit test 1 - MRXN23-601 1_x

### DIFF
--- a/api/apps/api/src/modules/scenarios-features/scenario-features-gap-data.service.ts
+++ b/api/apps/api/src/modules/scenarios-features/scenario-features-gap-data.service.ts
@@ -17,6 +17,7 @@ import { ScenarioAccessControl } from '@marxan-api/modules/access-control/scenar
 import { assertDefined } from '@marxan/utils';
 import { forbiddenError } from '@marxan-api/modules/access-control';
 import { Either, left, right } from 'fp-ts/lib/Either';
+import { plainToClass } from 'class-transformer';
 
 @Injectable()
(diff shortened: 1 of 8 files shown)


hotzevzl:
@andresgnlez thank you for reviewing the tiny changes in this PR. I have checked the way this works and it all looks ok - one thing I should mention is that because of reasons, the ORM gives us these values as numbers-in-a-string now (basically, because the possible precision of the db types for these values may not "fit" within JS numeric types) -- so while this "works", I wanted to make you aware of this in case you can think of any possible caveats/edge cases with this.

Or maybe I should "just" re-cast these numbers-in-a-string to proper JS numbers in the API, and avoid getting numbers in weird formats to API clients to start with... I'd do this next week

andresgnlez:
@hotzevzl happy to test this once is available in m23 instance, AFAIK the change is exclusive of the `gap-data` endpoint, isn't it? Let me know if you plan to extend this change through the codebase, if so, I might need to take a deeper look as we might be using `Number` methods straightforward in some cases or just the fact of having arithmetic operations between strings and number (they should work, but in JS world anything could happen 😬 ) scares me.

PR #1690, opened 2024-09-04
feature bulking: added edition modes

andresgnlez:
### Overview

This PR solves the casuistic of an user being able to perform unwanted changes when performing certain bulking actions.

image

### Designs

### Testing instructions

### Feature relevant tickets

---

- [ ] Update CHANGELOG file

### DIFF
--- a/app/layout/project/sidebar/scenario/grid-setup/features/target-spf/bulk-action-menu/modals/edit/index.tsx
+++ b/app/layout/project/sidebar/scenario/grid-setup/features/target-spf/bulk-action-menu/modals/edit/index.tsx
@@ -10,13 +10,21 @@ import { useToasts } from 'hooks/toast';
 import Button from 'components/button';
 import Field from 'components/forms/field';
 import Label from 'components/forms/label';
+import Radio from 'components/forms/radio';
 import { composeValidators } from 'components/forms/validations';
 import { Feature } from 'types/api/feature';
 
 export type FormValues = {
   target: number;
   spf: number;
+  mode: 'all' | 'only-target' | 'only-spf';
 };
+
+const EDITION_MODES = [
+  { value: 'all', label: 'Target & SPF' },
+  { value: 'only-target', label: 'Target' },
+  { value: 'only-spf', label: 'SPF' },
+];
 
 const INPUT_CLASSES =
   'h-10 w-full rounded-md border border-gray-400 px-3 text-gray-900 focus:border-none focus:outline-none focus:ring-1 focus:ring-blue-600';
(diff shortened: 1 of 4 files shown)


andresgnlez:
@hotzevzl I think the solution (and the pic) are quite self-explanatory, but let me know if something comes up.

hotzevzl (approved):
@andresgnlez thank you - I like your solution, clear and simple (at least to understand and use, not sure about the implementation itself 😬 but that was on you, in any case 😅).

I think this solves nicely the UX limitations that users were reporting, from the original design.

One small thing - while trying to punch in some random values when using this feature branch, I run into a small glitch (probably unrelated, maybe it's been there all along and we never noticed nor did users): for specific values (so far, I've been able to trigger it by setting target to 55% specifically, I couldn't figure out other such values), I think that the value gets mangled before being sent to the API, setting it to `55.00000000000001` 🤔 

The screenshots below show what I see (this is Firefox 129.0 on Arch Linux, in case it's browser-related) after saving a bulk edit with target set by myself to `55`, once I see the feature list again:

Screenshot 2024-09-06 at 09-44-53 Scenarios Edit test 1 - MRXN23-524 refetching cost surfaces test 1_x

And once I reopen the bulk edit modal (while selecting only one feature, of course - if I select more than one I get the default `50` and `1` values respectively):

Screenshot 2024-09-06 at 09-44-37 Scenarios Edit test 1 - MRXN23-524 refetching cost surfaces test 1_x

hotzevzl:
also in terms of UX here - maybe when bulk editing with > 1 feature selected, rather than showing some default values (50% and 1 respectively for target and fpf), should we leave the inputs blank? Really minor, and I don't think this will ever be a real cause of confusion for users (such as "whoops, I set all the targets to 50%, I didn't really want to do that, I just forgot to edit that value"), and it's also something that - if ever happens - is easy to recover from, by bulk editing again the relevant features, so I'm ok with us leaving things as they are (and have always been, IIRC) in terms of these defaults when the bulk edit modal is brought up.

hotzevzl:
and another small thing - at last in FF 129, if I bulk select all features by toggling the check box above the features, then edit the features and save the edits, once the processing finishes, I am back to the list of features, with none of the features selected, but the "select all" check box selected - again, super minor and I'd try to avoid cramming more things into this PR, but in case it's something really easy to change and regression-proof, please consider if it's worth changing:

Screenshot 2024-09-06 at 10-00-00 Scenarios Edit test 1 - MRXN23-524 refetching cost surfaces test 1_x

PR #1750, opened 2026-06-19
feat(scripts): split-feature amounts backfill (phase 2) [MRXNM-82]

KevSanchez:
Follow-up to #1749 (merged). Adds `data/scripts/feature-split-backport/backfill_split_feature_amounts.py` — phase 2 of legacy split restoration.

## What
Recomputes `feature_amounts_per_planning_unit` for legacy split features that have geometry (`feature_data_stable_ids`, after the phase-1 backport) but never had per-PU amounts materialized — so they stay blank in the scenario abundance view (`forProject=true` tile path). Mirrors `ComputeArea.computeAreaPerPlanningUnitOfFeature` / `FeatureAmountsPerPlanningUnitService.computeMarxanAmountPerPlanningUnit` exactly.

## Safety
Scoped to the relevant splits only: `operation='split'` AND non-empty `stable_ids` AND no existing amounts. Each compute is per-feature (its own stable_ids + its own project's PUs) so it can never write amounts for another feature. Dry-run default, idempotent.

## Prod buckets (639 splits)
- 422 reconstructable + has amounts → phase-1 backport alone
- 41 reconstructable, no amounts (live scenarios) → **this script**
- 176 broken/dead → delete (data cleanup)

## Validation (staging, 2026-06-19)
Phase-1 relinked the Spain split (5485 ids); this script's dry-run computed its amounts (2 PU rows); a fresh UI split on Rwanda rendered correctly (geometry + amounts). README documents the two-phase Nectar runbook.

### DIFF
--- a/data/scripts/feature-split-backport/README.md
+++ b/data/scripts/feature-split-backport/README.md
@@ -1,11 +1,18 @@
-# Backport `feature_data_stable_ids` for legacy split features [MRXNM-82]
+# Restore legacy split features [MRXNM-82]
 
-Legacy materialized split features (`(apidb)features` rows with
-`from_geoprocessing_ops->>'operation' = 'split'`) were created before the
-materialized-splits work and have an empty `feature_data_stable_ids`. The new
-code renders them with no geometry. `backport_feature_data_stable_ids.py`
-reconstructs that linkage from the parent feature's
-`(geodb)feature_properties_kv` → `features_data.stable_id`.
+Two one-off scripts that repair legacy materialized split features
+(`(apidb)features` rows with `from_geoprocessing_ops->>'operation' = 'split'`)
+created before the materialized-splits work:
+
+1. **`backport_feature_data_stable_ids.py`** — relinks geometry: populates the
+   empty `features.feature_data_stable_ids` from the parent feature's
+   `(geodb)feature_properties_kv` -> `features_data.stable_id`.
+2. **`backfill_split_feature_amounts.py`** — recomputes the per-planning-unit
+   amounts (`feature_amounts_per_planning_unit`) for the subset of live splits
+   that never had them.
+
+New splits created by the current `split-operation` code need NEITHER — they get
+both at creation. These scripts only touch pre-existing legacy splits.
 
 ## Subset definition (decided 2026-06-18)
 
@@ -22,7 +29,7 @@ rows — NOT bbox-filtered. Rationale:
 
 Value match mirrors `SplitQuery`: `trim('"' FROM fpkv.value::text)`.
 
-## Usage
+## Phase 1 — geometry backport (`backport_feature_data_stable_ids.py`)
 
 ```bash
 # via the bastion tunnel (staging->9432, prod->9433), dry-run (default):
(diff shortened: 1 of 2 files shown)


KevSanchez:
Heads-up from running this on production (2026-06-22):

**The backport ran fine** — 463 splits relinked. **This amounts backfill is only safe for small splits as written.**

On the 20 large `world_ecosystems_simp` globals, the per-feature `ST_Union` is computationally infeasible: a **3,317-geometry** split ran **2h45m** of CPU-bound `ST_Union` before I terminated it. The script then stalls indefinitely, and since it wraps everything in a **single end-of-run commit**, that also strands the already-computed features.

I applied it scoped to the 21 trivial (n=1) gap features only (committed cleanly). The 20 large ones are left without amounts and tracked as **MRXNM-99** (they need a non-`ST_Union` computation; geometry already renders via the stable-id backport).

Suggest before merge/unscoped use: add a **stable-id-count / time guard** (skip + report features above a threshold) and/or **commit per feature** so a slow one can't strand the rest. Also note `pg_cancel_backend` does not stop a stuck PostGIS `ST_Union` — `pg_terminate_backend` is needed.

KevSanchez:
Guardrails implemented in 85ee0f24b (on this branch):

- **`--max-stable-ids N` (default 1000)** — primary guard; skips + reports splits whose stable-id count exceeds N instead of attempting the infeasible `ST_Union` (deferred to MRXNM-99).
- **`--statement-timeout S` (default 600)** — secondary net per feature (best-effort; documented that a stuck PostGIS `ST_Union` may not honor it).
- **Per-feature commit** — a slow/failed feature can no longer strand the ones already computed; failures are caught + reported, non-fatal.

Verified on prod (dry-run): 463 candidates → 443 skipped (amounts present), 20 skipped (too-large globals), 0 computed, **0 `ST_Union` queries executed**. Safe to run unscoped now.

KevSanchez:
## Per-PU-union compute — large splits are now tractable (and already correct)

Followed up on the `--max-stable-ids` guard by actually fixing the compute it was guarding against.

**The fix.** The amount compute built one global `ST_Union` of all of a split's geometries *before* touching the project's PUs — superlinear in vertex count, which is what ran 2h45m on prod. Flipped it to **intersect-per-PU-first, then union per PU**. Identical result because `Area((⋃gᵢ)∩pu) = Area(⋃(gᵢ∩pu))`, but cost now scales with the project's PU count, not the feature's global extent.

**Verified 3 ways on prod:**
- Feasible: worst case 58,820 ids @ 7,123 PUs → **~19s** (was killed at 2h45m); 37,578 ids @ 30,837 PUs → ~72s.
- Equivalent to the old global-union compute: max relative diff **2e-15**.
- Matches the API's own stored amounts: **exact** (0.0 diff).

So `--max-stable-ids` now defaults **OFF** (kept as an optional cap) and `--statement-timeout` becomes a reliable net.

**Ran it over the 20 outstanding heavy splits on prod (dry-run): all compute 0 PU rows.** They're the ecosystem classes (Cool Temperate / Boreal / Snow & Ice) whose global geometry doesn't overlap their **tropical Zambia / subtropical Brisbane** project grids — so they correctly have no per-PU amounts. The data was already right; the only defect was that *verifying* it used to take 2h45m instead of 19s. Nothing written to prod. MRXNM-99 resolved.

